PR #128, opened 2022-08-25
[RSDK-490] Add support for Operations

njooma:
This is a first pass at adding support for Operations.

Here's the basic premise:

The main point of supporting Operations in the SDKs is to cancel/stop any tasks when the opid is cancelled. When an opid is cancelled on RDK, the associated context is cancelled as well, meaning the GRPC request sent from RDK to SDK is also cancelled. The python SDK is made aware of that cancellation and cancels the service task associated with that request.

While all that is great because it means that any async tasks spawned by the service task are automatically cancelled, any long-running synchronous code within those tasks are not cancelled. Further, it does not give the async task an opportunity to clean up once it receives the cancellation event.

These methods need a way to see if the task was cancelled and if so, perform the necessary steps to stop long running tasks and clean up (similar to how, in `go`, these functions read `<-ctx.Done()`).  Furthermore, any `await` tasks running within these functions need to not automatically cancel, as that would also prevent any cleanup.

I introduced an `Operation` type that these methods can access to determine if the `operation.is_cancelled()`. The operation knows whether it is cancelled by its enclosing function, which will notify the operation when it receives a cancellation event. 

The enclosing function is a wrapper I created on the `service_base` (`run_with_operation`) that takes in the component's method, any arguments, and calls it while waiting for cancellation events. It also shields any async tasks within that component method so that those tasks aren't automatically cancelled (providing the component the opportunity to check the operation state and cleanup). 

In order for the component methods to receive and read from the operation, it must accept keyword arguments `**kwargs`.

As an example, imagine a motor that wants to go to a specific position. It sets its power, then periodically checks to see if it has arrived at its destination. During those checks, we can see if the operation has ben cancelled and, if so, perform our cleanup tasks. 

```python3
import asyncio
from typing import Any, Dict, Optional

from viam.components.motor import Motor
from viam.operations import Operation

class CustomMotor(Motor):

    async go_to(self, rpm: float, position_revolutions: float, extra: Optional[Dict[str, Any]] = None, **kwargs):
        operation: Operation = kwargs["operation"]
        self.rpm = rpm

        while not self.is_at_destination():
            if await operation.is_cancelled():
                # DO CLEANUP -- stop motor
                self.rpm = 0
                break
            await asyncio.sleep(0.1)

        self.rpm = 0
```

This is currently in draft so that you can see how I've updated Arm and Sensor and review what the end result would look like. If y'all like this current system, I'll add all the other components and also write documentation.

Some annoying things that I currently can't see my way around:

1. Every component method must now also accept `**kwargs` in order to receive an operation
2. You have to explicitly get and type the operation from `**kwargs`
3. Unlike in go where checking `ctx.Done()` is a normal part of the language, the same can't be said in python. So we just have to document and trust that the user will do the right thing (I guess we have to do that as well in go but as it's part of the paradigm, it feels less bad)


stuqdog (approved):
lgtm

njooma:
Closing in favor of https://github.com/viamrobotics/viam-python-sdk/pull/147

PR #147, opened 2022-09-26
[RSDK-490] Operations

njooma:
See https://github.com/viamrobotics/viam-python-sdk/pull/128 for context

**Why the new PR?**
Open sourcing caused forks to get messed up so I had to create a new PR

**What's new?**
The final 2 commits address @edaniels [comment](https://github.com/viamrobotics/viam-python-sdk/pull/128#discussion_r972338874) around making the use of Operations optional. Now, instead of every function automatically injecting an operation and shielding asyncio calls from cancellation, users must explicitly opt-in to operations by decorating any functions with `@run_with_operation`.

```python3
from viam.components.arm import Arm
from viam.operations import run_with_operation

class MyCoolArm(Arm):

    @run_with_operation
    async def move_to_position(self, pose: Pose, world_state: WorldState, extra: Dict[str, Any], **kwargs):
        ...
        operation = self.get_operation(kwargs)
        if await operation.is_cancelled():
            await self.stop()
        ...
```

If a user **does not decorate**, then they do not get an operation and async tasks cancel automatically without providing the opportunity to clean up.


cheukt (approved):
lgtm, this is great - this is a little more verbose but the explicit decoration will give users more control

njooma:
> lgtm, this is great - this is a little more verbose but the explicit decoration will give users more control

Ya I originally assumed (incorrectly) that we wanted to enforce operations

stuqdog (approved):
lgtm

PR #148, opened 2022-09-27
Proto4

njooma:


njooma:
@maximpertsov not sure why the license finder is failing on protobuf when the license hasn't changed in [4.5 years](https://github.com/protocolbuffers/protobuf/blob/main/LICENSE)...

maximpertsov:
> @maximpertsov not sure why the license finder is failing on protobuf when the license hasn't changed in [4.5 years](https://github.com/protocolbuffers/protobuf/blob/main/LICENSE)...

I'm actually surprised it passed the first time - that license you linked isn't one that we explicitly allow-listed. It's permissive enough so we should just add it though.

PR #149, opened 2022-09-28
Edit GetObjectPointClouds Segmenter method

bhaney:
the parameter list was left over from the old segmenter method. This is updated and fixed.


njooma (approved):
Lgtm

PR #151, opened 2022-09-30
Rust and Release

njooma:
Add the rust library, and release!

* Add a new `dial` function that calls out to rust
* Deprecate old `dial_direct` function (still available)
* Update the release workflow to
    * Grab the most recent compiled rust and include it in the package
    * Create a GitHub draft release and tag
* Create a publish workflow to push binaries to PyPI on release published
* Update README to include new installation instructions


npmenard (approved):
LGTM

PR #153, opened 2022-10-03
Remove logging for unimplemented components

njooma:


cheukt:
on `get_component` on a component, can we discern whether the component type itself is unimplemented or if the component is missing and return different errors based on that?

njooma:
If I understand your question correctly, yes there is a way to determine if a component is implemented but not imported, vs a component that is not implemented in python. However, that check is gross and not guaranteed to work (list the subdirectories in the `components` directory and see if the component name exists).

But I don't think it's really necessary. If a user tries to access a component that is implemented, they have to import it, which should automatically register it. If they try to access a component that isn't implemented in python.. well they can't

cheukt (approved):
true, lgtm

PR #157, opened 2022-10-10
One Runtime to Rule them All

njooma:
Only allowed to initialize the rust runtime once, and thus only release it once.


stuqdog (approved):
lgtm

PR #162, opened 2022-10-18
Dial robot in subthread to avoid blocking

biotinker:


stuqdog (approved):
seems reasonable to me! I defer to naveed as the python expert, though.

src/viam/rpc/dial.py
```diff
@@ -222,7 +223,7 @@ async def dial(address: str, options: Optional[DialOptions] = None) -> ViamChann
 
     runtime = _Runtime()
 
-    path, path_ptr = runtime.dial(address, opts)
+    path, path_ptr = await asyncio.to_thread(runtime.dial, address, opts)
```
njooma, line 226: I believe the `runtime.dial` function is not thread safe due to the reverse semaphore not being thread safe. We could make the semaphore (`PointerCounter`) thread safe by adding locks, or we could run the actual blocking part of the function (`_lib.dial`, L191) in `asyncio.to_thread`

biotinker: I think we would want both, no? Adding locks on the semaphore would fix other thread safety issues that have been cropping up like the one with `_ptr` I posted in slack, while `asyncio.to_thread` will fix the blocking issue.

biotinker: Should the lock go on the `_Runtime` object to be used next to the semaphore, or should we add to `PointerCounter` itself?

njooma: I think the lock should go on the PointerCounter object. Something like

```python3
class PointerCounter:
    def __init__(self) -> None:
        self._event = Event()
        self._count = 0
        self._event.set()
        self._lock = Lock() <<<<<<<<

    def increment(self) -> int:
        with self._lock: <<<<<<<<
            self._count += 1
            self._event.clear()
            return self._count

    def decrement(self) -> int:
        with self._lock: <<<<<<<<
            assert self._count > 0, "Pointer count cannot go below zero"
            self._count -= 1
            if self._count == 0:
                self._event.set()
            return self._count
    
    ...
```

biotinker: Yah that's about what I prototyped.

That also means that increment and decrement need to be `async` though

biotinker: The bubbling up of `async` on `PointerCounter` goes all the way to changing `ViamChannel`, and I think also does not fully solve the problem.

I think instead we want a lock on `_Runtime`, as without one, multiple threads opening a connection simultaneously can both hit `if not hasattr(cls, "_shared"):` and then do all sorts of bad things

biotinker:
@njooma Updated again to put a threading lock on the initialization of _Runtime, and also moved the semaphore increment.

Tested against everything I've been doing that's been breaking things so far and have yet to see any problems.

src/viam/rpc/dial.py
njooma: Should we not lock both `_Runtime` and `PointerCounter`?

src/viam/rpc/dial.py
```diff
@@ -160,8 +162,11 @@ class _Runtime:
     _lib: ctypes.CDLL
     _ptr: ctypes.c_void_p
     _semaphore: PointerCounter = PointerCounter()
+    _lock: threading.Lock = threading.Lock()
 
     def __new__(cls):
+        cls._lock.acquire()
+        cls._semaphore.increment()
```
njooma, line 169: Can we increment the semaphore right before releasing the lock? Just in case there are any errors that pop up in between incrementing and returning, we don't want to have the semaphore to have a falsely-inflated count

src/viam/rpc/dial.py
biotinker: I'm okay with this.

Are you okay with me using a threading.Lock on PointerCounter (like with _Runtime) rather than asyncio.Lock, for thread safety?

src/viam/rpc/dial.py
biotinker: If _shared has been initialized, then the if doesn't execute and this is already right before the release.

So if an error happens, it's during the initialization of _shared, so that leaves us with a malformed _shared object, which poisons any future use or future invocations, so the semaphore count is the least of our problems.

But I think we want to increment no matter what. Here's why:

In the case that the above happens and the program manages to survive- if the semaphore isn't incremented and something else comes along and cleans up the memory, then the broken instantiation will cause a panic if it's called. If its semaphore is incremented and its ability to connect is broken, then calling it will merely error, not panic.

src/viam/rpc/dial.py
biotinker: Also to avoid the `async` proliferation that will happen with an asyncio Lock

njooma: Yeah if the reason for doing this is thread safety then let's use a threading lock rather than an asyncio one

src/viam/rpc/dial.py
njooma: Fair point. I also just reviewed the code again, and with us calling `release` if dial fails to return a socket *should* cover all cases where the program doesn't just crash

src/viam/rpc/dial.py
```diff
@@ -160,8 +162,11 @@ class _Runtime:
     _lib: ctypes.CDLL
     _ptr: ctypes.c_void_p
     _semaphore: PointerCounter = PointerCounter()
+    _lock: threading.Lock = threading.Lock()
```
njooma, line 165: Since we are using `_semaphore` and `_lock` as [class variables now](https://github.com/viamrobotics/viam-python-sdk/pull/162/files#diff-e4e6e5a5f62866d78507a238287a5c75651040234a5d11e6ab0ee089acbca90dR168-R169), these should now be 

```python3
from typing import ClassVar

_semaphore: ClassVar[PointerCounter] = PointerCounter()
_lock: ClassVar[threading.Lock] = threading.Lock()
```

(Currently they're instance variables with default values)

src/viam/rpc/dial.py
biotinker: Updated

src/viam/rpc/dial.py
```diff
@@ -216,13 +224,12 @@ def _release(cls):
 
 async def dial(address: str, options: Optional[DialOptions] = None) -> ViamChannel:
     opts = options if options else DialOptions()
+    # ~ await asyncio.sleep(1)
```
njooma, line 227: nit: delete this line

src/viam/rpc/dial.py
```diff
@@ -234,7 +241,7 @@ def release():
         channel = ViamChannel(chan, release)
         return channel
 
-    runtime.release()
+    await runtime.release()
```
njooma, line 244: `runtime.release()` is not an awaitable method

njooma (changes requested):

src/viam/rpc/dial.py
biotinker: Whoops, this is left over from testing async PointerCounter. Sorry

src/viam/rpc/dial.py
biotinker: Done

PR #165, opened 2022-10-20
Rsdk 589/audioinput

njooma:
Audio input!!

Implements audio input in python (except for recording).

Also moves some things around into a media directory. Eventually we'll be able to stream video content natively.


src/viam/components/audio_input/service.py
```diff
@@ -0,0 +1,102 @@
...
+            audio_input = self.get_component(request.name)
+        except ComponentNotFoundError as e:
+            raise e.grpc_error
+        response = (await audio_input.get_properties()).proto
+        await stream.send_message(response)
+
+    async def Record(self, stream: Stream[RecordRequest, HttpBody]) -> None:
+        raise NotSupportedError("Recording audio input is not supported").grpc_error
+
+        # TODO: Eventually implement recording
```
stuqdog, line 60: probably should link this to a specific ticket.

src/viam/media/media.py
```diff
@@ -0,0 +1,48 @@
+from collections.abc import AsyncIterator
+from typing import Protocol, TypeVar
+
+MediaType = TypeVar("MediaType", covariant=True)
+
+
+class MediaStream(Protocol[MediaType]):
+    async def next(self) -> MediaType:
+        ...
```
stuqdog, line 9: just for my own understanding: how does the `...` syntax differ from `Pass`?  And, would it be helpful to have an `unimplemented` message returned rather than not doing anything?

tests/mocks/components.py
```diff
@@ -78,6 +83,29 @@ async def is_moving(self) -> bool:
...
+    def __init__(self, name: str, properties: AudioInput.Properties):
+        super().__init__(name)
+        self.properties = properties
+
+    async def stream(self, **kwargs) -> AudioStream:
+        async def read() -> AsyncIterator[Audio]:
+            for i in range(10):
+                yield Audio(
+                    AudioChunkInfo(
+                        sample_format=SampleFormat.SAMPLE_FORMAT_FLOAT32_INTERLEAVED,
```
stuqdog, line 96: Is it worth testing against other formats as well?

examples/server/v1/components.py
```diff
@@ -78,6 +87,68 @@ async def is_moving(self):
...
+    def __del__(self):
+        self.timer.cancel()
+
+    @run_with_operation
+    async def stream(self, **kwargs) -> AudioStream:
+        operation = self.get_operation(kwargs)
+
+        length = self.sample_rate * self.latency.total_seconds()
+        num_chunks = self.sample_rate / length
+        angle = math.pi * 2 / (length * num_chunks)
```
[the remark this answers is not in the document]
maximpertsov: yeah, +1 for a comment or breaking this out into a function that gives more context on what's happening here

src/viam/components/audio_input/service.py
njooma: Talked with the product team about it -- this is more of a "if people ask for it" kind of reminder. Not something we need to keep track of (yet)

src/viam/media/media.py
njooma: No real difference between `...` and `pass`. Stylistically I prefer `...` because to me it's signals "implementation defined elsewhere," whereas pass feels more like a noop.

tests/mocks/components.py
njooma: Not really since this doesn't actually produce audio, just a fake set of bytes that I can test against

examples/server/v1/components.py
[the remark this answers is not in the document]
njooma: Added comments so that people understand I'm just creating a pure 440Hz tone

stuqdog (approved):
lgtm

cheukt (approved):

PR #169, opened 2022-11-02
[RSDK-199] Add timeouts to components

njooma:
Add timeouts to components and their clients. The `timeout` parameter is one that always has to be named. 

Also changed in this PR is that any `extra` parameters are now also required to be explicitly named.

### DIFF
--- a/src/viam/operations.py
+++ b/src/viam/operations.py
@@ -1,12 +1,10 @@
 import asyncio
 import functools
 import time
-from typing import Any, Callable, Coroutine, TypeVar, cast
+from typing import Any, Callable, Coroutine, Optional, TypeVar, cast
 from uuid import UUID, uuid4
 
 from typing_extensions import Self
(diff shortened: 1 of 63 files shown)


src/viam/operations.py
```diff
@@ -89,10 +93,15 @@ async def wrapper(*args: P.args, **kwargs: P.kwargs) -> T:
         kwarg_names = ", ".join([f"{key}={value}" for (key, value) in kwargs.items()])
         operation = Operation(f"{func_name}({arg_names}{', ' if len(arg_names) else ''}{kwarg_names})", event)
         kwargs[Operation.ARG_NAME] = operation
+        timeout = kwargs.get("timeout", None)
```
cheukt, line 96: add a test for this?

cheukt (approved):
lgtm with a test for operation changes

PR #170, opened 2022-11-03
[RSDK-556] remove width/height from camera.GetImage response and crea…

gvaradarajan:
…te custom RGBA header

NOTE: Only to be released immediately upon merge of https://github.com/viamrobotics/api/pull/84

### DIFF
--- a/src/viam/media/video.py
+++ b/src/viam/media/video.py
@@ -59,23 +59,9 @@ class CameraMimeType(str, Enum):
         if isinstance(image, RawImage):
             return image.data
 
-        if self == CameraMimeType.VIAM_RGBA:
-            data = bytearray()
-            width, height = image.size
-            data.extend(RGBA_MAGIC_NUMBER)
-            data.extend(width.to_bytes(4, byteorder="big"))
-            data.extend(height.to_bytes(4, byteorder="big"))
-            for y in range(height):
-                for x in range(width):
-                    data.extend(image.getpixel((x, y)))
-            return bytes(data)
-        if self == CameraMimeType.JPEG:
+        if self.name in LIBRARY_SUPPORTED_FORMATS:
             buf = BytesIO()
-            image.save(buf, format="JPEG")
-            return buf.getvalue()
-        if self == CameraMimeType.PNG:
-            buf = BytesIO()
-            image.save(buf, format="PNG")
+            image.save(buf, format=self.name)
             return buf.getvalue()
         else:
             raise ValueError(f"Cannot encode image to {self}")
--- a/src/viam/media/video.py
+++ b/src/viam/media/video.py
@@ -1,4 +1,4 @@
-from .viam_rgba_plugin import RGBA_FORMAT_LABEL, RGBA_MAGIC_NUMBER
+from .viam_rgba_plugin import RGBA_FORMAT_LABEL
 from enum import Enum
 from io import BytesIO
 from typing import NamedTuple, Union, Tuple
--- a/src/viam/media/video.py
+++ b/src/viam/media/video.py
@@ -2,6 +2,7 @@ from .viam_rgba_plugin import RGBA_FORMAT_LABEL, RGBA_MAGIC_NUMBER
 from enum import Enum
 from io import BytesIO
 from typing import NamedTuple, Union, Tuple
+from typing_extensions import Self
 from PIL.Image import Image
 
 LAZY_SUFFIX = "+lazy"
@@ -40,7 +41,7 @@ class CameraMimeType(str, Enum):
     UNSUPPORTED = "unsupported"
 
     @classmethod
-    def extract_from_lazy(cls, value: str) -> Tuple["CameraMimeType", bool]:
+    def from_lazy(cls, value: str) -> Tuple[Self, bool]:
         is_lazy = False
         mime_type = value
         if value.endswith(LAZY_SUFFIX):
@@ -49,6 +50,10 @@ class CameraMimeType(str, Enum):
         if not cls.is_supported(value) and not is_lazy:
             mime_type = CameraMimeType(CameraMimeType.UNSUPPORTED)
         return (cls(mime_type), is_lazy)
+
+    @property
+    def with_lazy_suffix(self) -> str:
+        return f"{self.value}{LAZY_SUFFIX}"
 
     def encode_image(self, image: Union[Image, RawImage]) -> bytes:
         if isinstance(image, RawImage):
--- a/src/viam/media/video.py
+++ b/src/viam/media/video.py
@@ -1,15 +1,24 @@
+from .viam_rgba_plugin import RGBA_FORMAT_LABEL, RGBA_MAGIC_NUMBER
 from enum import Enum
 from io import BytesIO
-from typing import NamedTuple, Union
-
+from typing import NamedTuple, Union, Tuple
 from PIL.Image import Image
+
+LAZY_SUFFIX = "+lazy"
+
+# Formats that are supported by PIL
+LIBRARY_SUPPORTED_FORMATS = ["JPEG", "PNG", RGBA_FORMAT_LABEL]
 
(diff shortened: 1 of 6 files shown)


src/viam/media/video.py
```diff
@@ -18,26 +27,43 @@ class RawImage(NamedTuple):
...
 class CameraMimeType(str, Enum):
-    RAW = "image/vnd.viam.rgba"
+    VIAM_RGBA = "image/vnd.viam.rgba"
     JPEG = "image/jpeg"
     PNG = "image/png"
     PCD = "pointcloud/pcd"
+    UNSUPPORTED = "unsupported"
+
+    @classmethod
+    def extract_from_lazy(cls, value: str) -> Tuple["CameraMimeType", bool]:
```
njooma, line 43: Should we have a corresponding `@property` method to create a lazy mime type from an existing one? That way people can request a lazy image

Or should we have separate lazy enum members, e.g. `PNG_LAZY = "image/png+lazy"`?

njooma (commented):
Just a few questions/nits, but great overall

src/viam/media/video.py
```diff
@@ -18,26 +27,43 @@ class RawImage(NamedTuple):
...
+        is_lazy = False
+        mime_type = value
+        if value.endswith(LAZY_SUFFIX):
+            mime_type = value[: (len(value) - len(LAZY_SUFFIX))]
+            is_lazy = True
+        if not cls.is_supported(value) and not is_lazy:
```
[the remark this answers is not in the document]
gvaradarajan: I think the mime type matters for when you want to encode the data back into its requested format.

src/viam/media/video.py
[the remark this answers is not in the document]
gvaradarajan: ah, my Python knowledge was out of date, thank you

src/viam/media/video.py
```diff
@@ -18,26 +27,43 @@ class RawImage(NamedTuple):
...
+            mime_type = CameraMimeType(CameraMimeType.UNSUPPORTED)
+        return (cls(mime_type), is_lazy)
 
     def encode_image(self, image: Union[Image, RawImage]) -> bytes:
         if isinstance(image, RawImage):
             return image.data
+
+        if self == CameraMimeType.VIAM_RGBA:
+            data = bytearray()
```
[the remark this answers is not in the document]
gvaradarajan: uh, well, tbh I tried to use the Image.save pattern and it still didn't work. But I can give it another shot

src/viam/media/video.py
gvaradarajan: I think I would rather do the first one, but yeah some solution for this is necessary and I missed it

src/viam/media/video.py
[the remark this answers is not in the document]
gvaradarajan: got it to work

njooma (approved):
Nice!

src/viam/media/video.py
```diff
@@ -1,15 +1,25 @@
...
     """A raw bytes representation of an image.
 
-    When requesting `CameraMimeType.RAW` or if the camera returns a mime type that is not supported,
-    the returned type will be this RawImage type.
+    A RawImage should be returned instead of a PIL Image instance under one of
+    the following conditions
+
+    1) The requested mime type has the LAZY_SUFFIX string appended to it
+    2) The requested mime type is not supported for decoding/encoding by Viam's
+    Python SDK
```
bhaney, line 22: From what I understand, LAZY is only for while in transit. If your python SDK is the final destination, it should probably decoded, even if its LAZY. @njooma  Will there be times when the python SDK will be an intermediate hop and be used as a server for a camera as well? i.e. python will be sending images it has received from another server.

tests/mocks/components.py
```diff
@@ -317,12 +317,11 @@ def __init__(self, name: str):
         super().__init__(name)
 
     async def get_image(self, mime_type: str = CameraMimeType.PNG, **kwargs) -> Union[Image.Image, RawImage]:
-        if not CameraMimeType.is_supported(mime_type) or mime_type == CameraMimeType.RAW:
+        mime_type, is_lazy = CameraMimeType.from_lazy(mime_type)
+        if is_lazy or (not CameraMimeType.is_supported(mime_type)):
             return RawImage(
                 data=self.image.convert("RGBA").tobytes("raw", "RGBA"),
```
bhaney, line 323: [q] the call to convert-- will this end up decoding the image anyway, just to convert to RGBA and then put it back into bytes? 

If that's the case, it defeats the purpose of lazy, which is about not touching the bytes it receives and just passing them along to the next computer that needs it.

tests/mocks/components.py
```diff
@@ -317,12 +317,11 @@ def __init__(self, name: str):
...
                 mime_type=mime_type,
-                width=self.image.width,
-                height=self.image.height,
             )
         return self.image.copy()
```
bhaney, line 326: Does the copy slow us down? Why can't it return the image directly?

src/viam/media/video.py
```diff
@@ -18,33 +28,40 @@ class RawImage(NamedTuple):
...
+        if value.endswith(LAZY_SUFFIX):
+            mime_type = value[: (len(value) - len(LAZY_SUFFIX))]
+            is_lazy = True
+        if not cls.is_supported(value) and not is_lazy:
+            mime_type = CameraMimeType(CameraMimeType.UNSUPPORTED)
+        return (cls(mime_type), is_lazy)
+
+    @property
+    def with_lazy_suffix(self) -> str:
+        return f"{self.value}{LAZY_SUFFIX}"
```
bhaney, line 56: [q] is there a chance that value might already the lazy suffix on it?

src/viam/media/video.py
```diff
@@ -18,33 +28,40 @@ class RawImage(NamedTuple):
...
     def encode_image(self, image: Union[Image, RawImage]) -> bytes:
         if isinstance(image, RawImage):
             return image.data
-        if self == CameraMimeType.JPEG:
-            buf = BytesIO()
-            image.save(buf, format="JPEG")
-            return buf.getvalue()
-        if self == CameraMimeType.PNG:
+
+        if self.name in LIBRARY_SUPPORTED_FORMATS:
```
bhaney, line 62: [q] there is no chance that self.name will have `lazy` appended to it?

bhaney (commented):
A few questions about the implementation, but it's looking good!

src/viam/media/video.py
bhaney: Is there is a chance that python could be a server, then what's you're doing is right Gautham, and anything with the lazy suffix should just return the raw bytes

tests/mocks/components.py
bhaney: maybe more a question for @njooma

src/viam/media/video.py
njooma: I don't see why a user couldn't use the python-sdk as a camera server -- it definitely has that capability to provide images as well as consume them

tests/mocks/components.py
njooma: This is just the test case -- but yes if someone were to do this then it would defeat the purpose of lazy.

tests/mocks/components.py
njooma: Again, this is just in the tests, so it's moot

src/viam/media/video.py
njooma: No -- the only values allowed are the ones defined above. These enums are not extendible, so there's no risk of that

src/viam/media/video.py
njooma: Similarly to above: nope! Name can only be as we define it, so as long as we don't define a lazy member, it can't happen

bhaney (approved):
All questions answered! LGTM

PR #173, opened 2022-11-09
RSDK-806 - gripper extra params

stuqdog:


src/viam/components/gripper/client.py
```diff
@@ -24,17 +25,23 @@ def __init__(self, name: str, channel: Channel):
         self.client = GripperServiceStub(channel)
         super().__init__(name)
 
-    async def open(self, *, timeout: Optional[float] = None):
-        request = OpenRequest(name=self.name)
+    async def open(self, *, timeout: Optional[float] = None, extra: Optional[Dict[str, Any]] = None):
```
njooma, line 28: It's not strictly necessary in the language, but for consistency with the other components, let's have `extra` come BEFORE `timeout`:
```python3
async def open(self, *, extra: ..., timeout: ...):
```

src/viam/components/gripper/gripper.py
```diff
@@ -15,14 +15,26 @@ class Gripper(ComponentBase):
     """
 
     @abc.abstractmethod
-    async def open(self, *, timeout: Optional[float] = None, **kwargs):
+    async def open(
+        self,
+        *,
+        timeout: Optional[float] = None,
+        extra: Optional[Dict[str, Any]] = None,
```
njooma, line 22: Similar to the client, let's have `extra` come before `timeout`

njooma (changes requested):
solid first pass!

njooma (approved):
Nice

PR #176, opened 2022-11-10
RSDK-812-817: Extra params for Services

cheukt:


src/viam/services/sensors.py
```diff
@@ -23,17 +23,21 @@ def __init__(self, name: str, channel: Channel):
         self.client = SensorsServiceStub(channel)
         self.name = name
 
-    async def get_sensors(self) -> List[ResourceName]:
+    async def get_sensors(self, extra: Optional[Mapping[str, Any]] = None) -> List[ResourceName]:
```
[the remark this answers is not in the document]
cheukt: done

src/viam/services/vision.py
```diff
@@ -134,52 +155,73 @@ async def get_detections(self, image: Union[Image, RawImage], detector_name: str
             confidence score of the labels, around the found objects in the next 2D image
             from the given camera, with the given detector applied to it.
         """
+        if extra is None:
+            extra = {}
         mime_type = CameraMimeType.JPEG
+        if isinstance(image, RawImage):
```
cheukt, line 161: @njooma these linting errors should've been found somehow with the earlier camera changes - is there anything we can do to make the linting more robust?

njooma: We can have much stricter mypy policies as part of our test suite, which should catch these statically typed issues

njooma: I'll add that to my todos

src/viam/services/vision.py
```diff
@@ -1,10 +1,11 @@
 import json
 from dataclasses import dataclass
 from enum import Enum
+from io import BytesIO
 from typing import Any, List, Mapping, Optional, Sequence, Union
 
 from grpclib.client import Channel
-from PIL.Image import Image
+from viam.media.viam_rgba_plugin import Image
```
njooma, line 8: This is just an alias for `from PIL import Image`. Not sure why you needed to import this (and then change all the `Image` definitions to `Image.Image`...

cheukt: @gvaradarajan says we need it to support the viam raw format (otherwise it isn't registered?)

njooma: Yuuup I see

njooma (approved):

src/viam/services/vision.py
cheukt: thanks!

PR #181, opened 2022-11-15
[DATA-665] Change getImage default to JPEG

kharijarrett:
Small change in code, big change in quality of life. Please bless my people with default JPEG encoding. They have suffered enough under the PNG reign. Added benefit of not breaking anything. <3

### DIFF
--- a/src/viam/components/camera/camera.py
+++ b/src/viam/components/camera/camera.py
@@ -31,7 +31,7 @@ class Camera(ComponentBase):
         """The distortion parameters of the camera"""
 
     @abc.abstractmethod
-    async def get_image(self, mime_type: str = CameraMimeType.PNG, *, timeout: Optional[float] = None, **kwargs) -> Union[Image, RawImage]:
+    async def get_image(self, mime_type: str = CameraMimeType.JPEG, *, timeout: Optional[float] = None, **kwargs) -> Union[Image, RawImage]:
         """Get the next image from the camera as an Image or RawImage.
         Be sure to close the image when finished.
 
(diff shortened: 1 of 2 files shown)


cheukt (commented):
might want to update the ExampleCamera (in examples/server/v1/components.py) and MockCamera (in tests/mocks/components.py) for consistency

kharijarrett:
Just changed the ExampleCamera and MockCamera too. Thanks for the heads up!

njooma (approved):

cheukt:
just make sure the tests are passing - they're failing right now

kharijarrett:
Hi, I've not (completely) forgotten about this PR! I'm just doing something in RDK (sending along streamType) simultaneously to make the transition smoother for people who might be using depth images.

kharijarrett:
I know what I said earlier but I wanted to just comment with some more recent info. The goal here is to send the mimeType decision to the camera server (RDK) side. I'm setting the default mimeType here to an empty string, and on the RDK side if you get an empty string (no explicitly requested mimetype), there's logic to follow based on ya camera image_type (depth, color, etc.). Still wanna push this at/around the same time as RDK changes (sooooon)

cheukt (approved):

njooma (commented):
@kharijarrett do we want to implement those RDK changes here in the python SDK as well in case the camera server is in python and not in the RDK?

kharijarrett:
Probably not tbh... If a user doesn't see the RDK, that means they are writing their own camera server and client in Python. I think at that point, they oughta be capable of choosing their own default mimetype

njooma:
So let the user create a pipeline for what to do if the mime type is empty? I'm ok with that

PR #184, opened 2022-11-17
allow viamrobotics members to release

cheukt:
was gonna try restricting to our team only, but kinda annoying to do.

can use https://github.com/garnertb/get-team-members, but then have to to install `gh` and `jq` onto the image, and I haven't found an image with both preinstalled

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -28,24 +28,18 @@ jobs:
       sha: ${{ steps.commit.outputs.commit_long_sha }}
       version: ${{ steps.bump_version.outputs.version }}
     steps:
-      - name: Install the gh cli
-        uses: dev-hanz-ops/install-gh-cli-action@v0.1.0
-
-      - name: Install jq
-        run: apt install jq -y
-
-      - name: Get SDK/NetCode team members
-        id: get-members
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -20,14 +20,27 @@ on:
...
-    if: github.repository_owner == 'viamrobotics' && github.ref == 'refs/heads/main' && contains(fromJson('["njooma"]'), github.actor)
+    if: github.repository_owner == 'viamrobotics' && github.ref == 'refs/heads/main')
     runs-on: [self-hosted, x64]
     container:
       image: ghcr.io/viamrobotics/canon:amd64
     outputs:
       sha: ${{ steps.commit.outputs.commit_long_sha }}
       version: ${{ steps.bump_version.outputs.version }}
     steps:
+      - name: Check if organization member
```
njooma, line 31: Is there no way to check if they're a team member, rather than an org member?

cheukt: updated to check team member

njooma (approved):

.github/workflows/release.yml
cheukt: update to update - reverted to this because github workflows evaluate weird

PR #192, opened 2022-12-08
RSDK-872: Log to stdout, errors to stderr

cheukt:
sends errors to stderr and regular logs to stdout.

follows how we do it in the rdk


njooma (approved):
nice.

PR #198, opened 2022-12-19
RSDK-623: remove position_reporting from MotorStatus

stevebriskin:
Rebuild with the latest protos, which 1. add IsMoving to a few components and 2. remove position_reporting from Motor.Status.

For #1, I'm unsure if more work must be done since is_moving currently returns an unimplemented error.
For #2, only minor changes were required.


cheukt (approved):
we should make the `is_moving` an abstract method on the components, but I would be ok with making it a neweng ticket and add tests for it then to limit scope of this pr.

PR #199, opened 2022-12-19
Update Version

Kschappacher:


maximpertsov (approved):

njooma (changes requested):
NOT NEEDED. Please use the GitHub actions to do this, do not do it manually

njooma:
This PR is not needed. In order to create a release for python, please use the GitHub action `release` and use the `patch` parameter in the manual run dropdown menu. This will also create all the release binaries, as well as a draft GitHub release. 

Then go to the release that was just created, auto-generate the release notes, and then hit release. Everything else is automatic.

PR #202, opened 2023-01-13
RSDK-857 - add pcd docs

cheukt:
Adds an example to vision.get_object_point_clouds() and camera.get_point_clouds() so users can more easily use the returned point clouds

### DIFF
--- a/src/viam/components/camera/camera.py
+++ b/src/viam/components/camera/camera.py
@@ -52,6 +52,20 @@ class Camera(ComponentBase):
         should encode the bytes into the formatted suggested
         by the mimetype.
 
+        To deserialize the returned information into a numpy array, use the Open3D library.
+        ::
+
+            import numpy as np
+            import open3d as o3d
+
+            data, _ = await camera.get_point_cloud()
+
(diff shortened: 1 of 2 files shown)


src/viam/components/camera/camera.py
```diff
@@ -52,6 +52,20 @@ async def get_point_cloud(self, *, timeout: Optional[float] = None, **kwargs) ->
...
 
+        To deserialize the returned information into a numpy array, use the Open3D library.
+        ::
+
+            import numpy as np
+            import open3d as o3d
+
+            data, _ = await camera.get_point_cloud()
+
+            # write the point cloud into a temporary file
```
njooma, line 63: You can probably use a `BytesIO` buffer rather than writing out to a file in case there are permissions issues with writing to disk

njooma (approved):

src/viam/components/camera/camera.py
cheukt: unfortunately the api only takes file name as input arg https://github.com/isl-org/Open3D/issues/1146

PR #204, opened 2023-01-19
RSDK-670: update docs around creating resource names

cheukt:
some updates around those docs.

also update get_pose so that we don't pass in [] for the optional parameter (just for safety reasons)

### DIFF
--- a/src/viam/services/motion.py
+++ b/src/viam/services/motion.py
@@ -41,12 +41,11 @@ class MotionServiceClient(ServiceClientBase):
         """Plan and execute a movement to move the component specified to its goal destination.
 
         Note: Frames designated with respect to components can also be used as the ``component_name`` when calling for a move. This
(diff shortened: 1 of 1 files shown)


spieswl (approved):
LGTM!

src/viam/services/motion.py
```diff
@@ -109,13 +108,19 @@ async def get_pose(
...
         extra: Optional[Mapping[str, Any]] = None,
         timeout: Optional[float] = None,
     ) -> PoseInFrame:
         """
-        Get the Pose and observer frame for any given component on a robot.
+        Get the Pose and observer frame for any given component on a robot. A ``component_name`` can be created like this::
+
+            component_name = Arm.get_resource_name("arm")
+
+        Note that the example uses the Arm class, but any component class that inherits from ``ComponentBase`` will work
```
njooma, line 121: Back ticks around `Arm`

njooma (approved):

PR #209, opened 2023-02-02
[RSDK-1822] Subtypes and ResourceNames

njooma:
The associated ticket number is actually for adding module support to the python SDK, but a bunch of preliminary work had to happen first to make it possible. Because this prelim work is actually quite large, I'm making a PR now so that we don't have to review a super big PR when everything is ready. 

So this PR focuses on a couple of things:
1. Adding the module protos and its proto requirements
2. Adding `Subtype` support in python. This was ignored the first time around because it wasn't really needed back then. But because we're now adding modules, modules will need their own subtypes. 
3. Reworking internals to use ResourceName everywhere. We can't assume that resource names will always be `rdk/{component|service}/{type}/{name}` anymore with the introduction of modules. So move things internally from `name: str` to `name: ResourceName`


src/viam/errors.py
```diff
@@ -53,15 +53,15 @@ def grpc_error(self) -> GRPCError:
         return GRPCError(self.grpc_code, self.message)
 
 
-class ComponentNotFoundError(ViamGRPCError):
+class ResourceNotFoundError(ViamGRPCError):
     """
     Exception raised when a component is not found in the registry
```
cheukt, line 58: component -> resource

src/viam/resource/types.py
```diff
@@ -0,0 +1,44 @@
...
+    """The subtype of the component e.g. `servo`, `arm`, `vision`
+    """
+
+    def __init__(self, namespace: str, resource_type: str, resource_subtype: str):
+        self.namespace = namespace.lower()
+        self.resource_type = resource_type.lower()
+        self.resource_subtype = resource_subtype.lower()
+
+    def __str__(self) -> str:
+        return f"{self.namespace}/{self.resource_type}/{self.resource_subtype}"
```
cheukt, line 31: should be `:` instead of `/`

src/viam/resource/registry.py
```diff
@@ -6,9 +6,11 @@
 
 from viam.components.component_base import ComponentBase
 from viam.components.service_base import ComponentServiceBase
-from viam.errors import ComponentNotFoundError, DuplicateComponentError
+from viam.errors import DuplicateComponentError, ResourceNotFoundError
```
cheukt, line 9: might make sense to also update to `DuplicateResourceError` now

src/viam/resource/types.py
```diff
@@ -0,0 +1,44 @@
...
+    namespace: str
+    """The namespace of the resource
+    """
+
+    resource_type: str
+    """The type of the resource, e.g. `component` or `service`
+    """
+
+    resource_subtype: str
+    """The subtype of the component e.g. `servo`, `arm`, `vision`
```
cheukt, line 22: component -> resource

src/viam/utils.py
```diff
@@ -77,23 +77,16 @@ def value_to_primitive(value: Value) -> Any:
 
 def resource_names_for_component(component: ComponentBase) -> List[ResourceName]:
```
cheukt, line 78: somewhat unrelated to this PR but I'm pretty confused about what this function does - don't we only have one resource name per resource? and we can construct it with `component.get_resource_name(component.name)`?

cheukt (changes requested):
generally looks good, few cosmetic changes requested and an unrelated question

src/viam/utils.py
njooma: Components can technically have multiple resource names -- say I have a custom component that implements both Servo and Gripper, it could theoretically have 2 resource names (I should be able to get it as either `rdk:component:servo:custom` or `rdk:component:gripper:custom`

cheukt: ah yes I remember this now :)

cheukt (approved):
LGTM!

maximpertsov (approved):

PR #210, opened 2023-02-06
RSDK 691 concise way to call remote components

purplenicole730:


tests/test_resource_manager.py
```diff
@@ -58,3 +58,28 @@ def test_not_found(self):
...
+        assert component.name == "remote2:arm1"
+        component = manager.get_component(MockArm, MockArm.get_resource_name("remote2:arm1"))
+        assert component.name == "remote2:arm1"
+
+        with pytest.raises(ResourceNotFoundError):
+            component = manager.get_component(MockArm, MockArm.get_resource_name("arm"))
+        component = manager.get_component(MockArm, MockArm.get_resource_name("remote1:arm"))
+        assert component.name == "remote1:arm"
+        component = manager.get_component(MockArm, MockArm.get_resource_name("remote2:arm"))
+        assert component.name == "remote2:arm"
```
cheukt, line 85: would like a test case with remote of remotes (so `remote1:remote2:arm` and `remote1:remote2:remote3:arm` for example)

cheukt (changes requested):
looks pretty good, would like to see another test or two

src/viam/components/resource_manager.py
```diff
@@ -39,6 +42,12 @@ def register(self, component: ComponentBase):
             if hasattr(subtype, "get_resource_name"):
                 rn = subtype.get_resource_name(component.name)  # type: ignore
                 rnames[rn] = component
+            if rn.name.__contains__(":"):
+                short_name = rn.name.split(":")[-1]
+                if short_name in self._short_to_long_name and rn not in self._short_to_long_name[short_name]:
```
njooma, line 47: On line 17 you specify the type of `_short_to_long_name` as `Dict[str, ResourceName]` but here, you're using it as `Dict[str, List[ResourceName]]`. Please update the typedef

src/viam/components/resource_manager.py
```diff
@@ -64,6 +74,9 @@ def get_component(self, of_type: Type[ResourceType], name: ResourceName) -> Reso
         if component and isinstance(component, of_type):
             return component
 
+        if name.name in self._short_to_long_name and len(self._short_to_long_name[name.name]) == 1:
```
njooma, line 77: Ok this makes me think we shouldn't be using a `List` at all to store the `ResourceName`s. Because if there is a collision (multiple of the same short names), then we store the additional ResourceNames. But down here, if the list is greater than 1, we ignore the short name. Then what's the point of storing it?

I think that you if you encounter a collision when creating the dictionary (line 47), then you should simply delete the entry from the dictionary itself and pretend like it was never there.

njooma (changes requested):
Looks good overall! A few changes:
1. There's a type mismatch for the `_short_to_long_name` dictionary
2. I think storing the ResourceNames as an array inside of `_short_to_long_name` is not necessary and we can probably streamline it by deleting the entry on collision

src/viam/components/resource_manager.py
cheukt: we should store something because if there were 3 of the same name, we also shouldn't have an entry. We could go back and delete the entry at the end, but we shouldn't delete during the process of creating the dict.

njooma: Big brain

src/viam/components/resource_manager.py
purplenicole730: Thanks for catching that!

src/viam/components/resource_manager.py
purplenicole730: I was also wondering what the best method for this was, since I was thinking the same thing as Cheuk. I was tempted to delete the multiple entries at the end of `ResourceManager.__init__()`, but this would make it so that when `register()` is called separately elsewhere, then `_short_to_long_name()` will no longer be accurate.
Is there another smarter way to do this?

src/viam/components/resource_manager.py
```diff
@@ -14,16 +14,19 @@ class ResourceManager:
...
         self.components = {}
+        self._short_to_long_name = {}
         for component in components:
             self.register(component)
 
     def register(self, component: ComponentBase):
         """
         Register a new component with the registry.
         Components may not have the same name.
+        If a component is remote, save a short name version.
```
cheukt, line 29: nit: update desc to `If a component is remote and the short name is unique`

src/viam/components/resource_manager.py
cheukt: we could just make `register` an internal method `_register` and hope people don't register anything on their own. We can't guarantee anything in Python haha.
The Go SDK implementation uses a second map (dupMap) which tracks the keys to delete at the end, so that we can make  `_short_to_long_name` `Dict[str,ResourceName]`, which might look and feel better.
I think either works

purplenicole730: I definitely agree that `Dict[str, ResourceName]` looks way better, but we already use `register` elsewhere, in `RobotClient.refresh()`.

purplenicole730: I definitely agree that Dict[str, ResourceName] looks way better, but we already use register elsewhere, in RobotClient.refresh().

cheukt: oh ya I missed that

cheukt (approved):
LGTM!

src/viam/components/resource_manager.py
```diff
@@ -39,6 +42,12 @@ def register(self, component: ComponentBase):
             if hasattr(subtype, "get_resource_name"):
                 rn = subtype.get_resource_name(component.name)  # type: ignore
                 rnames[rn] = component
+            if ":" in rn.name:
```
njooma, line 45: It's entirely possible that `rn` here doesn't exists (since it's created above within an `if` statement). 

It's probably better to move this to a separate `for` loop once the `rnames` dict is done populating.

src/viam/components/resource_manager.py
```diff
@@ -64,6 +74,9 @@ def get_component(self, of_type: Type[ResourceType], name: ResourceName) -> Reso
         if component and isinstance(component, of_type):
             return component
 
+        if name.name in self._short_to_long_name and len(self._short_to_long_name[name.name]) == 1:
+            component = self.components.get(self._short_to_long_name[name.name][0], None)
+            return component
```
njooma, line 79: `component` here could be `None`. You should either

1. Check that `component` is not `None` before returning OR
2. Return the result of recursively calling this function, but with the correct `ResourceName` (`return self.get_component(of_type, self._short_to_long_name[name.name][0])`)

njooma (changes requested):
Getting close! Some small changes

njooma (approved):

PR #215, opened 2023-02-21
Robot reconnect logic

stuqdog:
Adds reconnect logic to a python robot. This is necessary for webRTC connections, which unlike gRPC will not automatically attempt to connect.


src/viam/robot/client.py
```diff
@@ -119,26 +130,38 @@ async def with_channel(cls, channel: Union[Channel, ViamChannel], options: Optio
...
         self._should_close_channel = False
+        self._options = options
         await self.refresh()
 
         if options.refresh_interval > 0:
             self._refresh_task = asyncio.create_task(
                 self._refresh_every(options.refresh_interval), name=f"{viam._TASK_PREFIX}-robot_refresh_metadata"
             )
 
+        if options.check_connected_time > 0 and options.reconnect_every > 0:
```
stuqdog, line 147: @dannenberg @njooma this is a bit opinionated in that, if a user sets either of these to `0`, then they will continue to see the same behavior as before. Only when both are set will errors be immediately seen. It's based on the `Go` logic, but I'm happy to see this changed as appropriate.

src/viam/robot/client.py
```diff
@@ -173,6 +196,52 @@ async def _refresh_every(self, interval: int):
...
+    async def _check_connection(self, check_every: int, reconnect_every: int):
+        while True:
+            if self._connected:
+                wait_time = check_every
+            else:
+                wait_time = reconnect_every
+
+            await asyncio.sleep(wait_time)
+
+            if not self._address:
```
stuqdog, line 208: @njooma @dannenberg unfortunately we rely on knowing the address in order to reconnect, so a robot that was created directly through a call to `with_channel` would never be able to reconnect. I suspect this is fine since most use cases are going to be connecting via address, but wanted to flag it.

njooma:
@stuqdog made some changes, but I can't request you as a reviewer since you made the original PR. So if you want to take a look and leave comments yourself, feel free!

src/viam/robot/client.py
njooma: I changed the logic such that we always check for connection and log and at minimum log any errors. Then if enabled, also try and reconnect

stuqdog:
> @stuqdog made some changes, but I can't request you as a reviewer since you made the original PR. So if you want to take a look and leave comments yourself, feel free!

I am not allowed to officially review the PR but lgtm! 💸

njooma (approved):

PR #220, opened 2023-02-24
[RSDK-1822] Modular Components

njooma:
This PR enables Modules, but only for components thus far. Services will come in another PR later.

It includes a lot of renames (because everything used to be named for components instead of resources), and also some new fields around Subtypes and Models. 

It also includes a new example for Modules, with its own README to help people get started.


src/viam/resource/types.py
```diff
@@ -41,4 +51,130 @@ def __eq__(self, other: "Subtype") -> bool:
...
+
+
+def resource_name_from_string(string: str) -> ResourceName:
+    parts = string.split(":")
+    if len(parts) < 4:
+        raise ValueError(f"{string} is not a valid ResourceName")
+    return ResourceName(namespace=parts[0], type=parts[1], subtype=parts[2], name=":".join(parts[3:]))
+
+
+ComponentCreator: TypeAlias = Callable[[Mapping[ResourceName, "ComponentBase"], ComponentConfig], "ComponentBase"]
```
cheukt, line 180: Can we name this ResourceCreator to limit the amount of future changes we have to make?

cheukt (commented):
Looks great! the example is great and easy to test out. the few comments I have are around naming and perhaps typing so that we don't end up having to duplicate work too much when adding support for services. The RDK has associated tech debt which makes it hard to combine components and services easily, but if we could do that as much as possible now, I'd imagine our work later with be reduced

src/viam/resource/types.py
njooma: The reason I called it `ComponentCreator` for now was because I haven't yet determined what changes I'll ned to make to convert this to an actual `ResourceCreator` (specifically since the input and return types are still `ComponentBase` and not a nonexistent `ResourceBase`)

It's kind of like a breadcrumb/clue in the code that it needs more thought and updating

njooma: Also because this type is used internally (users don't ever have to name this type specifically, just conform to it), renaming this later shouldn't be breaking

cheukt (approved):

PR #222, opened 2023-02-27
RSDK 1982 Add DoCommand to components

purplenicole730:
Components already had a `do_command` function, but routed to `generic`. 

The changes introduced in this PR:
- Rerouted do_command() from `generic` to `protocol.common` for each component
- Add DoCommand to component service files
- Add service tests to each component

### DIFF
--- a/src/viam/components/arm/arm.py
+++ b/src/viam/components/arm/arm.py
@@ -112,13 +112,3 @@ class Arm(ComponentBase):
             bool: Whether the arm is moving.
         """
         ...
-
-    @abc.abstractmethod
-    async def do_command(self, command: Dict[str, Any], *, timeout: Optional[float] = None, **kwargs) -> Dict[str, Any]:
-        """
-        Receive arbitrary commands
-
-        Returns:
-            Dict[str, Any]: Arbitrary response
-        """
-        ...
--- a/src/viam/components/arm/arm.py
+++ b/src/viam/components/arm/arm.py
@@ -112,3 +112,13 @@ class Arm(ComponentBase):
             bool: Whether the arm is moving.
         """
         ...
+
+    @abc.abstractmethod
+    async def do_command(self, command: Dict[str, Any], *, timeout: Optional[float] = None, **kwargs) -> Dict[str, Any]:
+        """
+        Receive arbitrary commands
+
+        Returns:
+            Dict[str, Any]: Arbitrary response
+        """
+        ...
--- a/tests/mocks/components.py
+++ b/tests/mocks/components.py
@@ -104,6 +104,9 @@ class MockArm(Arm):
     async def is_moving(self) -> bool:
         return not self.is_stopped
 
+    async def do_command(self, command: Dict[str, Any], *, timeout: Optional[float] = None, **kwargs) -> Dict[str, Any]:
+        return {"hello": "world"}
+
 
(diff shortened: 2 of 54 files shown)


tests/mocks/components.py
```diff
@@ -104,6 +104,9 @@ async def stop(self, *, extra: Optional[Dict[str, Any]] = None, timeout: Optiona
     async def is_moving(self) -> bool:
         return not self.is_stopped
 
+    async def do_command(self, command: Dict[str, Any], *, timeout: Optional[float] = None, **kwargs) -> Dict[str, Any]:
+        return {"hello": "world"}
```
njooma, line 108: I would make all the return something like `{"command": command}`, that way you can test whether commands are being successfully sent from `service -> component` and `client -> service -> component`

tests/test_arm.py
```diff
@@ -66,8 +65,8 @@ async def test_is_moving(self):
 
     @pytest.mark.asyncio
     async def test_do(self):
-        with pytest.raises(NotImplementedError):
-            await self.arm.do_command({"command": "args"})
+        resp = await self.arm.do_command({"command": "args"})
+        assert resp == {"hello": "world"}
```
njooma, line 69: If you follow my suggestion in `mocks.components`, then the assert should now be `assert responsibility == {"command": {"command": "args"}}`

You could of course save the command we pass as a parameter as a var:

```python3
command = {"command": "args"}
resp = await self.arm.do_command(command)
assert resp == {"command": command}
```

src/viam/components/arm/arm.py
```diff
@@ -112,3 +112,13 @@ async def is_moving(self) -> bool:
             bool: Whether the arm is moving.
         """
         ...
+
+    @abc.abstractmethod
+    async def do_command(self, command: Dict[str, Any], *, timeout: Optional[float] = None, **kwargs) -> Dict[str, Any]:
```
njooma, line 117: This technically doesn't do anything, because the `ComponentBase` class (which all components inherit from) already has a `do_command` function that has a default implementation (it raises a `NotImplementedError` that you then have to override if you want to implement a `do`).

njooma: This comment is applicable to all components

njooma (changes requested):
Just a couple of ideas for changes. Open to discussion!

src/viam/components/arm/arm.py
benjirewis: Should we just remove `do_command` from `ComponentBase`?

benjirewis (approved):
LGTM % Naveed's comments. Great work! 🧑‍🔧

src/viam/components/arm/arm.py
njooma: If we remove it from `ComponentBase`, this becomes a breaking change -- any existing custom components that rely on the default `NotImplemented` version of `do` would fail to load because they wouldn't implement all the abstract methods of this class.

Also, because *every* component should implement `do`, having it in `ComponentBase` means we reduce repetition.

And this allows modular components to opt-in to `do` easily, since all components (including modular ones) must inherit from `ComponentBase`. 

Those are just my reasons for keeping it in `ComponentBase`, but if anyone has strong feelings as to putting them in each individual component, I'm all ears

cheukt (approved):
LGTM

PR #225, opened 2023-03-01
proxy asyncio.to_thread

njooma:
using `asyncio.to_thread` was the only function that was not python 3.8 compatible. It turns out that this function is literally just some sugar that wraps the code I've added directly

https://github.com/python/cpython/blob/main/Lib/asyncio/threads.py

### DIFF
--- a/src/viam/utils.py
+++ b/src/viam/utils.py
@@ -3,7 +3,6 @@ import contextvars
 import functools
 import sys
 import threading
-from asyncio import Event
 from typing import Any, Dict, List, Mapping, SupportsFloat, Type, TypeVar
 
 from google.protobuf.json_format import MessageToDict, ParseDict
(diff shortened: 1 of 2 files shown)


src/viam/utils.py
```diff
@@ -184,3 +188,12 @@ async def wait(self) -> None:
...
             return self._count
+
+
+async def to_thread(func, /, *args, **kwargs):
+    if sys.version_info >= (3, 9):
+        return await asyncio.to_thread(func, *args, **kwargs)
+    loop = asyncio.events.get_running_loop()
+    ctx = contextvars.copy_context()
+    func_call = functools.partial(ctx.run, func, *args, **kwargs)
+    return await loop.run_in_executor(None, func_call)
```
maximpertsov, line 199: I might be helpful to add a comment in code with a link to this logic as you did the PR

maximpertsov (approved):

PR #226, opened 2023-03-01
RSDK 850 specify docommand value types

purplenicole730:


njooma:
So I know this is currently a draft, but I wanted to leave a comment before you went any further down this path.

The reason we're changing these from `Any` to explicit types is actually because of protobuf -- underneath the `DoCommand`, we actually use a [`Struct`](https://github.com/viamrobotics/api/blob/9bdcd0ee8e338c2c153b4aa5088727f916b49d64/proto/viam/common/v1/common.proto#L169). Inside a `Struct` are [`Value`](https://protobuf.dev/reference/csharp/api-docs/class/google/protobuf/well-known-types/value)s. `Value`s can only have certain types:
> null, a number, a string, a boolean, a recursive struct value, or a list of values.

When we call functions like [`viam.utils.dict_to_struct`](https://github.com/viamrobotics/viam-python-sdk/blob/db7a405b9c45226393e66c286439bd09f88fa5f6/src/viam/utils.py#L126) or [`viam.utils.struct_to_dict`](https://github.com/viamrobotics/viam-python-sdk/blob/db7a405b9c45226393e66c286439bd09f88fa5f6/src/viam/utils.py#L132), we're converting the values received in that dictionary or `Struct` into [python primitives](https://github.com/viamrobotics/viam-python-sdk/blob/db7a405b9c45226393e66c286439bd09f88fa5f6/src/viam/utils.py#L75) or [protobuf `Value`s](https://github.com/viamrobotics/viam-python-sdk/blob/db7a405b9c45226393e66c286439bd09f88fa5f6/src/viam/utils.py#L27). You can also see how [google themselves do it](https://github.com/protocolbuffers/protobuf/blob/main/python/google/protobuf/internal/well_known_types.py#L435).

So now that we have that background, we can understand why we need to update our typedefs. Furthermore, you can see why `DoCommandTypes` isn't necessary a great name for it -- this affects more than just the `DoCommand`; it affects every call that uses a protobuf `Struct` as its data structure. A more accurate name might be `ValueTypes`. 

Also, you shouldn't have to redefine `DoCommandTypes` in every single component/service that supports do, as that's super redundant. Instead, a potential spot for this typedef could be in `viam.utils`, which is where we also define the functions `value_to_primitive`, `dict_to_struct`, etc. Then you could simply import the typedef like anything else: `from viam.utils import ValueTypes`. 

Finally, we should update the function definitions of all the `do_command`s to use the new `ValueTypes` typedef, but also update the `primitive_to_value` and `value_to_primitive` functions to not take in/return `Any`, but rather `ValueTypes`. Similarly, we should update the `dict_to_struct` and `struct_to_dict` functions to also take in/return mappings of `str: ValueTypes`, rather than `str: Any`.

njooma (changes requested):
Just a small thing about adding documentation for the new type, otherwise this is great!

src/viam/utils.py
```diff
@@ -24,7 +24,11 @@
     from typing_extensions import ParamSpec
 
 
-def primitive_to_value(v: Any) -> Value:
+# Types that can be encoded into a protobuf `Value`
+ValueTypes = Union[bool, SupportsBytes, SupportsFloat, List, Mapping, str, None]
+
```
njooma, line 29: Sorry, should've specified better. This should be in docstring format where the comment is enclosed by triple quotes and is below the definition:

```python3
ValueTypes = Union[bool, SupportsBytes, SupportsFloat, List, Mapping, str, None]
"""Types that can be encoded into a protobuf ``Value``"""
```

njooma (changes requested):

njooma (approved):

PR #231, opened 2023-03-03
Update example.ipynb to clarify custom implementation requirements

JessamyT:
Not confident in my formatting, nor that this is the best way to word this.


njooma (approved):
Had to make a change for it to work properly, but this is good!

PR #232, opened 2023-03-06
RSDK 2128 allow partial movement sensor readings dictionary

purplenicole730:
[RSDK-2128](https://viam.atlassian.net/browse/RSDK-2128)

Currently, when a movement sensor does not have keys for all the possible readings, it returns an error. For example, the accelerometer on the Viam Rovers by default only have a linear acceleration reading set up. When a user calls `get_reading()` on the rover's accelerometer, an `GRPCError` is raised, failing out of the function altogether without any return.

The goal of this ticket is to be able to call `get_readings()` and return a dictionary that has default values for whatever reading is not supported on the sensor.

Calling `get_readings()` still returns an ERROR logs, but is able to return an incomplete library of `readings`.
```
2023-03-03 18:04:31,722         INFO    viam.rpc.dial (dial.py:212)     Connecting to socket: /tmp/proxy-WokwCPR4.sock
2023-03-03T23:04:31.854702Z ERROR viam_rust_utils::rpc::client_stream: Error processing trailers: Code=2 message=Position Unimplemented    
2023-03-03T23:04:31.854729Z ERROR viam_rust_utils::rpc::client_channel: error deserializing message: Code=2 message=Position Unimplemented    
(<Status.UNKNOWN: 2>, 'Code=2 message=Position Unimplemented', None)
2023-03-03T23:04:31.882735Z ERROR viam_rust_utils::rpc::client_stream: Error processing trailers: Code=2 message=LinearVelocity Unimplemented    
2023-03-03T23:04:31.882754Z ERROR viam_rust_utils::rpc::client_channel: error deserializing message: Code=2 message=LinearVelocity Unimplemented    
(<Status.UNKNOWN: 2>, 'Code=2 message=LinearVelocity Unimplemented', None)
2023-03-03T23:04:31.906334Z ERROR viam_rust_utils::rpc::client_stream: Error processing trailers: Code=2 message=AngularVelocity Unimplemented    
2023-03-03T23:04:31.906350Z ERROR viam_rust_utils::rpc::client_channel: error deserializing message: Code=2 message=AngularVelocity Unimplemented    
(<Status.UNKNOWN: 2>, 'Code=2 message=AngularVelocity Unimplemented', None)
2023-03-03T23:04:31.950363Z ERROR viam_rust_utils::rpc::client_stream: Error processing trailers: Code=2 message=CompassHeading Unimplemented    
2023-03-03T23:04:31.950380Z ERROR viam_rust_utils::rpc::client_channel: error deserializing message: Code=2 message=CompassHeading Unimplemented    
(<Status.UNKNOWN: 2>, 'Code=2 message=CompassHeading Unimplemented', None)
2023-03-03T23:04:31.971760Z ERROR viam_rust_utils::rpc::client_stream: Error processing trailers: Code=2 message=Orientation Unimplemented    
2023-03-03T23:04:31.971777Z ERROR viam_rust_utils::rpc::client_channel: error deserializing message: Code=2 message=Orientation Unimplemented    
(<Status.UNKNOWN: 2>, 'Code=2 message=Orientation Unimplemented', None)
{'linear_acceleration': y: 919.6875
z: 9580.078125
}
```

[RSDK-2128]: https://viam.atlassian.net/browse/RSDK-2128?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/viam/components/movement_sensor/movement_sensor.py
+++ b/src/viam/components/movement_sensor/movement_sensor.py
@@ -1,6 +1,7 @@
 import abc
 import asyncio
 from typing import Any, Dict, Final, Mapping, Optional, Tuple
+from grpclib import GRPCError
 
 from viam.proto.common import GeoPoint, Orientation, Vector3
 from viam.proto.component.movementsensor import GetPropertiesResponse
@@ -100,7 +101,7 @@ class MovementSensor(Sensor):
 
     async def get_readings(self, *, extra: Optional[Dict[str, Any]] = None, timeout: Optional[float] = None, **kwargs) -> Mapping[str, Any]:
         """Obtain the measurements/data specific to this sensor.
-        If a sensor is not configured to have a measurement or fails to read a piece of data, return empty objects or 0.
+        If a sensor is not configured to have a measurement or fails to read a piece of data, it will not appear in the readings dictionary.
         An ERROR log will still appear stating that the measurement/data was unimplemented.
 
         Returns:
@@ -125,12 +126,19 @@ class MovementSensor(Sensor):
             return_exceptions=True,
         )
 
-        return {
-            "position": pos[0] if not isinstance(pos, Exception) else GeoPoint(),
-            "altitude": pos[1] if not isinstance(pos, Exception) else 0,
-            "linear_velocity": lv if not isinstance(lv, Exception) else Vector3(),
-            "angular_velocity": av if not isinstance(av, Exception) else Vector3(),
-            "linear_acceleration": la if not isinstance(la, Exception) else Vector3(),
-            "compass": comp if not isinstance(comp, Exception) else 0,
-            "orientation": orient if not isinstance(orient, Exception) else Orientation(),
-        }
+        readings = {}
+        if not isinstance(pos, GRPCError):
+            readings["position"] = pos[0]
+            readings["altitude"] = pos[1]
+        if not isinstance(lv, GRPCError):
+            readings["linear_velocity"] = lv
+        if not isinstance(av, GRPCError):
+            readings["angular_velocity"] = av
+        if not isinstance(la, GRPCError):
+            readings["linear_acceleration"] = la
+        if not isinstance(comp, GRPCError):
+            readings["compass"] = comp
+        if not isinstance(orient, GRPCError):
+            readings["orientation"] = orient
+
+        return readings
--- a/src/viam/components/movement_sensor/movement_sensor.py
+++ b/src/viam/components/movement_sensor/movement_sensor.py
@@ -100,7 +100,7 @@ class MovementSensor(Sensor):
 
     async def get_readings(self, *, extra: Optional[Dict[str, Any]] = None, timeout: Optional[float] = None, **kwargs) -> Mapping[str, Any]:
         """Obtain the measurements/data specific to this sensor.
-        If a sensor is not configured to have a measurement or piece of data, it will be missing from the return dictionary.
+        If a sensor is not configured to have a measurement or fails to read a piece of data, it will be missing from the return dictionary.
         An ERROR log will still appear stating that the measurement/data was Unimplemented.
 
         Returns:
--- a/src/viam/components/movement_sensor/movement_sensor.py
+++ b/src/viam/components/movement_sensor/movement_sensor.py
@@ -3,6 +3,7 @@ import asyncio
 from typing import Any, Dict, Final, List, Mapping, Optional, Tuple
 from grpclib import GRPCError
 
+from viam.errors import MethodNotImplementedError, NotSupportedError
 from viam.proto.common import GeoPoint, Orientation, Vector3
 from viam.proto.component.movementsensor import GetPropertiesResponse
 from viam.resource.types import RESOURCE_NAMESPACE_RDK, RESOURCE_TYPE_COMPONENT, Subtype
@@ -129,6 +130,7 @@ class MovementSensor(Sensor):
 
         # Add returned value to the readings dictionary if value is of expected type; omit if unimplemented.
         def add_reading(name: str, reading, returntype: List) -> None:
+            possible_error_types = (NotImplementedError, MethodNotImplementedError, NotSupportedError)
             if type(reading) in returntype:
                 if name == "position":
                     readings["position"] = reading[0]
@@ -136,7 +138,7 @@ class MovementSensor(Sensor):
                 else:
                     readings[name] = reading
                 return
-            elif isinstance(reading, GRPCError) and "Unimplemented" in str(reading.message):
+            elif isinstance(reading, possible_error_types) or (isinstance(reading, GRPCError) and "Unimplemented" in str(reading.message)):
                 return
             raise reading
 
--- a/src/viam/components/movement_sensor/movement_sensor.py
+++ b/src/viam/components/movement_sensor/movement_sensor.py
@@ -1,7 +1,7 @@
 import abc
-import asyncio
 from typing import Any, Dict, Final, Mapping, Optional, Tuple
 
+from grpclib import GRPCError
 from viam.proto.common import GeoPoint, Orientation, Vector3
 from viam.proto.component.movementsensor import GetPropertiesResponse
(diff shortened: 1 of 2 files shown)


maximpertsov (changes requested):
looks good, but my only thought is we should explicitly document that errors result in missing keys, since this could be surprising to clients. Disregard my comment if we already document this at a higher level.

maximpertsov:
> > looks good, but my only thought is we should explicitly document that errors result in missing keys, since this could be surprising to clients. Disregard my comment if we already document this at a higher level.
> 
> @maximpertsov Where/how should we document this?

In the docstring for `get_readings`

src/viam/components/movement_sensor/movement_sensor.py
```diff
@@ -100,6 +100,8 @@ async def get_accuracy(
 
     async def get_readings(self, *, extra: Optional[Dict[str, Any]] = None, timeout: Optional[float] = None, **kwargs) -> Mapping[str, Any]:
         """Obtain the measurements/data specific to this sensor.
+        If a sensor is not configured to have a measurement or piece of data, it will be missing from the return dictionary.
```
maximpertsov, line 103:
```suggestion

        If a sensor is not configured to have a measurement or fails to read a piece of data, it will be missing from the return dictionary.
```

maximpertsov (commented):
sorry, was just discussing this @njooma, did we consider returning `None` rather than omitting the key?

src/viam/components/movement_sensor/movement_sensor.py
```diff
@@ -112,20 +112,21 @@ async def get_readings(self, *, extra: Optional[Dict[str, Any]] = None, timeout:
...
+        functions = {
+            "position": self.get_position,
+            "linear_velocity": self.get_linear_velocity,
+            "angular_velocity": self.get_angular_velocity,
+            "linear_acceleration": self.get_linear_acceleration,
+            "compass": self.get_compass_heading,
+            "orientation": self.get_orientation,
         }
+        readings = {}
+        for property, function in functions.items():
```
njooma, line 127: My only issue with doing it this way is that we're now awaiting each function in series instead of running them concurrently.

I think this can be refactored back to the original code, but with `return_exceptions=True` (see [documentation](https://docs.python.org/3.11/library/asyncio-task.html#asyncio.gather))

This would result in something like:

```python3
        (pos, lv, av, la, comp, orient) = await asyncio.gather(
            self.get_position(extra=extra, timeout=timeout),
            self.get_linear_velocity(extra=extra, timeout=timeout),
            self.get_angular_velocity(extra=extra, timeout=timeout),
            self.get_linear_acceleration(extra=extra, timeout=timeout),
            self.get_compass_heading(extra=extra, timeout=timeout),
            self.get_orientation(extra=extra, timeout=timeout),
            return_exceptions=True,
        )
        return {
            "position": pos[0] if not isinstance(pos, Exception) else GeoPoint(),
            "altitude": pos[1] if not isinstance(pos, Exception) else 0,
            "linear_velocity": lv if not isinstance(lv, Exception) else Vector3(),
            "angular_velocity": av if not isinstance(av, Exception) else Vector3(),
            "linear_acceleration": la if not isinstance(la, Exception) else Vector3(),
            "compass": comp if not isinstance(comp, Exception) else 0,
            "orientation": orient if not isinstance(orient, Exception) else Orientation(),
        }
```

This way, we address @maximpertsov issue of not dropping keys (and also be in line with RDK, which does not drop keys), and we continue to keep things parallelized.

njooma (changes requested):

njooma:
> sorry, was just discussing this @njooma, did we consider returning `None` rather than omitting the key?

In [RDK](https://github.com/viamrobotics/rdk/blob/2a42f5ce1b587987b5a15b0986d6910fe3d2c0f4/components/movementsensor/gpsnmea/serial.go#L172), we return empty objects/`0` values on error, which we should also do here.

njooma (approved):

maximpertsov (changes requested):
https://github.com/viamrobotics/viam-python-sdk/pull/232#issuecomment-1457173463

TLDR, I think returning default values is pretty bad and we should not do that in RDK either - I'd be more comfortable with either returning `None` or omitting the keys.

cheukt:
Thinking about it more, I think we should omit keys on unimplemented errors, error out if it is not an unimplemented error, and also make a ticket so that we omit keys in rdk on error

src/viam/components/movement_sensor/movement_sensor.py
```diff
@@ -100,6 +101,8 @@ async def get_accuracy(
 
     async def get_readings(self, *, extra: Optional[Dict[str, Any]] = None, timeout: Optional[float] = None, **kwargs) -> Mapping[str, Any]:
         """Obtain the measurements/data specific to this sensor.
+        If a sensor is not configured to have a measurement or fails to read a piece of data, it will not appear in the readings dictionary.
+        An ERROR log will still appear stating that the measurement/data was unimplemented.
```
cheukt, line 105: we're not currently logging these, so we can remove this line

src/viam/components/movement_sensor/movement_sensor.py
```diff
@@ -108,24 +111,34 @@ async def get_readings(self, *, extra: Optional[Dict[str, Any]] = None, timeout:
...
-            "altitude": alt,
-            "linear_velocity": lv,
-            "angular_velocity": av,
-            "linear_acceleration": la,
-            "compass": comp,
-            "orientation": orient,
-        }
+
+        readings = {}
+        if not isinstance(pos, GRPCError):
```
cheukt, line 130: I'd like these checks to be affirmative - so checking that the returned values is of the type expected, instead of it not being an exception.

cheukt (changes requested):

src/viam/components/movement_sensor/movement_sensor.py
njooma: How it currently works in RDK:

* For a local sensor, it will check if the error is "Unimplemented." If yes, it will add a default value. If not, it will raise the error
* For a remote sensor (using the client), it will raise all errors, including "Unimplemented" ones

My understanding is that if we want to somewhat mimic RDK functionality (but with key omissions), then we should check if this error is either `NotImplementedError` or `viam.errors.MethodNotImplementedError`. If yes, we **OMIT** the value from the dict. If it is any other type of error, we should raise the error. 

@cheukt please confirm?

cheukt: I think your understanding is fair - just have to explicit that it could be brittle

njooma: I think this is the same level of brittleness as in RDK

maximpertsov: [nit], but non-blocking - maybe we should stick the brittle logic into a function like we having in RDK where we do a similar string-based check? https://github.com/viamrobotics/rdk/blob/main/robot/client/client.go#L112-L114

maximpertsov (approved):

src/viam/components/movement_sensor/movement_sensor.py
```diff
@@ -108,24 +110,41 @@ async def get_readings(self, *, extra: Optional[Dict[str, Any]] = None, timeout:
...
+        # Add returned value to the readings dictionary if value is of expected type; omit if unimplemented.
+        def add_reading(name: str, reading, returntype: List) -> None:
+            if type(reading) in returntype:
+                if name == "position":
+                    readings["position"] = reading[0]
+                    readings["altitude"] = reading[1]
+                else:
+                    readings[name] = reading
+                return
+            elif isinstance(reading, GRPCError) and "Unimplemented" in str(reading.message):
```
njooma, line 139: Add the local `NotImplementedError` `viam.errors.MethodNotImplementedError` and `Viam.errors.NotSupportedError`

njooma (changes requested):

cheukt (approved):
LGTM pending @njooma 's suggestions

njooma (approved):

PR #237, opened 2023-03-10
[RSDK-2227] Modular Services

njooma:
co-authored by @purplenicole730 

This introduces modular services to the python sdk.

Some notes
* Lots of renaming from `component` to `resource`
* There are some naming peculiarities around Viam services and gRPC services
  * Take a look at the Summation Service example, where the service is called `SummationService`, the gRPC service is called `SummationRPCService`, and the client is `SummationServiceClient`. 
  * We're open to suggestions for other naming schemes
* This does reorganize and rename a few classes, which will be breaking. But modules are mostly internal still anyway, so we can work directly with sol eng to rectify these
* The organization of classes and types in the `resource` directory could be improved, particularly because it takes a lot of imports to get modular resources to work. Perhaps we can have some sort of meta import like `from module.requirements import *` or something of the sort


src/viam/robot/client.py
```diff
@@ -284,10 +286,16 @@ def get_component(self, name: ResourceName) -> ComponentBase:
...
         """
-        if name.type != "component":
-            raise ViamError(f"ResourceName does not describe a component: {name}")
+        if name.type != RESOURCE_TYPE_COMPONENT:
+            raise ValueError(f"ResourceName does not describe a component: {name}")
         with self._lock:
-            return self._manager.get_component(ComponentBase, name)
+            return self._manager.get_resource(ComponentBase, name)
+
+    def get_service(self, name: ResourceName) -> ServiceBase:
```
stuqdog, line 294: Should this have similar documentation to the `get_component` func?

tests/mocks/module/proto/summation.proto
stuqdog, line 1: (minor) As far as I can tell, this is identical to the `summation.proto` in the `examples` dir (actually the `tests` and `examples` for summation look pretty much identical to me across several files). Is there a particular reason to have two versions floating around?

src/viam/robot/client.py
njooma: You are absolutely right. I'll add it now

tests/mocks/module/proto/summation.proto
njooma: These are indeed identical. The reason I have them in both is because I want people to be able to go to the `examples` directory to see how to actually build things, but I don't want to import from the examples directory in the tests. I feel like those things shouldn't be so tightly connected (although I also dislike the repetition)

examples/module/client.py
```diff
@@ -38,6 +40,11 @@ async def main():
     # resp = await gizmo.do_one_bidi_stream(["arg1", "arg2", "arg3"])
     # print("do_one_bidi_stream result:", resp)
 
+    # ####### SUMMATION ####### #
+    summer = SummationService.from_robot(robot, name="mysum1")
+    sum = await summer.sum([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
+    print(f"The sum of the numbers [0, 10] is {sum}")
```
cheukt, line 46: nit: [0,10)?

src/viam/resource/registry.py
```diff
@@ -70,8 +74,9 @@ class Registry:
     resource using ``Registry.register(...)``.
     """
 
-    _SUBTYPES: ClassVar[Dict[Subtype, ResourceRegistration]] = {}
-    _COMPONENTS: ClassVar[Dict[str, ComponentCreator]] = {}
+    _SUBTYPES: ClassVar[Dict["Subtype", ResourceRegistration]] = {}
+    _COMPONENTS: ClassVar[Dict[str, "ComponentCreator"]] = {}
```
cheukt, line 78: I do feel fairly strongly about only having one Creator registry (even if we have to do a `Union["ComponentCreator","ServiceCreator"]` value typing) - I think it'll make future extensions easier. thoughts?

cheukt (commented):
generally looks good, do have a larger question around having multiple registries though.

src/viam/resource/registry.py
njooma: That was the original plan, but we decided to follow [RDK's lead](https://github.com/viamrobotics/rdk/blob/c720c2e49cac884fb0ad1c4c0bb0205256fa900b/registry/registry.go#L161). Happy to change it but I also don't think it limits us too much. These are internal types, we can always rename/rearrange them later without breaking.

I also think this might do the opposite of limit us. Any future extensions might not conform nicely to either of those types, and then we'd just keep growing the types allowed. By keeping these separate, we can add new types as we see fit without changing existing data structures.

tests/mocks/module/proto/summation.proto
stuqdog: Makes sense. would love ideally to not have the repetition but agree that a too-tight coupling is probably worse. seems fine to me to leave as-is.

stuqdog (approved):

src/viam/resource/registry.py
cheukt: the longer term goal for the RDK is to also move to a single registry as well. I think growing the types allowed is fine, as long as the keys can be uniquely generated. One additional feature we're adding is config validation, and adding it in one place is a bit nicer than two. We might add more types later, but I would push for making the function signatures be the same (takes a config and returns some kind of resource). 
I agree about the bit that it is mostly internal, but the external bit (how devs register new models) would change if we combine the registries later on. So in that sense, the ask is actually to allow devs to register new component/service models the same way.

src/viam/resource/registry.py
```diff
@@ -152,28 +132,8 @@ def lookup_subtype(cls, subtype: "Subtype") -> ResourceRegistration:
...
-        with cls._lock:
-            try:
-                return cls._COMPONENTS[f"{subtype}/{model}"]
-            except KeyError:
-                raise ResourceNotFoundError(subtype.resource_type, subtype.resource_subtype)
-
-    @classmethod
-    def lookup_service(cls, subtype: "Subtype", model: "Model") -> "ServiceCreator":
-        """Lookup and retrieve a registered service by its name
+    def lookup_resource(cls, subtype: "Subtype", model: "Model") -> "ResourceCreator":
```
cheukt, line 135: nit: `lookup_creator` or `lookup_resource_creator`? `resource` is slightly ambiguous here

cheukt (approved):
LGTM! small naming nit but otherwise looks great

src/viam/resource/registry.py
njooma: What if we mirror the way we do subtypes? We have `register_subtype`, `lookup_subtype`, and `SUBTYPES`. We could do `register_resource_model`, `lookup_resource_model`, and `RESOURCE_MODELS`? Or `register_resource_creator`, ...

cheukt: I like `resource_creator` as less ambiguous of the two

cheukt (approved):

PR #242, opened 2023-03-20
RSDK-2203 - Update server hostname if address and auth_entity don't match

cheukt:
not sure if we should be a bit more strict on this check, but this is a quick pass on letting users connect to app.viam robots on the same host


cheukt:
updated to not disable the check and instead be a bit smarter about replacing it

PR #243, opened 2023-03-22
Close RobotClient on error

njooma:
Close the RobotClient (and more importantly, any sockets/channels) if creating the RobotClient errors.


stuqdog (approved):
lgtm!

PR #244, opened 2023-03-23
Rsdk 2071 support modular validation and implicit dependencies

purplenicole730:
[RSDK-2071](https://viam.atlassian.net/browse/RSDK-2071)

Changes in this PR:
- Adds a new resource validator type with `lookup` and `register` functions to work with and store resource validators.
- Adds `validate_config` function that allows a module to look up and run (if any) a stored resource validator.
- Implements `ValidateConfig` to call `validate_config`.
- Adds a new gRPC error called ValidationError that will be raised when any Exception is thrown during `validate_config`.
- Test `validate_config` with example `Gizmo` module that stores a validator during init.
- Adds a validator function section to `README.md`

cc @benjirewis 

[RSDK-2071]: https://viam.atlassian.net/browse/RSDK-2071?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/examples/module/src/gizmo/__init__.py
+++ b/examples/module/src/gizmo/__init__.py
@@ -10,5 +10,4 @@ from .my_gizmo import MyGizmo
 
 Registry.register_subtype(ResourceRegistration(Gizmo, GizmoService, lambda name, channel: GizmoClient(name, channel)))
 
-Registry.register_resource_creator(Gizmo.SUBTYPE, MyGizmo.MODEL, MyGizmo.new)
-Registry.register_resource_validator(Gizmo.SUBTYPE, MyGizmo.MODEL, MyGizmo.validate_config)
+Registry.register_resource_creator(Gizmo.SUBTYPE, MyGizmo.MODEL, MyGizmo.new, MyGizmo.validate_config)
--- a/examples/module/src/gizmo/__init__.py
+++ b/examples/module/src/gizmo/__init__.py
@@ -10,3 +10,4 @@ from .my_gizmo import MyGizmo
 Registry.register_subtype(ResourceRegistration(Gizmo, GizmoService, lambda name, channel: GizmoClient(name, channel)))
 
 Registry.register_resource_creator(Gizmo.SUBTYPE, MyGizmo.MODEL, MyGizmo.new)
+Registry.register_resource_validator(Gizmo.SUBTYPE, MyGizmo.MODEL, MyGizmo.validate_config)
--- a/examples/module/src/gizmo/__init__.py
+++ b/examples/module/src/gizmo/__init__.py
@@ -2,7 +2,7 @@
 This file registers the Gizmo subtype with the Viam Registry, as well as the specific MyGizmo model.
 """
 
-from viam.components.motor import *  # Need to import motor to so the component registers itself
+from viam.components.motor import *  # noqa: F403 Need to import motor to so the component registers itself
 from viam.resource.registry import ResourceRegistration, Registry
 
 from .api import Gizmo, GizmoClient, GizmoService
--- a/examples/module/src/gizmo/__init__.py
+++ b/examples/module/src/gizmo/__init__.py
@@ -2,7 +2,7 @@
 This file registers the Gizmo subtype with the Viam Registry, as well as the specific MyGizmo model.
 """
 
-from viam.components.motor import *  # noqa: F403 Need to import motor to so the component registers itself
+from viam.components.motor import *  # noqa: F403 Need to import motor so the component registers itself
 from viam.resource.registry import ResourceRegistration, Registry
 
 from .api import Gizmo, GizmoClient, GizmoService
--- a/examples/module/src/gizmo/__init__.py
+++ b/examples/module/src/gizmo/__init__.py
@@ -3,11 +3,11 @@ This file registers the Gizmo subtype with the Viam Registry, as well as the spe
 """
 
 from viam.components.motor import *  # noqa: F403 Need to import motor so the component registers itself
-from viam.resource.registry import ResourceRegistration, Registry
+from viam.resource.registry import Registry, ResourceCreatorRegistration, ResourceRegistration
 
 from .api import Gizmo, GizmoClient, GizmoService
 from .my_gizmo import MyGizmo
 
 Registry.register_subtype(ResourceRegistration(Gizmo, GizmoService, lambda name, channel: GizmoClient(name, channel)))
 
-Registry.register_resource_creator(Gizmo.SUBTYPE, MyGizmo.MODEL, MyGizmo.new, MyGizmo.validate_config)
+Registry.register_resource_creator(Gizmo.SUBTYPE, MyGizmo.MODEL, ResourceCreatorRegistration(MyGizmo.new, MyGizmo.validate_config))
--- a/examples/module/src/gizmo/__init__.py
+++ b/examples/module/src/gizmo/__init__.py
@@ -2,6 +2,7 @@
 This file registers the Gizmo subtype with the Viam Registry, as well as the specific MyGizmo model.
 """
 
+from viam.components.motor import *
 from viam.resource.registry import ResourceRegistration, Registry
 
 from .api import Gizmo, GizmoClient, GizmoService
--- a/examples/module/src/gizmo/__init__.py
+++ b/examples/module/src/gizmo/__init__.py
@@ -2,7 +2,7 @@
 This file registers the Gizmo subtype with the Viam Registry, as well as the specific MyGizmo model.
 """
 
-from viam.components.motor import *
+from viam.components.motor import *  # Need to import motor to so the component registers itself
 from viam.resource.registry import ResourceRegistration, Registry
 
 from .api import Gizmo, GizmoClient, GizmoService
--- a/examples/module/src/gizmo/my_gizmo.py
+++ b/examples/module/src/gizmo/my_gizmo.py
@@ -29,15 +29,15 @@ class MyGizmo(Gizmo, Reconfigurable):
         return gizmo
 
     @classmethod
-    def validate_config(cls, config: ComponentConfig) -> List[str]:
+    def validate_config(cls, config: ComponentConfig) -> Sequence[str]:
         # Custom validation can be done by specifiying a validate function like this one. Validate functions
-        # can throw `throw` errors that will be returned to the parent through gRPC. Validate functions can
-        # also return a vector of strings representing the implicit dependencies of the resource.
+        # can raise errors that will be returned to the parent through gRPC. Validate functions can
+        # also return a sequence of strings representing the implicit dependencies of the resource.
         if "invalid" in config.attributes.fields:
             raise Exception(f"'invalid' attribute not allowed for model {cls.SUBTYPE}:{cls.MODEL}")
         motor = [config.attributes.fields["motor"].string_value]
         if motor == []:
-            raise Exception("A motor is required for Gizmo module.")
+            raise Exception("A motor is required for Gizmo component.")
         return motor
 
     async def do_one(self, arg1: str, **kwargs) -> bool:
--- a/examples/module/src/gizmo/my_gizmo.py
+++ b/examples/module/src/gizmo/my_gizmo.py
@@ -36,10 +36,10 @@ class MyGizmo(Gizmo, Reconfigurable):
         if "invalid" in config.attributes.fields:
             raise Exception(f"'invalid' attribute not allowed for model {cls.SUBTYPE}:{cls.MODEL}")
         arg1 = config.attributes.fields["arg1"].string_value
-        if arg1 is None:
-            raise Exception("arg1 attribute is required for Gizmo component.")
+        if arg1 == "":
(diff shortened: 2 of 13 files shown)


examples/module/src/gizmo/my_gizmo.py
```diff
@@ -28,6 +28,10 @@ def new(cls, config: ComponentConfig, dependencies: Mapping[ResourceName, Resour
         gizmo.my_arg = config.attributes.fields["arg1"].string_value
         return gizmo
 
+    @classmethod
+    def validate_config(cls, config: ComponentConfig) -> List[str]:
```
benjirewis, line 32: `async`?

examples/module/src/gizmo/my_gizmo.py
```diff
@@ -28,6 +28,10 @@ def new(cls, config: ComponentConfig, dependencies: Mapping[ResourceName, Resour
...
+        return [config.attributes.fields["arg1"].string_value]
```
benjirewis, line 33: Returning the value of `arg1` here will mean that the gizmo will depend on some resource `arg1` which doesn't exist. Let's talk about what we want to do in this function offline.

src/viam/module/module.py
```diff
@@ -176,3 +178,14 @@ def add_model_from_registry(self, subtype: Subtype, model: Model):
...
         except ResourceNotFoundError:
             raise ValueError(f"Cannot add model because it has not been registered. Subtype: {subtype}. Model: {model}")
+
+    async def validate_config(self, request: ValidateConfigRequest) -> ValidateConfigResponse:
+        config: ComponentConfig = request.config
+        subtype = Subtype.from_string(config.api)
+        model = Model.from_string(config.model)
+        validator = Registry.lookup_validator(subtype, model)
+        if validator is not None:
+            dependencies = validator(config)
```
benjirewis, line 188: I think we'll want some `try/except` logic around `validator` to catch potential errors from `validator(config)`. We'll want to return a gRPC error instead of raising an error within the Python module.

src/viam/resource/types.py
```diff
@@ -196,4 +196,5 @@ def resource_name_from_string(string: str) -> ResourceName:
     return ResourceName(namespace=parts[0], type=parts[1], subtype=parts[2], name=name)
 
 
+Validator: TypeAlias = Callable
```
benjirewis, line 199: I wonder if we can be any more strict with this type definition. Perhaps `Callable[[ComponentConfig], List[str]]`?

benjirewis (commented):
Looking great! I know this in draft but thought I'd leave a few comments to potentially discuss offline.

benjirewis:
We'll also want to update the module example [README](https://github.com/viamrobotics/viam-python-sdk/blob/main/examples/module/README.md) to describe the process of registering a custom validator function.

src/viam/resource/types.py
purplenicole730: I agree. I was doing that at first and it was giving me some Type errors so I kept it that way just for now.

src/viam/errors.py
```diff
@@ -83,3 +83,13 @@ class NotSupportedError(ViamGRPCError):
...
+
+
+class ValidationError(ViamGRPCError):
+    """
+    Exception raised when there is an error during module validation
+    """
+
+    def __init__(self, message: str):
+        self.message = message
+        self.grpc_code = Status.FAILED_PRECONDITION
```
purplenicole730, line 95: I'm not entirely sure which Status to use here.

examples/module/src/gizmo/my_gizmo.py
```diff
@@ -28,6 +28,18 @@ def new(cls, config: ComponentConfig, dependencies: Mapping[ResourceName, Resour
         gizmo.my_arg = config.attributes.fields["arg1"].string_value
         return gizmo
 
+    @classmethod
+    def validate_config(cls, config: ComponentConfig) -> List[str]:
+        # Custom validation can be done by specifiying a validate function like this one. Validate functions
+        # can throw `throw` errors that will be returned to the parent through gRPC. Validate functions can
```
benjirewis, line 34:
```suggestion
        # can throw `raise` errors that will be returned to the parent through gRPC. Validate functions can
```

examples/module/src/gizmo/my_gizmo.py
```diff
@@ -28,6 +28,18 @@ def new(cls, config: ComponentConfig, dependencies: Mapping[ResourceName, Resour
...
+        # also return a vector of strings representing the implicit dependencies of the resource.
```
benjirewis, line 35:
```suggestion
        # also return a List of strings representing the implicit dependencies of the resource.
```

examples/module/src/gizmo/my_gizmo.py
```diff
@@ -28,6 +28,18 @@ def new(cls, config: ComponentConfig, dependencies: Mapping[ResourceName, Resour
...
+        if "invalid" in config.attributes.fields:
+            raise Exception(f"'invalid' attribute not allowed for model {cls.SUBTYPE}:{cls.MODEL}")
+        motor = [config.attributes.fields["motor"].string_value]
+        if motor == []:
+            raise Exception("A motor is required for Gizmo module.")
```
benjirewis, line 40:
```suggestion
            raise Exception("A motor is required for Gizmo component.")
```

src/viam/errors.py
benjirewis: `FAILED_PRECONDITION` seems fine to me. Go and C++ both return `UNKNOWN`.

src/viam/resource/registry.py
```diff
@@ -112,6 +114,28 @@ def register_resource_creator(cls, subtype: "Subtype", model: "Model", creator:
                 raise DuplicateResourceError(key)
             cls._RESOURCES[key] = creator
 
+    @classmethod
+    def register_resource_validator(cls, subtype: "Subtype", model: "Model", validator: "Validator"):
+        """Register a specific validator function
```
benjirewis, line 119: Some inaccuracies in this doc comment. `creator (ResourceCreator)` arg should be `validator (Validator)` arg. I _think_ this only raises `DuplicateResourceError`. And, there should be no `Returns` section.

src/viam/resource/types.py
benjirewis: Bumping this + wondering if we can do a more strict type; no worries if not.

benjirewis (commented):
Looking great 🧑‍🔧 ! Just a few more nits.

src/viam/errors.py
cheukt: I think `FAILED_PRECONDITION` is actually the wrong status here https://grpc.github.io/grpc/core/md_doc_statuscodes.html. The system itself is not in a bad state, so maybe `INVALID_ARGUMENT`?

examples/module/src/gizmo/__init__.py
```diff
@@ -2,6 +2,7 @@
 This file registers the Gizmo subtype with the Viam Registry, as well as the specific MyGizmo model.
 """
 
+from viam.components.motor import *
```
njooma, line 5: You should add an inline comment here explaining why we need this. Something like `from viam.components.motor import *  # Need to import motor to so the component registers itself`

examples/module/src/gizmo/my_gizmo.py
njooma: I don' think `raise` needs to be in backticks

examples/module/src/gizmo/my_gizmo.py
```diff
@@ -28,6 +28,18 @@ def new(cls, config: ComponentConfig, dependencies: Mapping[ResourceName, Resour
         gizmo.my_arg = config.attributes.fields["arg1"].string_value
         return gizmo
 
+    @classmethod
+    def validate_config(cls, config: ComponentConfig) -> List[str]:
```
njooma, line 32: This should return a `Sequence` (which `List` is a subclass of)

examples/module/src/gizmo/my_gizmo.py
```diff
@@ -28,6 +28,18 @@ def new(cls, config: ComponentConfig, dependencies: Mapping[ResourceName, Resour
...
+        # Custom validation can be done by specifiying a validate function like this one. Validate functions
+        # can throw `throw` errors that will be returned to the parent through gRPC. Validate functions can
+        # also return a vector of strings representing the implicit dependencies of the resource.
+        if "invalid" in config.attributes.fields:
+            raise Exception(f"'invalid' attribute not allowed for model {cls.SUBTYPE}:{cls.MODEL}")
+        motor = [config.attributes.fields["motor"].string_value]
+        if motor == []:
+            raise Exception("A motor is required for Gizmo module.")
+        return motor
```
njooma, line 41: Can we also add a validation that `arg1` exists in the config? Since that is also required?

src/viam/errors.py
njooma: Agree, I think `INVALID_ARGUMENT` is better

src/viam/resource/types.py
njooma: We should definitely make this more strict. And again, as above, this should return a Sequence rather than a List

I also think it should be below `ResourceCreator` for alphabetical reasons :)

src/viam/resource/registry.py
```diff
@@ -151,6 +175,18 @@ def lookup_resource_creator(cls, subtype: "Subtype", model: "Model") -> "Resourc
             except KeyError:
                 raise ResourceNotFoundError(subtype.resource_type, subtype.resource_subtype)
 
+    @classmethod
+    def lookup_validator(cls, subtype: "Subtype", model: "Model") -> Union["Validator", None]:
```
njooma, line 179: What do people think of returning a no-op instead of `None` here? Something like:

```python3
return cls._VALIDATORS.get(f"{subtype}/{model}", lambda x: return [])
```

Pros:
1. You don't have to check whether a validator exists

Cons:
1. You don't really *know* if a validator exists. But how important is it for that information to be accurate? Do you care if a validator exists? Or do you simply care to call it and have the correct output?

njooma (changes requested):

src/viam/resource/registry.py
benjirewis: That sounds good to me. I don't think we care if the validator exists; there's no special logic in that case except that we want to return an empty list of strings back to the parent.

src/viam/resource/registry.py
purplenicole730: Oh my, thank you for catching this!

benjirewis (approved):
LGTM! Great job + thanks for this 🧑‍🔧

examples/module/src/gizmo/__init__.py
```diff
@@ -2,7 +2,7 @@
 This file registers the Gizmo subtype with the Viam Registry, as well as the specific MyGizmo model.
 """
 
-from viam.components.motor import *  # Need to import motor to so the component registers itself
+from viam.components.motor import *  # noqa: F403 Need to import motor to so the component registers itself
```
njooma, line 5: nit: there's an extra `to` in the sentence: "Need to import motor TO so the component registers itself". Can remove here and below

njooma (approved):
Just a small nit, but otherwise good!

examples/module/src/gizmo/__init__.py
```diff
@@ -10,3 +11,4 @@
 Registry.register_subtype(ResourceRegistration(Gizmo, GizmoService, lambda name, channel: GizmoClient(name, channel)))
 
 Registry.register_resource_creator(Gizmo.SUBTYPE, MyGizmo.MODEL, MyGizmo.new)
+Registry.register_resource_validator(Gizmo.SUBTYPE, MyGizmo.MODEL, MyGizmo.validate_config)
```
cheukt, line 14: I would actually roll this into the resource creator registry using a dataclass (so we'd register the creator and validator at the same time), like how we do it in Golang and C++

cheukt (changes requested):
Looks great! had a suggestion to reduce the amount of registries

src/viam/resource/registry.py
```diff
@@ -111,25 +128,7 @@ def register_resource_creator(cls, subtype: "Subtype", model: "Model", creator:
...
-
-        Raises:
-            DuplicateResourceError: Raised if the Subtype and Model pairing already has a registered validator
-        """
-        key = f"{subtype}/{model}"
-        with cls._lock:
-            if key in cls._VALIDATORS:
-                raise DuplicateResourceError(key)
-            cls._VALIDATORS[key] = validator
+            cls._RESOURCES[key] = registration
```
cheukt, line 131: we will want to check that resource creator is not empty in this case - validator can be empty tho

cheukt (changes requested):
One last request around checking that the registration actually contains something to register

cheukt (approved):
LGTM

src/viam/resource/registry.py
cheukt: we can keep as is for now - we're placing more burden on devs but if they have some kind of linting turned on it should be fine

PR #248, opened 2023-03-27
Fix log level setting

njooma:
re-up of https://github.com/viamrobotics/viam-python-sdk/pull/247

### DIFF
--- a/src/viam/logging.py
+++ b/src/viam/logging.py
@@ -61,6 +61,7 @@ def addHandlers(logger: logging.Logger):
 
 
 def setLevel(level: int):
+    global LOG_LEVEL
     LOG_LEVEL = level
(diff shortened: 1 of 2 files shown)


benjirewis (commented):
I removed myself on the last version of this PR but I can review this one if I have more context. I see that you're making `LOG_LEVEL` global, but this also seems to make gRPC servers not sensitive to `SIGINT` or `SIGTERM`. Why are we doing that?

njooma:
The `graceful_exit` context manager takes care of that for us actually, and was overwriting our own `SIGINT` and `SIGTERM` handlers. So it still works the same way, but we actually does what we expect it to

benjirewis (approved):

PR #250, opened 2023-03-29
[RSDK-2412] odrive module fix

njooma:
termination signals were not getting forwarded to the python module from the shell script that started the process. simple fix: just `exec` the python module from inside the shell script

### DIFF
--- a/src/viam/module/module.py
+++ b/src/viam/module/module.py
@@ -75,14 +75,17 @@ class Module:
 
     async def start(self):
         """Start the module service and gRPC server"""
-        await self.server.serve(path=self._address)
+        try:
+            await self.server.serve(log_level=self._log_level, path=self._address)
+        finally:
+            await self.stop()
 
     async def stop(self):
         """Stop the module service and gRPC server"""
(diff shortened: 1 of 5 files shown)


src/viam/module/module.py
```diff
@@ -75,14 +75,17 @@ async def _get_dependencies(self, dependencies: Sequence[str]) -> Mapping[Resour
...
+        finally:
+            await self.stop()
 
     async def stop(self):
         """Stop the module service and gRPC server"""
+        LOGGER.debug("Shutting down module")
         try:
             if self.parent is not None:
                 await self.parent.close()
-            self.server.close()
```
purplenicole730, line 85: This now means that if `self.parent` is None, it will not close. As long as this is on purpose!

purplenicole730 (approved):
Making sure that one thing is on purpose, but otherwise, LGTM!

src/viam/module/module.py
njooma: Server closure happens automatically now, we don't have to manually call it!

PR #254, opened 2023-04-10
RSDK-458 add server error interceptor

purplenicole730:


src/viam/rpc/server.py
```diff
@@ -45,6 +46,24 @@ def __init__(self, components: List[ResourceBase], *, module_service: Optional["
         if module_service is not None:
             services.append(module_service)
         services = ServerReflection.extend(services)
+
+        # Go into every service and replace the methods with a wrapped method that has error handling
+        for service in services:
+
+            def update_mapping():
```
zaporter-work, line 53: [q] Why did you put create a function to do this instead of just keeping it in the body of the loop?

src/viam/rpc/server.py
```diff
@@ -114,3 +134,21 @@ async def create_and_serve(
         """
         server = cls(components)
         await server.serve(host, port, log_level, path=path)
+
+
+def wrapper(func: Callable):
```
zaporter-work, line 139: [nit]
```suggestion
def error_wrapper(func: Callable):
```
```suggestion
def error_wrapping_wrapper(func: Callable):
```
```suggestion
def grpc_error_wrapper(func: Callable):
```
```suggestion
def wrap_with_grpc_error(func: Callable):
```
...? Not sure if any of these are perfect but a bit more description could help make this more obvious

zaporter-work (approved):
LGTM  % nits.

src/viam/rpc/server.py
purplenicole730: This is done so that the updating maps happens during runtime. Doing it beforehand in the loop caused it so that `mapping` pushed into the server was the last `mapping` value.

src/viam/rpc/server.py
purplenicole730: I agree with this. I'll hold off to see if the others have a preferred name.

src/viam/rpc/server.py
zaporter-work: Got it! That makes sense.

zaporter-work: Thank you!

src/viam/rpc/server.py
```diff
@@ -55,6 +71,7 @@ async def _grpc_event_handler(self, event: RecvRequest):
             host = address[0]
             port = address[1]
         msg = f"[gRPC Request] {host or 'xxxx'}:{port or 'xxxx'} - {event.method_name}"
+        print(f"{event.method_func} {event.method_name} {event.__dir__}")
```
cheukt, line 74: unnecessary?

src/viam/rpc/server.py
benjirewis: I like `grpc_error_wrapper`.

src/viam/rpc/server.py
```diff
@@ -45,6 +46,21 @@ def __init__(self, components: List[ResourceBase], *, module_service: Optional["
         if module_service is not None:
             services.append(module_service)
         services = ServerReflection.extend(services)
+
+        # Go into every service and replace the methods with a wrapped method that has error handling
+        for service in services:
```
cheukt, line 51: we don't actually want to do this - we only want to wrap methods that derive from `ResourceRPCServiceBase`, SignalingService for example we don't need to wrap

benjirewis (commented):
How does this change interact with the various calls to `grpc_error` (e.g. [here](https://github.com/viamrobotics/viam-python-sdk/blob/main/src/viam/module/module.py#L194)). Will we be "double" wrapping errors?

src/viam/rpc/server.py
```diff
@@ -114,3 +131,21 @@ async def create_and_serve(
         """
         server = cls(components)
         await server.serve(host, port, log_level, path=path)
+
+
+def wrapper(func: Callable):
```
cheukt, line 136: would add a test or two for this

src/viam/rpc/server.py
```diff
@@ -45,6 +46,21 @@ def __init__(self, components: List[ResourceBase], *, module_service: Optional["
         if module_service is not None:
             services.append(module_service)
         services = ServerReflection.extend(services)
+
+        # Go into every service and replace the methods with a wrapped method that has error handling
+        for service in services:
+
+            def update_mapping():
+                dict = service.__mapping__()
```
cheukt, line 54: nit: `mapping` or `handler_map`?

cheukt (changes requested):
looks great! would like to see some tests and a pass at better var names

src/viam/rpc/server.py
```diff
@@ -114,3 +131,21 @@ async def create_and_serve(
...
+
+    Args:
+        func (Callable): The function that should be wrapped
+    """
+
+    async def function(*args, **kwargs):
+        try:
+            new_func = await func(*args, **kwargs)
+            return new_func
+        except Exception as e:
```
cheukt, line 148: in cases of ViamErrors, we should pull the GRPCError out of that object and raise that

njooma, line 148: You should add an additional except right above this to catch and re-raise grpc errors (so we don't wrap those as status unknown)

```python3
try:
    ...
except GRPCError:
    raise
except Exception as e:
    raise GRPCError(Status.UNKNOWN, e)
```

src/viam/rpc/server.py
njooma, line 136: This wrapper should be private (start with an `_` underscore)

njooma (changes requested):

src/viam/rpc/server.py
purplenicole730: Should this be in a new `test_server` file or is there somewhere we can fit this?

njooma (approved):
nice

tests/test_server.py
```diff
@@ -0,0 +1,70 @@
...
+            await wrapped_raise_grpcerror()
+        assert e_info.value.args[0] == Status.CANCELLED
+        assert e_info.value.args[1] == "this is a fake GRPCError"
+
+    async def test_patch_mappings(self):
+        services = []
+        for registration in Registry.REGISTERED_SUBTYPES().values():
+            if registration.resource_type == FakeComponent:
+                continue
+            services.append(registration.rpc_service(manager=ResourceManager([ExampleArm("arm0")])))
```
cheukt, line 63: is there a reason why we're using `ExampleArm` instead of `MockArm`?

tests/test_server.py
```diff
@@ -0,0 +1,70 @@
...
+        services = ServerReflection.extend(services)
+        is_moving_handler = services[0].__mapping__()["/viam.component.arm.v1.ArmService/IsMoving"]
+        services = _patch_mappings(services)
+        patched_is_moving_handler = services[0].__mapping__()["/viam.component.arm.v1.ArmService/IsMoving"]
+        assert is_moving_handler[0] != patched_is_moving_handler[0]
+        assert is_moving_handler[1:3] == patched_is_moving_handler[1:3]
+        assert "_grpc_error_wrapper" in str(patched_is_moving_handler[0])
```
cheukt, line 70: I want to actually run `patched_is_moving_handler[0]` and see that we're getting the correct error back and then also a case where we don't get an error. I can see just subclassing `MockArm` and make `is_moving` error out

cheukt (changes requested):
would like test the behavior of the code a bit more, but otherwise looks great!

tests/test_server.py
```diff
@@ -0,0 +1,70 @@
...
+        wrapped_raise_grpcerror = _grpc_error_wrapper(raise_grpcerror)
+        with pytest.raises(GRPCError) as e_info:
+            await wrapped_raise_grpcerror()
+        assert e_info.value.args[0] == Status.CANCELLED
+        assert e_info.value.args[1] == "this is a fake GRPCError"
+
+    async def test_patch_mappings(self):
+        services = []
+        for registration in Registry.REGISTERED_SUBTYPES().values():
+            if registration.resource_type == FakeComponent:
```
cheukt, line 61: this is fine but can you add a comment about why we need this?

tests/test_server.py
purplenicole730: No reason, will make the change

tests/test_server.py
```diff
@@ -0,0 +1,81 @@
...
+        arm = MockArm("arm0")
+        manager = ResourceManager([arm])
+        service = ArmService(manager)
+        is_moving_handler = service.__mapping__()["/viam.component.arm.v1.ArmService/IsMoving"]
+
+        async with ChannelFor([service]) as channel:
+            client = ArmServiceStub(channel)
+            request = IsMovingRequest(name=arm.name)
+            response: IsMovingResponse = await client.IsMoving(request)
+            assert response.is_moving is False
```
cheukt, line 67: this doesn't really test much - we should either patch the service first and see that no error is raised (so that we test that the wrapped method isn't raising an error every time we call it), or we set `arm.is_moving = raise_viamgrpcerror` and not patch the mappings to test that there is indeed a difference in behavior after patching.

cheukt (approved):
LGTM - just have a request for the test

PR #256, opened 2023-04-11
RSDK-2483: Encoder API SDK changes (Python)

martha-johnston:
This change adds support for the new encoder API in the Python SDK. All of the tests in `test_encoder.py` pass successfully, however there is a consistent IDE error of a type mismatch whenever using the PositionType enum. It shouldn't be an actual issue (as it's not causing any tests to fail) but for reference, those errors occur at the following locations:
- `src/viam/components/encoder/service.py`, line 49, `request.position_type`, and line 51, `pos_type`
- `tests/mocks/components.py`, line 442, `return ... self.position_type`
- `examples/server/v1/components.py`, line 454, `return ... self.position_type`


src/viam/components/encoder/encoder.py
```diff
@@ -0,0 +1,76 @@
...
+        **kwargs,
+    ) -> Tuple[float, PositionType]:
+        """
+        Report the position of the encoder.
+        The value returned is the current position in terms of ticks or degrees,
+        and whether it is a relative or absolute position.
+
+        Returns:
+            float: Position of the encoder which can either be ticks since last zeroing
+                   for an incremental encoder or degrees for an absolute encoder.
```
randhid, line 56: Just so we avoid `incremental` model language confustion
```suggestion
                   for a relative encoder or degrees for an absolute encoder.
```

randhid (commented):
Few nits.

src/viam/components/encoder/encoder.py
```diff
@@ -0,0 +1,76 @@
...
+        *,
+        extra: Optional[Dict[str, Any]] = None,
+        timeout: Optional[float] = None,
+        **kwargs,
+    ) -> Tuple[float, PositionType]:
+        """
+        Report the position of the encoder.
+        The value returned is the current position either in relative units (ticks away from a zero position)
+        or absolute units (degrees along a circle).
+
```
njooma, line 53: This should have a comment describing the parameter `position_type` and what it is/used for

njooma (changes requested):

njooma (changes requested):
So the test is failing because evidently python 3.9 is being stricter with types. So everywhere you take `PositionType` as a param/return, you should make it `PositionType.ValueType`. 

[nit] Also, be sure to organize your imports. You can do so in VS Code by pressing Option+Shift+O

src/viam/components/encoder/encoder.py
```diff
@@ -0,0 +1,76 @@
...
+        position_type: Optional[PositionType.ValueType] = None,
+        *,
+        extra: Optional[Dict[str, Any]] = None,
+        timeout: Optional[float] = None,
+        **kwargs,
+    ) -> Tuple[float, PositionType.ValueType]:
+        """
+        Report the position of the encoder.
+        The value returned is the current position in terms of it's ``position_type``. The position will be either in relative units
+        (ticks away from a zero position) for PositionTypeTICKS or absolute units (degrees along a circle) for PositionTypeDEGREES.
```
njooma, line 52:
```suggestion
        (ticks away from a zero position) for ``PositionType.TICKS`` or absolute units (degrees along a circle) for ``PositionType.DEGREES``.
```

njooma (changes requested):

njooma (approved):

PR #260, opened 2023-04-18
[RSDK-2485] Client Reconnect

njooma:
This PR tackles a few issues:

1. When a robot client loses connection and reconnects, failed reconnection attempts could leave behind orphaned sockets
2. After a successful reconnection, existing resource clients would be useless with dead connections. Introduce a `ReconfigurableResourceRPCClientBase` type which all built in resource clients use
3. [flyby] Fix readme typo
4. [flyby] Add import sorting to our format function (and VSCode settings)
5. [flyby] Update pyproject dependencies
6. [flyby] Fix some Board type checker issues

For reviewers, the main file you should be checking is `robot.client`


src/viam/module/module.py
```diff
@@ -79,6 +80,12 @@ async def start(self):
         """Start the module service and gRPC server"""
         try:
             await self.server.serve(log_level=self._log_level, path=self._address)
+        except SystemExit as e:
```
cheukt, line 83: what does this do? maybe leave a comment around this bit

.vscode/settings.json
```diff
@@ -27,4 +27,7 @@
         "tobytes",
         "viam"
     ],
+    "isort.args": [
+        "-l 140 --profile black"
```
cheukt, line 31: surprised at some of the changes black is ok with, is it the line length change?

src/viam/robot/client.py
```diff
@@ -185,22 +187,37 @@ async def refresh(self):
...
-                manager.register(Registry.lookup_subtype(Subtype.from_resource_name(rname)).create_rpc_client(rname.name, self._channel))
-            except ResourceNotFoundError:
-                pass
         with self._lock:
+            if resource_names == self._resource_names:
+                return
+            for rname in resource_names:
+                if rname.type not in [RESOURCE_TYPE_COMPONENT, RESOURCE_TYPE_SERVICE]:
+                    continue
+                if rname.subtype == "remote":
```
cheukt, line 196: should `remote` also be a constant?

src/viam/robot/client.py
```diff
@@ -185,22 +187,37 @@ async def refresh(self):
...
-            if rname.type not in [RESOURCE_TYPE_COMPONENT, RESOURCE_TYPE_SERVICE]:
-                continue
-            if rname.subtype == "remote":
-                continue
-            try:
-                manager.register(Registry.lookup_subtype(Subtype.from_resource_name(rname)).create_rpc_client(rname.name, self._channel))
-            except ResourceNotFoundError:
-                pass
         with self._lock:
+            if resource_names == self._resource_names:
```
cheukt, line 191: does order matter here?

src/viam/robot/client.py
```diff
@@ -185,22 +187,37 @@ async def refresh(self):
...
+            for rname in resource_names:
+                if rname.type not in [RESOURCE_TYPE_COMPONENT, RESOURCE_TYPE_SERVICE]:
+                    continue
+                if rname.subtype == "remote":
+                    continue
+
+                if rname in self._manager.resources:
+                    res = self._manager.get_resource(ResourceBase, rname)
+                    if isinstance(res, ReconfigurableResourceRPCClientBase):
+                        res.reset_channel(self._channel)
```
cheukt, line 202: so if the resource list is different, do we always replace the channel even if we didn't lose connection?

cheukt (commented):
a few comments, how hard is it to add tests around this behavior?

src/viam/rpc/dial.py
```diff
@@ -136,9 +136,11 @@ class ViamChannel:
...
     def close(self):
         if not self._closed:
-            self.channel.close()
-            self.release()
-            self._closed = True
+            try:
+                self.channel.close()
+            finally:
+                self.release()
+                self._closed = True
```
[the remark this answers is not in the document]
njooma: Added

.vscode/settings.json
njooma: Line length and ordering (alphabetical, constants vs everything else)

src/viam/robot/client.py
njooma: We only use it here, so I don't think it needs to be a constant yet. Once we use it elsewhere, then yes I think it should be a constant

src/viam/robot/client.py
njooma: Don't think so?

src/viam/robot/client.py
njooma: Updated so that we only replace the client/channel if the channel has changed

src/viam/module/module.py
njooma: comments added

src/viam/rpc/dial.py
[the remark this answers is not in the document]
njooma: We do now

maximpertsov (approved):

biotinker:
Removing myself from reviewers since I have insufficient background on this codebase to be super productive; happy to test this on prior failure modes ones it's merged though

njooma:
> Removing myself from reviewers since I have insufficient background on this codebase to be super productive; happy to test this on prior failure modes ones it's merged though

Yup, just put you on so you could track progress

src/viam/robot/client.py
```diff
@@ -364,6 +400,15 @@ def resource_names(self) -> List[ResourceName]:
         with self._lock:
             return [r for r in self._resource_names]
 
+    def _close_channel(self, *, tab_count=0):
```
stuqdog, line 403: (q) I'm not sure I get why the `tab_count` exists. Why do we want to indent when calling `close()` specifically?

cheukt (approved):

src/viam/module/module.py
benjirewis: It does not resolve RSDK-2551, and I'm actually not sure we should add this @njooma (totally open to other opinions). The except here doesn't actually catch anything from grpclib. It's confusing, but the only error you can catch from grpclib's second stage of graceful exit is `asyncio.CancelledError`. From what I can tell, there's no way to catch the raised `SystemExit(143)`.

src/viam/robot/client.py
njooma: Because I call this internal function from a 2 difference places, and I want the logs to reflect a certain level of indentation depending on where it was called from (e.g. if it was called as part of a task that closes a lot of things, it should be indented. if it was closed on its own, it should not be indented)

src/viam/module/module.py
njooma: Removed

PR #263, opened 2023-04-20
Rsdk 2634 fix movement sensor property link

purplenicole730:
RSDK-2634

Originally the ticket was only supposed to fix the movement sensor property link in the docs, but it was expanded to include fixing the docs for other return types and parameters as well.
For example, the links for `Pose` would not point to the same link in the description and header of many functions. Now both links are pointing to the same link.

image
image

In addition, for clarity and better direction, it was decided for the __init__ files of each resource to import from the proto instead of the client/server/resource class. This way, the type links generated in the docs stay in the resource section (if appropriate).


src/viam/components/arm/arm.py
```diff
@@ -45,13 +45,10 @@ async def move_to_position(
...
         """
         Move the end of the arm to the Pose specified in ``pose``.
-        When obstacles are specified in ``world_state``, the motion plan of the arm will avoid them.
 
         Args:
 
-            pose (Pose): The destination Pose for the arm.
-
-            world_state (WorldState): The obstacles for the arm to avoid on its way to ``pose``.
+            pose (viam.proto.common.Pose): The destination Pose for the arm.
```
stuqdog, line 51: (q) This feels a little wordy. If necessary than disregard, but are we not able to keep this as just `Pose` (or analagous types below)?

src/viam/robot/client.py
```diff
@@ -517,8 +519,8 @@ async def stop_all(self, extra: Dict[ResourceName, Dict[str, Any]] = {}):
         Cancel all current and outstanding operations for the robot and stop all actuators and movement
 
         Args:
-            extra (Dict[ResourceName, Dict[str, Any]]): Any extra parameters to pass to the components' ``stop`` methods, keyed on the
-                                                        component's ``ResourceName``
+            extra (Dict[viam.proto.common.ResourceName, Dict[str, Any]]): Any extra parameters to pass to the components' ``stop`` methods,
+                keyed on thecomponent's ``ResourceName``
```
stuqdog, line 523: (nit) `thecomponent` -> `the component`.

Also, `s/component/resource`? I'd expect that `stop_all` would also stop services.

stuqdog (approved):
LGTM modulo a question and a nit.

src/viam/components/arm/arm.py
purplenicole730: The reason for this change is that Sphinx is not able to find the same class `Pose` as the one linked in the header. (Problems already known in Sphinx with using multiple targets.) The description on the PR has an image that shows you what I mean.
We believe it's better to have the description point to the same link, even if it means it's wordier.

If anything, I would argue that we should be calling `from viam.components.arm import Pose` instead of `from viam.proto.common import Pose`. That way, both the header and comments match up without the wordiness, but I'm not entirely sure why this was done this way.

purplenicole730: Thank you for the feedback! It helped me take a step back and realized there might be a better way, which is exactly what I suggested by importing `Pose` from `arm` rather than `proto`. I'll be making those changes now!

src/viam/components/movement_sensor/movement_sensor.py
```diff
@@ -20,7 +23,35 @@ class MovementSensor(Sensor):
...
+    @dataclass
+    class Properties:
+        linear_acceleration_supported: bool
+        angular_velocity_supported: bool
+        orientation_supported: bool
+        position_supported: bool
+        compass_heading_supported: bool
+        linear_velocity_supported: bool
+
+        def proto(self) -> GetPropertiesResponse:
```
njooma, line 35:
```suggestion
        @property
        def proto(self) -> GetPropertiesResponse:
```

src/viam/components/movement_sensor/service.py
```diff
@@ -121,7 +121,7 @@ async def GetProperties(self, stream: Stream[GetPropertiesRequest, GetProperties
             raise e.grpc_error
         timeout = stream.deadline.time_remaining() if stream.deadline else None
         response = await sensor.get_properties(extra=struct_to_dict(request.extra), timeout=timeout, metadata=stream.metadata)
-        await stream.send_message(response)
+        await stream.send_message(MovementSensor.Properties.proto(response))
```
njooma, line 124: I think this should be `response.proto`

njooma (changes requested):

njooma (approved):

PR #272, opened 2023-04-26
Rsdk 1831 slam wrapper

purplenicole730:
From now on, service classes in `server.py` files are `ResourceRPCService`, including in components. Service clients now don't have the word `Service` between the `Resource` and `Client`.

This is to help prepare as we move towards supporting `services`.


tests/test_slam_service.py
```diff
@@ -0,0 +1,137 @@
...
+    GetPositionResponse,
+    SLAMServiceStub,
+)
+from viam.resource.manager import ResourceManager
+from viam.services.slam import Pose, SLAMServiceClient, SLAMServiceRPCService
+from viam.utils import dict_to_struct, struct_to_dict
+
+from .mocks.services import MockSLAMService
+
+INTERNAL_STATE_CHUNKS = [bytes(5), bytes(2)]
```
stuqdog, line 22: (q) There's some repetition here with `tests/mocks/services.py` and `examples/servers/v1/services.py`. I assume there's a good reason for the duplication at least as far as the example code goes, but is it possible/would it make sense to dedup in the `tests/` code?

stuqdog (approved):
LGTM, modulo one question!

tests/test_slam_service.py
purplenicole730: I know that they're used for different things. I followed the way we do components for services. We also do the same for `modules`. As for why we split them this way, I'm not entirely sure.

src/viam/services/slam/slam.py
```diff
@@ -0,0 +1,50 @@
+import abc
+from typing import Final, List, Optional
+
+from viam.resource.types import RESOURCE_NAMESPACE_RDK, RESOURCE_TYPE_SERVICE, Subtype
+
+from ..service_base import ServiceBase
+from . import Pose
+
+
+class SLAMService(ServiceBase):
```
cheukt, line 10: just SLAM is fine

src/viam/services/slam/service.py
```diff
@@ -0,0 +1,78 @@
...
+        timeout = stream.deadline.time_remaining() if stream.deadline else None
+        position = await slam.get_position(timeout=timeout)
+        response = GetPositionResponse(pose=position)
+        await stream.send_message(response)
+
+    async def DoCommand(self, stream: Stream[DoCommandRequest, DoCommandResponse]) -> None:
+        request = await stream.recv_message()
+        assert request is not None
+        try:
+            arm = self.get_resource(request.name)
```
cheukt, line 72: arm -> slam

src/viam/services/slam/slam.py
```diff
@@ -0,0 +1,50 @@
...
+    This acts as an abstract base class for any drivers representing specific
+    arm implementations. This cannot be used on its own. If the ``__init__()`` function is
+    overridden, it must call the ``super().__init__()`` function.
+    """
+
+    SUBTYPE: Final = Subtype(RESOURCE_NAMESPACE_RDK, RESOURCE_TYPE_SERVICE, "slam")
+
+    @abc.abstractmethod
+    async def get_internal_state(self, *, timeout: Optional[float]) -> List[bytes]:
+        """Get the internal state of the SLAM algorithm required to continue mapping/localization
```
cheukt, line 23: nit: docs should end with periods

cheukt (changes requested):
small nits, but generally looks great!

src/viam/services/slam/service.py
purplenicole730: Thanks for catching this! Haha

src/viam/rpc/server.py
```diff
@@ -29,7 +29,7 @@ class Server(ResourceManager):
     gRPC Server
     """
 
-    def __init__(self, components: List[ResourceBase], *, module_service: Optional["ModuleService"] = None):
+    def __init__(self, components: List[ResourceBase], *, module_service: Optional["ModuleRPCService"] = None):
```
cheukt, line 32: rename this to resources and then update the docstring

tests/test_slam_service.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+class TestSLAMService:
+    name = "slam"
+    slam = MockSLAM(name="slam")
+
+    @pytest.mark.asyncio
+    async def test_get_internal_state_chunks(self):
+        chunks = await self.slam.get_internal_state()
+        assert chunks == INTERNAL_STATE_CHUNKS
```
cheukt, line 34: can now assert against MockSLAM.INTERNAL_STATE_CHUNKS

tests/test_slam_service.py
```diff
@@ -0,0 +1,137 @@
+from typing import List
```
cheukt, line 1: rename file to `test_slam`

cheukt (approved):

PR #283, opened 2023-05-08
New release workflow

njooma:
This reworked workflow has the following changes:

* When updating the version number, it will now commit and push that to a new branch: `rc-{VERSION}`, rather than committing and pushing it directly to `main`
* It will then open a PR from the `rc` branch to `main`
* The new release will be made from the `rc` branch, not `main`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -20,12 +20,12 @@ on:
 
 jobs:
   prepare:
-    if: github.repository_owner == 'viamrobotics' && github.ref == 'refs/heads/main'
-    runs-on: [self-hosted, x64]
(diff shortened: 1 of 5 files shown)


stuqdog (approved):

purplenicole730 (approved):

maximpertsov (commented):
~Just curious, does this action work if we don't bump the version?~

~I noticed on my related typescript PR that `peter-evans/create-pull-request@v5` will not open a new PR is there is no diff. Granted, I don't know how much we care about this scenario, but just wanted to point it out.~

EDIT: disregard - presumably there will be SOME diff in this case, so this shouldn't be an issue.

maximpertsov (approved):

PR #292, opened 2023-05-23
Add new class method to create module from args

njooma:
(flyby) remove unused mock test in base

### DIFF
--- a/examples/module/src/main.py
+++ b/examples/module/src/main.py
@@ -1,5 +1,4 @@
 import asyncio
-import sys
 
 from viam.module.module import Module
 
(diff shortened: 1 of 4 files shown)


benjirewis (approved):
This LGTM and works with the RDK changes added in [this commit](https://github.com/viamrobotics/rdk/commit/e7cb18c12bf7892d4ab74e06b1f9e39754decb67). Tested that `LOGGER.debug` statements are printed when server is started with `-debug` or module is configured with `log_level: debug`. Tested that `LOGGER.debug` statements are not printed when server is started with `-debug` but module is configured with `log_level: warn`.

cheukt (approved):

purplenicole730 (approved):

PR #302, opened 2023-05-26
Hoist error throwing to superclass

njooma:
Moved throwing of the `ResourceNotFoundError.grpc_error` to the superclass of all `<Resource>RPCService`s. This reduces a ton of boilerplate, especially around creating new services/module services


stuqdog (approved):
lgtm

PR #306, opened 2023-05-30
Add vision as modular service RSDK-3353

njooma:
Needed for @thegreatco 

Sets up vision as a service that can be implemented


tests/test_robot.py
maximpertsov, line 1: [q] why is this file changing? does this test need to add a service that is not modular?

maximpertsov (approved):

tests/test_robot.py
njooma: Ya exactly -- these tests a service client that is just a client (no model/service)

PR #309, opened 2023-06-05
RSDK-3449: update getkinematics and getgeometries wrappers

purplenicole730:
[RSDK-3449 ticket](https://viam.atlassian.net/browse/RSDK-3449)

This is to get ready for the new automated proto update that introduced two new methods. The new common methods `GetKinematics` and `GetGeometries` are now added to the component wrappers where needed.
Service tests have also been added to confirm that a `MethodNotImplementedError` is raised.

The `base` component now also has a `GetProperties` function that gives the `width` and `turn radius`. The PR's scope was changed to include the inclusion of this change so that all errors from the new protos update were addressed.


jckras (approved):
LGTM

benjirewis:
It looks like CI is failing because of issues with imports; any idea what's going on there?

src/viam/components/base/client.py
```diff
@@ -107,13 +109,18 @@ async def stop(
         request = StopRequest(name=self.name, extra=dict_to_struct(extra))
         await self.client.Stop(request, timeout=timeout)
 
-    async def is_moving(self, *, extra: Optional[Dict[str, Any]] = None, timeout: Optional[float] = None) -> bool:
-        if extra is None:
-            extra = {}
+    async def is_moving(self, *, timeout: Optional[float] = None) -> bool:
```
benjirewis, line 112: What happened to the `extra` parameter here?

src/viam/components/base/service.py
```diff
@@ -111,3 +124,6 @@ async def DoCommand(self, stream: Stream[DoCommandRequest, DoCommandResponse]) -
         result = await base.do_command(command=struct_to_dict(request.command), timeout=timeout, metadata=stream.metadata)
         response = DoCommandResponse(result=dict_to_struct(result))
         await stream.send_message(response)
+
+    async def GetGeometries(self, stream: Stream[GetGeometriesRequest, GetGeometriesResponse]) -> None:
+        raise MethodNotImplementedError("GetGeometries").grpc_error
```
benjirewis, line 129: Do all the services not need `GetKinematics`, too?

purplenicole730: No, GetKinematics is only used by the `arm` component.

src/viam/components/base/client.py
purplenicole730: I noticed that `IsMovingRequest` doesn't actually take `extra`, so I deleted it.

benjirewis: Ah interesting, yeah I thought all gRPC messages had an `extra` field for some reason, but looks like you have to explicitly define one.

benjirewis: This is hypothetically backward-breaking though, what are your thoughts on removing this @njooma?

purplenicole730: @benjirewis Naveed has agreed that removing it would be the way to go.

benjirewis (approved):
LGTM! Thanks

PR #310, opened 2023-06-06
Allow all resource types for dependencies, not just components

njooma:
Modular resources could only depend on components. NOW they can depend on services too :)

### DIFF
--- a/src/viam/module/module.py
+++ b/src/viam/module/module.py
@@ -7,7 +7,6 @@ from grpclib.utils import _service_name
 from typing_extensions import Self
 
 from viam import logging
-from viam.components.component_base import ComponentBase
 from viam.errors import ResourceNotFoundError, ValidationError
 from viam.proto.app.robot import ComponentConfig
 from viam.proto.module import (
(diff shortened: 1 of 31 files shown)


src/viam/module/module.py
```diff
@@ -83,17 +82,21 @@ async def _connect_to_parent(self):
...
+            return self.parent.get_service(name)
+        raise ValueError("Dependency does not describe a component nor a service")
+
+    async def _get_dependencies(self, dependencies: Sequence[str]) -> Mapping[ResourceName, ResourceBase]:
+        deps: Mapping[ResourceName, ResourceBase] = {}
         for dep in dependencies:
             rn = resource_name_from_string(dep)
-            component = await self._get_component(rn)
+            component = await self._get_resource(rn)
             deps[rn] = component
```
maximpertsov, line 100: was originally gonna say that the `component` variable name is wrong now, but maybe this can just be one line:

```suggestion
            deps[rn] = await self._get_resource(rn)
```

maximpertsov (changes requested):
We should add a test that includes a service in `tests/test_module.py` to validate this behavior - otherwise looks good!

njooma:
Also fixed an issue where adding a movement sensor/getting a status for a movement sensor would sporadically fail based on the ordering of whether `sensor` or `movement_sensor` came first (if `sensor` came first, then it would get added and `movement_sensor` would fail. But if `movement_sensor` got added first, then `sensor` would automatically get added as part of the `movement_sensor`). 

Doing this required reorganizing some imports to avoid circular imports

maximpertsov (approved):

clintpurser:
belated shipit

PR #327, opened 2023-06-22
Add new APIs To Gantry

randhid:
`MoveToPosition` now takes an optional list of speeds 
New `Home` in gantry API, returns a bool.

Do not merge in before most recent changes from proto are merged in.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "viam-sdk"
-version = "0.4.1"
+version = "0.4.2"
 description = "Viam Robotics Python SDK"
 authors = [ "Naveed <naveed@viam.com>" ]
 license = "Apache-2.0"
--- a/src/viam/sessions_client.py
+++ b/src/viam/sessions_client.py
@@ -36,11 +36,6 @@ def loop_kwargs():
     return {}
 
 
-async def delay(coro, seconds):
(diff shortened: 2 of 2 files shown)


src/viam/components/gantry/client.py
```diff
@@ -40,6 +45,7 @@ async def get_position(self, *, extra: Optional[Dict[str, Any]] = None, timeout:
     async def move_to_position(
         self,
         positions: List[float],
+        sppeds: List[float],
```
martha-johnston, line 48: typo: speeds

martha-johnston (approved):

cheukt (approved):

cheukt:
@randhid you can pull in the automated api changes into this pr - that way we can test your changes against it - or you can push your current changes into that pr. Either way would like to avoid having to push the gantry api changes in without the corresponding updates to the rest of the code

randhid:
> @randhid you can pull in the automated api changes into this pr - that way we can test your changes against it - or you can push your current changes into that pr. Either way would like to avoid having to push the gantry api changes in without the corresponding updates to the rest of the code

@cheukt @njooma I pulled in the API protos branch, but it has a ton of junk that's not gantry-related, that are also API changes for other resources. The Gantry tests are passing, I'm just unsure if you want this merged in in this state? I could try pulling in the gantry-only-API changes but I'm nervous about doing that.

PR #331, opened 2023-06-23
RSDK-3628 - Create Data Location Client Able to Connect to App

8ashar:
This PR includes a new (unimplemented) data client that uses location-level authentication to make requests to app. The methods included will allow for data to be uploaded to and retrieved from app programmatically.

### DIFF
--- a/src/viam/app/data/client.py
+++ b/src/viam/app/data/client.py
@@ -20,14 +20,26 @@ LOGGER = logging.getLogger(__name__)
 
 
 class DataClient:
-    """
-    gRPC client for uploading and retreiving data from app
+    """gRPC client for uploading and retreiving data from app
 
+    Use connect() to instantiate a DataClient::
+
+            DataClient.connect(...)
+
+    You must ``close()`` a DataClient after its use.
     """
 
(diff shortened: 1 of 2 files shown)


src/viam/app/data/client.py
```diff
@@ -0,0 +1,113 @@
...
+    async def remove_tags_from_binary_data_by_file_ids(self, file_ids: Optional[List[str]], tags: Optional[List[str]]) -> None:
+        raise NotImplementedError()
+
+    async def remove_tags_from_binary_data_by_filter(self, filter: Optional[Filter], tags: Optional[List[str]]) -> None:
+        raise NotImplementedError()
+
+    async def tags_by_filter(self, filter: Optional[Filter]) -> List[str]:
+        raise NotImplementedError()
+
+    # later
```
stuqdog, line 96: Can we make this comment a bit more explicit? `later` is a little vague. The same applies below.

stuqdog (changes requested):
Thanks for putting this together! A couple small comments and suggested API changes based on additional RDK digging, but in general this looks very good :)

PR #332, opened 2023-06-23
[RSDK-3628] Create Data Location Client Able to Connect to App

8ashar:
This PR includes a new (unimplemented) data client that uses location-level authentication to make requests to app. The methods included will allow for data to be uploaded to and retrieved from app programmatically.

### DIFF
--- a/src/viam/app/client.py
+++ b/src/viam/app/client.py
@@ -15,14 +15,12 @@ class AppClient:
     Use create() to instantiate an AppClient::
 
         AppClient.create(...)
-
-    You may ``close()`` an AppClient after its use.
     """
 
     @classmethod
     async def create(cls, dial_options: DialOptions) -> Self:
         """
-        Create an app client that establishes a connection to viam.app.com.
+        Create an app client that establishes a connection to app.viam.com.
 
         Args:
             dial_options (DialOptions): Required information for authorization and connection to app, creds and auth_entity are necessary.
--- a/src/viam/app/client.py
+++ b/src/viam/app/client.py
@@ -1,0 +1,48 @@
+from typing_extensions import Self
+
+from grpclib.client import Channel
+
+from viam.rpc.dial import DialOptions, _dial_app, _get_access_token
+from viam import logging
+from viam.app.data.client import DataClient
+
+LOGGER = logging.getLogger(__name__)
+
+
+class AppClient:
+    """gRPC client for all communication and interaction with app.
+
+    Use create() to instantiate an AppClient::
+
+        AppClient.create(...)
+
+    You may ``close()`` an AppClient after its use.
+    """
+
+    @classmethod
+    async def create(cls, dial_options: DialOptions) -> Self:
+        """
+        Create an app client that establishes a connection to viam.app.com.
+
+        Args:
+            dial_options (DialOptions): Required information for authorization and connection to app, creds and auth_entity are necessary.
+
+        Returns:
+            Self: the AppClient.
+        """
+        self = cls()
+        self._channel = await _dial_app(dial_options)
+        access_token = await _get_access_token(self._channel, dial_options.auth_entity, dial_options)
+        self._metadata = {"authorization": f"Bearer {access_token}"}
+        return self
+
+    _channel: Channel
+    _metadata: str
+    _closed: bool = False
+
+    @property
+    def data_client(self) -> DataClient:
+        return DataClient(self._channel, self._metadata)
+
+    async def close(self):
+        raise NotImplementedError()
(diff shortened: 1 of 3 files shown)


stuqdog (approved):
LGTM!

njooma (commented):
Thoughts:
1. Should we have an `AppClient` similar to how we have a `RobotClient` that manages the lifecycle of the connection? The call site could be something like `app.data_client.tabular_data_by_filter(...)`
2. If not an `AppClient` class under which all additional app API clients fall under, then perhaps a superclass that defines `connect` and `close`, with a default implementations so that code doesn't have to be repeated for each client class?

stuqdog:
> Thoughts:
> 
>     1. Should we have an `AppClient` similar to how we have a `RobotClient` that manages the lifecycle of the connection? The call site could be something like `app.data_client.tabular_data_by_filter(...)`
> 
>     2. If not an `AppClient` class under which all additional app API clients fall under, then perhaps a superclass that defines `connect` and `close`, with a default implementations so that code doesn't have to be repeated for each client class?

Good questions!  I agree that we should either create a broader `AppClient` class or have a superclass that all app API client classes inherit from. I'd initially had some hesitation out of concern that an `AppClient` class would be overbroad for someone who only wants access to data APIs, e.g., and hoped we could punt on this a bit to see how things are actually used. On reflection though I think I'm still operating in a "it's C++ and it's okay if we change things up" mentality, and the comparative stability of the python SDK encourages us to make a decision now. 

I think I favor the following:
1.  create an `AppClient` class. Given the relatively small number of cloud API clients compared to the world of possible resources, I think it's fine to explicitly enumerate them within the class. So for now, the class can just have `connect(...)`, `close()`, and `initialize_data_client(...)` as methods, and `data_client`, `_channel`, and `_metadata` as class data members.
2. Refactor `DataClient`'s `connect()` to take `channel` and `metadata` as args, update comment to clarify that it's not meant to be called on its own but should rather be called through an `AppClient`'s methods. Rename it as well, to `_connect()` at a minimum but probably to `_create()` or something? making it an `__init__(...)` strikes me as a code smell given the desire to _not_ have users call it directly, but I am no Python expert so I'll happily defer on this.
3. remove `DataClient`'s `close()` method (for now at least) to further encourage use of the `AppClient` as the user's point of entry. Adding it back is always safe if for some reason it turns out to be necessary.

njooma:
Per IRL discussion:
1. Create an `AppClient` that has `DataClient` as a member (likely a `@property`)
    * `AppClient` will do all the channel initialization/closing/life cycle managment
1. `DataClient` will only have an `__init__`, that takes a channel (and metadata)
1. Close doesn't really  matter since it's a pure gRPC connection (not webRTC), but we should be good citizens and expose that function anyway

src/viam/app/client.py
```diff
@@ -0,0 +1,48 @@
...
+
+        AppClient.create(...)
+
+    You may ``close()`` an AppClient after its use.
+    """
+
+    @classmethod
+    async def create(cls, dial_options: DialOptions) -> Self:
+        """
+        Create an app client that establishes a connection to viam.app.com.
```
stuqdog, line 25: `viam.app.com` -> `app.viam.com`

src/viam/app/data/client.py
```diff
@@ -0,0 +1,100 @@
...
+from viam.proto.app.datasync import DataSyncServiceStub
+from viam import logging
+
+LOGGER = logging.getLogger(__name__)
+
+
+class DataClient:
+    """gRPC client for uploading and retreiving data from app
+
+    Constructor must be passed a channel to app, along with the required metadata.
```
stuqdog, line 23: Might be good to expand this comment a bit to indicate that using an `AppClient` is the preferred way to interact with a `DataClient`.

src/viam/app/client.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class AppClient:
+    """gRPC client for all communication and interaction with app.
+
+    Use create() to instantiate an AppClient::
+
+        AppClient.create(...)
+
+    You may ``close()`` an AppClient after its use.
```
stuqdog, line 19: I think we should probably get rid of this line. It should be apparent by the existence of a `close()` method that it can be called, and any considerations for when it should or shouldn't be used probably better belongs in the `close()` documentation.

stuqdog (changes requested):
A couple small requested documentation changes, but it's looking good!

stuqdog (approved):
lgtm!

njooma (approved):
nice

PR #333, opened 2023-06-26
RSDK-3775 - Fix unawaited warning

cheukt:
seems like having async tasks call themselves causes issues so refactored it a bit to resemble our other recurring tasks. (https://stackoverflow.com/questions/37512182/how-can-i-periodically-execute-a-function-with-asyncio)

behavior should be the same

Screen Shot 2023-06-26 at 11 05 39 AM


benjirewis (approved):
Nice!

njooma (approved):

PR #334, opened 2023-06-27
[RSDK-3634] Implement data.proto APIs that Add/Remove Tags

8ashar:
This PR implements and adds docstrings to the gRPC methods that'll allow a DataClient to send a gGRPC request to app to add, remove, and retrieve tags. The PR also implements a function to retrieve bounding box labels. There are a few other minor changes throughout. Namely, some function signatures are updated to include defaults for certain parameters (i.e. empty filter), and the AppClient "close()" function is implemented with logging.

### DIFF
--- a/src/viam/app/client.py
+++ b/src/viam/app/client.py
@@ -1,5 +1,7 @@
-from grpclib.client import Channel
+from typing import Mapping
 from typing_extensions import Self
+
+from grpclib.client import Channel
 
 from viam import logging
 from viam.app.data.client import DataClient
@@ -9,11 +11,13 @@ LOGGER = logging.getLogger(__name__)
 
 
 class AppClient:
-    """gRPC client for all communication and interaction with app.
+    """
+    gRPC client for all communication and interaction with app.
 
     Use create() to instantiate an AppClient::
 
         AppClient.create(...)
+
     """
 
     @classmethod
@@ -34,12 +38,17 @@ class AppClient:
         return self
 
     _channel: Channel
-    _metadata: str
+    _metadata: Mapping[str, str]
     _closed: bool = False
 
     @property
     def data_client(self) -> DataClient:
         return DataClient(self._channel, self._metadata)
 
-    async def close(self):
-        raise NotImplementedError()
+    def close(self):
+        if self._closed:
+            LOGGER.debug("AppClient is already closed")
+            return
+        LOGGER.debug("Closing gRPC channel to app")
+        self._channel.close()
+        self._closed = True
--- a/src/viam/app/data/client.py
+++ b/src/viam/app/data/client.py
@@ -4,6 +4,18 @@ from grpclib.client import Channel
 
 from viam import logging
 from viam.proto.app.data import (
+    AddTagsToBinaryDataByIDsRequest,
+    AddTagsToBinaryDataByIDsResponse,
+    AddTagsToBinaryDataByFilterRequest,
+    AddTagsToBinaryDataByFilterResponse,
+    RemoveTagsFromBinaryDataByIDsRequest,
+    RemoveTagsFromBinaryDataByIDsResponse,
+    RemoveTagsFromBinaryDataByFilterRequest,
+    RemoveTagsFromBinaryDataByFilterResponse,
+    TagsByFilterRequest,
+    TagsByFilterResponse,
+    BoundingBoxLabelsByFilterRequest,
+    BoundingBoxLabelsByFilterResponse,
     BinaryDataByFilterRequest,
     BinaryDataByFilterResponse,
     BinaryDataByIDsRequest,
@@ -29,10 +41,10 @@ LOGGER = logging.getLogger(__name__)
 class DataClient:
     """gRPC client for uploading and retreiving data from app
 
-    Constructor is used by AppClient to instantiate relevant service stubs. Calls to DataClient methods should be made through AppClient.
+    Constructor is used by AppClient to instantiate relevant service stubs. Calls to DataClient methods should be made through AppClient
     """
 
-    def __init__(self, channel: Channel, metadata: str):
+    def __init__(self, channel: Channel, metadata: Mapping[str, str]):
         """
         Create a data client that establishes a connection to app.
 
@@ -46,7 +58,7 @@ class DataClient:
 
     _data_client: DataServiceStub
     _data_sync_client: DataSyncServiceStub
-    _metadata: str
+    _metadata: Mapping[str, str]
 
     async def tabular_data_by_filter(
         self,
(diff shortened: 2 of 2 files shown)


src/viam/app/client.py
```diff
@@ -0,0 +1,54 @@
...
+    def data_client(self) -> DataClient:
+        return DataClient(self._channel, self._metadata)
+
+    def close(self):
+        LOGGER.debug("Closing AppClient")
+        if self._closed:
+            LOGGER.debug("AppClient is already closed")
+            return
+        LOGGER.debug("Closing gRPC channel to app")
+        LOGGER.debug("Closing grpc-lib Channel instance")
```
njooma, line 52: We probably don't need this, since the line above says the same thing

src/viam/app/data/client.py
```diff
@@ -0,0 +1,159 @@
...
+    async def delete_binary_data_by_ids(self, file_ids: Optional[List[str]]) -> None:
+        raise NotImplementedError()
+
+    async def add_tags_to_binary_data_by_file_ids(self, file_ids: Optional[List[str]], tags: Optional[List[str]]) -> None:
+        """
+        Add tags to binary data using file ID's
+
+        Args:
+            file_ids (List[str]): List of file ID's specifying binary data to modify
+            tags (List[str]): List of tags to add to specified binary data
```
njooma, line 79: Should these be optional? Seems like adding tags to binary data by file ids should REQUIRE the file_ids and the tags

src/viam/app/data/client.py
```diff
@@ -0,0 +1,159 @@
...
+        request = AddTagsToBinaryDataByFileIDsRequest(file_ids=file_ids, tags=tags)
+        await self._data_client.AddTagsToBinaryDataByFileIDs(request, metadata=self._metadata)
+
+    async def add_tags_to_binary_data_by_filter(self, tags: Optional[List[str]], filter: Optional[Filter] = Filter()) -> None:
+        """
+        Add tags to binary data using a filter
+
+        Args:
+            filter (Filter): Filter specifying binary data to modify
+            tags (List[str]): List of tags to add to specified binary data
```
njooma, line 90: Same, should these be required?

Also, `Args` order should be in the same order as in the method signature

src/viam/app/data/client.py
```diff
@@ -66,19 +71,62 @@ async def delete_binary_data_by_ids(self, file_ids: Optional[List[str]]) -> None
...
 
-    async def tags_by_filter(self, filter: Optional[Filter]) -> List[str]:
-        raise NotImplementedError()
+    async def remove_tags_from_binary_data_by_file_ids(self, tags: Optional[List[str]], file_ids: Optional[List[str]] = Filter()) -> None:
+        """
+        Remove tags from binary data using file ids
+
+        Args:
+            file_ids (List[str]): List of file ID's specifying binary data to modify
+            tags (List[str]): List of tags to remove from specified binary data
```
njooma, line 101: Looks like we're going to run into the `Optional` vs `Required` question on all of these.

If these truly are optional, the docstring should mention what the behavior is when left empty.

njooma (changes requested):

src/viam/app/data/client.py
8ashar: See belo

src/viam/app/data/client.py
8ashar: See above

stuqdog (changes requested):
One nit, but on first pass looks good!

njooma (approved):

stuqdog:
@8ashar per this PR, `{add|remove}_tags_by_file_id` is going to become `..._by_id` instead. Might be good to change the user-facing method names now, just so it's always consistent.

stuqdog (changes requested):
LGTM but I'm requesting that we change the user-facing API of the `*_by_file_ids` methods to instead be `*_by_ids`, and update the args/comments appropriately. The methods are changing name and we should fix the user-facing experience to match right away, then update the methods once the new protos go into effect.

See [this](https://github.com/viamrobotics/api/pull/281) pr for details.

Edit: in addition, we should change the actual argument structure. Right now we're passing file IDs (which are strings) into a deprecated `file_id` field that isn't looked at. What we should pass are BinaryIDs (which are a proto type `BinaryID` and which include, among other things, an optional `file_id`) into the `binary_id` field that _is_ looked at.

src/viam/app/client.py
```diff
@@ -17,12 +17,14 @@ class AppClient:
...
         AppClient.create(...)
+
+    You may ``close()`` an AppClient after its use.
     """
 
     @classmethod
     async def create(cls, dial_options: DialOptions) -> Self:
         """
-        Create an app client that establishes a connection to app.viam.com.
+        Create an app client that establishes a connection to viam.app.com.
```
stuqdog, line 27: Looks like this reverted back to `viam.app.com`, we should change it back.

src/viam/app/data/client.py
```diff
@@ -18,14 +19,16 @@
...
 )
 from viam.proto.app.datasync import DataSyncServiceStub
 
 
 class DataClient:
-    """gRPC client for uploading and retreiving data from app
+    """
+    gRPC client for uploading and retreiving data from app
 
-    Constructor is used by AppClient to instantiate relevant service stubs. Calls to DataClient methods should be made through AppClient.
```
stuqdog, line 23: We probably want to still have this line.

src/viam/app/data/client.py
```diff
@@ -44,21 +47,16 @@ def __init__(self, channel: Channel, metadata: Mapping[str, str]):
     _data_sync_client: DataSyncServiceStub
     _metadata: Mapping[str, str]
 
-    async def tabular_data_by_filter(
-        self,
-        filter: Optional[Filter],
-        dest: Optional[str]
-    ) -> List[Mapping[str, Any]]:
+    async def close(self):
```
stuqdog, line 50: I think we don't need this, since `close` is now being handled by the `AppClient`.

src/viam/app/client.py
```diff
@@ -17,12 +17,14 @@ class AppClient:
     Use create() to instantiate an AppClient::
 
         AppClient.create(...)
+
+    You may ``close()`` an AppClient after its use.
```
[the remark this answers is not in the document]
8ashar: Thanks for picking this out I think I accidentally reset a commit or two too far back :/.

src/viam/app/data/client.py
```diff
@@ -65,36 +77,103 @@ async def delete_binary_data_by_filter(self, filter: Optional[Filter]) -> None:
...
-        raise NotImplementedError()
+        Returns:
+            List[str]: The list of bounding box labels
+        """
+        filter = filter if filter else Filter()
+        request = BoundingBoxLabelsByFilterRequest(filter=filter)
+        response: BoundingBoxLabelsByFilterResponse = await self._data_client.BoundingBoxLabelsByFilter(request, metadata=self._metadata)
+        return response.labels
+
+    async def data_capture_upload(self, metadata: UploadMetadata , sensor_contents: Optional[List[SensorData]] = None) -> None:
```
stuqdog, line 173: Awesome to see that you went ahead and implemented these two! We should add docstring comments explaining the semantics of these a bit. Also, have you tested them at all? Probably good to do so.

It's also totally fine to leave these for a subsequent PR, and in fact maybe good to do so if you think that testing/documenting might take some time.

stuqdog (commented):
One comment that can definitely be pushed to a subsequent PR. Also: we should hold off on merging this until the data proto updates have been merged.

src/viam/app/data/client.py
8ashar: As per my Slack message: I meant to leave this out of this PR! The functions work and upload data but trying to decipher the required parameters is its own mess that is probably better solved in the data_sync.proto ticket. I think I'm going to revert these back to raising unimplemented errors and commit again. I have a good local testing suite set up for these though that highlights what we might want to include

stuqdog (approved):
We should coordinate release timing between this, the other data API PR, and the proto updates, but LGTM! Thanks for pulling this together.

src/viam/app/data/client.py
```diff
@@ -65,36 +75,101 @@ async def delete_binary_data_by_filter(self, filter: Optional[Filter]) -> None:
...
+            tags (List[str]): List of tags to add to specified binary data. Must be non-empty
+            binary_ids (List[viam.app.proto.BinaryID]): List of BinaryIDs specifying binary data to tag. Must be non-empty
 
-    async def remove_tags_from_binary_data_by_file_ids(self, file_ids: Optional[List[str]], tags: Optional[List[str]]) -> None:
-        raise NotImplementedError()
+        Raises:
+            GRPCError: if no binary_ids or tags are provided
+        """
+        request = AddTagsToBinaryDataByIDsRequest(binary_ids=binary_ids, tags=tags)
+        _: AddTagstoBinaryDataByIDsResponse = await self._data_client.AddTagsToBinaryDataByIDs(request, metadata=self._metadata)
```
stuqdog, line 89: oh tiny nit here, `s/AddTagstoBinaryDataByIDsResponse/AddTags**T**oBinaryDataByIDsResponse`

8ashar: oouh nice catch

PR #335, opened 2023-06-28
RSDK-3788 - add getkinematics functionality in arm

purplenicole730:


src/viam/components/arm/arm.py
```diff
@@ -106,3 +106,17 @@ async def is_moving(self) -> bool:
...
+    async def get_kinematics(self, *, timeout: Optional[float] = None) -> Tuple[KinematicsFileFormat.ValueType, bytes]:
+        """
+        Get the kinematics information associated with the arm.
+
+        Returns:
+            Tuple[KinematicsFileFormat.Valuetype, bytes]:
+            (
+                format: The kinematics of the component in URDF format or Viam's kinematic parameter format (spatial vector algebra),
+                bytes: The byte contents of the file
+            )
```
njooma, line 120: You should run `make documentation` and double check to see if this renders in the documentation as expected

examples/module/src/module_arm/module_arm.py
```diff
@@ -0,0 +1,96 @@
...
+from typing import Any, ClassVar, Dict, Mapping, Optional, Self, Tuple
+from viam.components.arm import Arm, JointPositions, KinematicsFileFormat, Pose
+from viam.operations import run_with_operation
+from viam.proto.app.robot import ComponentConfig
+from viam.proto.common import ResourceName
+from viam.resource.base import ResourceBase
+from viam.resource.types import Model, ModelFamily
+
+
+class ModuleArm(Arm):
```
cheukt, line 12: would name it something else - MyArm or CustomArm, since it being a modular arm is not very descriptive

njooma (approved):
LGMT! Just one comment about documentation

examples/module/src/module_arm/module_arm.py
```diff
@@ -0,0 +1,96 @@
...
+
+    async def get_joint_positions(self, extra: Optional[Dict[str, Any]] = None, **kwargs) -> JointPositions:
+        return self.joint_positions
+
+    @run_with_operation
+    async def move_to_joint_positions(self, positions: JointPositions, extra: Optional[Dict[str, Any]] = None, **kwargs):
+        operation = self.get_operation(kwargs)
+
+        self.is_stopped = False
+        self.joint_positions = positions
```
raybjork, line 72: Same here: should this be after the sleeps?

examples/module/src/module_arm/module_arm.py
```diff
@@ -0,0 +1,96 @@
...
+from viam.operations import run_with_operation
+from viam.proto.app.robot import ComponentConfig
+from viam.proto.common import ResourceName
+from viam.resource.base import ResourceBase
+from viam.resource.types import Model, ModelFamily
+
+
+class ModuleArm(Arm):
+    # Subclass the Viam Arm component and implement the required functions
+    MODEL: ClassVar[Model] = Model(ModelFamily("rdk", "arm"), "module-arm")
```
cheukt, line 14: maybe `Model(ModelFamily("acme", "demo"), "name of this class")` to stay consistent with the other custom resources

examples/module/src/module_arm/module_arm.py
```diff
@@ -0,0 +1,96 @@
...
+    async def move_to_position(
+        self,
+        pose: Pose,
+        extra: Optional[Dict[str, Any]] = None,
+        **kwargs,
+    ):
+        operation = self.get_operation(kwargs)
+
+        self.is_stopped = False
+        self.position = pose
```
raybjork, line 51: Looks like if you're trying to mock this it should be set after the sleeps?

examples/module/src/module_arm/module_arm.py
```diff
@@ -0,0 +1,96 @@
...
+                await self.stop()
+                break
+
+        self.is_stopped = True
+
+    async def get_joint_positions(self, extra: Optional[Dict[str, Any]] = None, **kwargs) -> JointPositions:
+        return self.joint_positions
+
+    @run_with_operation
+    async def move_to_joint_positions(self, positions: JointPositions, extra: Optional[Dict[str, Any]] = None, **kwargs):
```
raybjork, line 68: Not sure what you're trying to accomplish with these functions.  If its just for mocks I think this is fine, but joint positions are coupled with end position so if theres another purpose we would want to consider that

examples/server/v1/components.py
```diff
@@ -90,6 +92,9 @@ async def stop(self, extra: Optional[Dict[str, Any]] = None, **kwargs):
     async def is_moving(self):
         return not self.is_stopped
 
+    async def get_kinematics(self) -> Tuple[KinematicsFileFormat.ValueType, bytes]:
```
cheukt, line 95: probably needs the **kwargs too

examples/server/v1/components.py
```diff
@@ -63,6 +64,7 @@ def __init__(self, name: str):
         )
         self.joint_positions = JointPositions(values=[0, 0, 0, 0, 0, 0])
         self.is_stopped = True
+        self.kinematics = (KinematicsFileFormat.KINEMATICS_FILE_FORMAT_SVA, b"\x00\x01\x02")
```
raybjork, line 67: Why an ExampleArm in addition to the ModuleArm?

raybjork (approved):
Reviewed from the perspective of the motion team rather than looking at code correctness/quality.  A couple questions, but nothing big enough to hold this up over. Thanks for doing this!

cheukt (changes requested):
can this PR be renamed `RSDK-3788 - XXXX`?
looks great! I would like to see some naming changes to be more consistent with other examples.

examples/module/src/module_arm/module_arm.py
purplenicole730: This is true! I agree with this sentiment. I copied and pasted this from an example module arm, but will definitely make this change. Thanks for catching!

examples/module/src/module_arm/module_arm.py
purplenicole730: Just for mocks!

examples/server/v1/components.py
purplenicole730: This `ExampleArm` is just another part of the `server` example that includes an example of every component we have. Almost, if not all, the values are arbitrary. The ticket specifically asked for a better arm module, which is what `ModuleArm` is for.

src/viam/components/arm/arm.py
purplenicole730: Thanks for the tip! I made the changes so that it now looks like this in the docs: 
image
The only nit I have is that I don't think `tuple containing` needs to have that highlight, but I can't seem to get rid of it without making it look weird. If you have tips, please let me know!

examples/module/src/arm/__init__.py
```diff
@@ -0,0 +1,8 @@
+"""
+This file registers the ModuleArm model with the Python SDK.
+"""
+
+from viam.resource.registry import Registry, ResourceCreatorRegistration
+from .my_arm import MyArm
+
+Registry.register_resource_creator(MyArm.SUBTYPE, MyArm.MODEL, ResourceCreatorRegistration(MyArm.new))
```
cheukt, line 8: Should it be Arm.SUBTYPE here to be consistent with what is in `main.py`?

examples/module/src/arm/my_arm.py
```diff
@@ -0,0 +1,96 @@
...
+            x=0,
+            y=0,
+            z=0,
+            o_x=0,
+            o_y=0,
+            o_z=1,
+            theta=0,
+        )
+
+        # Startinjoint positions
```
cheukt, line 28: Startin -> Starting

src/viam/components/arm/arm.py
cheukt: I think I prefer
```suggestion
        Returns:
            Tuple[KinematicsFileFormat.ValueType, bytes]:

                - KinematicsFileFormat.ValueType:
                  The format of the file, either in URDF format or Viam's kinematic parameter format (spatial vector algebra).

                - bytes:
                  The byte contents of the file.
```
which looks like 
Screen Shot 2023-06-29 at 17 40 59 PM
 let me know if you think a different format will look better!

cheukt (changes requested):
Can you update the example modules readme to reference this new example arm?

examples/module/src/arm/__init__.py
purplenicole730: Yes, it should. Making the change!

cheukt (approved):

PR #337, opened 2023-06-28
RSDK-3634 - Data wrappers

stuqdog:


src/viam/app/data/client.py
```diff
@@ -44,26 +57,109 @@ async def tabular_data_by_filter(
...
-        raise NotImplementedError()
+        Args:
+            filter (viam.app.data.Filter): When supplied, the binary data to delete will be filtered based on the provided constraints.
+                If not provided, all data will be deleted. Exercise caution before using this option.
+        """
+        filter = filter if filter else Filter()
+        request = DeleteBinaryDataByFilterRequest(filter=filter)
+        response: DeleteBinaryDataByFilterResponse = await self._data_client.DeleteBinaryDataByFilter(request, metadata=self._metadata)
+
+    async def delete_binary_data_by_ids(self, file_ids: List[str]) -> None:
```
stuqdog, line 154: Still need to test this method but didn't want to block starting review, working with data team on getting a test set up.

stuqdog:
Forgot to make one last change! Changing to draft, @8ashar @njooma I'll ping when it's ready for review :)

stuqdog:
Okay @njooma @8ashar this is now ready for review. All methods have been tested locally.

src/viam/app/data/client.py
```diff
@@ -41,37 +55,149 @@ def __init__(self, channel: Channel, metadata: str):
...
+            binary_ids (List[viam.proto.app.BinaryID]): The binary IDs of the data to be deleted. Must be non-empty.
+
+        Raises:
+            GRPCError: if no binary_ids are provided.
+        """
+        request = DeleteBinaryDataByIDsRequest(binary_ids=binary_ids)
+        _: DeleteBinaryDataByIDsResponse = await self._data_client.DeleteBinaryDataByIDs(request, metadata=self._metadata)
 
-    async def add_tags_to_binary_data_by_file_ids(self, file_ids: Optional[List[str]], tags: Optional[List[str]]) -> None:
+    async def add_tags_to_binary_data_by_binary_ids(self, binary_ids: Optional[List[str]], tags: Optional[List[str]]) -> None:
```
stuqdog, line 194: this is an accidental fly-by from running `s/file_ids/binary_ids` 😅

src/viam/app/data/client.py
```diff
@@ -41,37 +55,149 @@ def __init__(self, channel: Channel, metadata: str):
...
         raise NotImplementedError()
 
     async def add_tags_to_binary_data_by_filter(self, filter: Optional[Filter], tags: Optional[List[str]]) -> None:
         raise NotImplementedError()
 
-    async def remove_tags_from_binary_data_by_file_ids(self, file_ids: Optional[List[str]], tags: Optional[List[str]]) -> None:
+    async def remove_tags_from_binary_data_by_binary_ids(self, binary_ids: Optional[List[str]], tags: Optional[List[str]]) -> None:
```
stuqdog, line 200: also an accidental fly-by, see above!

src/viam/app/data/client.py
```diff
@@ -41,37 +55,149 @@ def __init__(self, channel: Channel, metadata: str):
...
+        response: BinaryDataByIDsResponse = await self._data_client.BinaryDataByIDs(request, metadata=self._metadata)
+        if dest:
+            try:
+                file = open(dest, 'w')
+                file.write(f"{response.data}")
+            except Exception as e:
+                LOGGER.error(f"Failed to write binary data to file {dest}", exc_info=e)
+        return response.data
 
     async def delete_tabular_data_by_filter(self, filter: Optional[Filter]) -> None:
```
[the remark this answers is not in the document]
stuqdog: Good question! I meant to leave a comment to this effect but forgot to. I opted to not providing a default for the `delete_` methods because I don't want to make it easier for people to delete everything. By not providing a default, users are forced to create a filter. This still allows for a "delete everything" case but hopefully makes users think a bit before doing so.

src/viam/app/data/client.py
```diff
@@ -41,37 +55,149 @@ def __init__(self, channel: Channel, metadata: str):
...
+
+        Args:
+            filter (viam.app.data.Filter): When supplied, the tabular data to delete will be filtered based on the provided constraints.
+                If not provided, all data will be deleted. Exercise caution before using this option.
+        """
+        filter = filter if filter else Filter()
+        request = DeleteTabularDataByFilterRequest(filter=filter)
+        _: DeleteTabularDataByFilterResponse = await self._data_client.DeleteTabularDataByFilter(request, metadata=self._metadata)
 
     async def delete_binary_data_by_filter(self, filter: Optional[Filter]) -> None:
```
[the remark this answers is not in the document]
stuqdog: See above :)

src/viam/app/data/client.py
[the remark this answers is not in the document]
8ashar: Love it

8ashar (approved):
LGTM

src/viam/app/data/client.py
```diff
@@ -41,37 +55,149 @@ def __init__(self, channel: Channel, metadata: str):
 
     async def tabular_data_by_filter(
         self,
-        filter: Optional[Filter],
-        dest: Optional[str]
+        filter: Optional[Filter] = None,
```
njooma, line 58: If all these `filter` params are optional (here and in the other methods), I think I'd prefer renaming these to simply be `tabular_data` or `binary_data` (dropping the `by_filter`). Especially since filter is optional with a default to `None`. 

Thoughts?

stuqdog: Ooh, good point. Hmm... I agree that `{tabular|binary}_data` is a better name in the abstract, but I think I have a weak preference for keeping the names as they are. Changing would create inconsistency with the other data getter methods, the `*_by_filter` data delete methods, and the proto code itself. 

That said, I don't feel particularly strongly about this so happy to go with the shorter names if you do!

njooma: I think the difference is that `by_id` REQUIRES an ID (or list of IDs), whereas these do not have any required params. Also, I don't think it's necessary to exactly match the proto names -- I do think the names should be descriptive in the language they are in.

That being said, I also don't have a super strong preference. I just want to make sure we discuss this prior to pushing, because I don't want to make a breaking change once this is released.

stuqdog: Yeah, totally valid. I think my main focus was on consistency with the `delete_by_filter` methods (where the filter is technically optional but I opted for not providing a default because I don't want to make it too easy for folks to delete everything). So people do have to provide _something_ for their filter (even if it is a `None`).

I very much agree that the names should be descriptive for their language, so if you think dropping the `by_filter` from the getters (and possibly the deleters?) is more Python idiomatic then I will readily defer to that judgment.

njooma: Hmmm... the `delete_by_filter` really makes me think we should keep these `data_by_filter` as well... Ok this is fine, let's keep it for now and if we want to change it we can always alias it with a deprecation warning

njooma (approved):

PR #345, opened 2023-07-06
[RSDK 3637] Implement data_sync.proto API

8ashar:
This PR implements the last two methods of our data client wrapper meant to allow users to programmatically upload data to app. As another minor update, the PR also modifies all methods that delete or remove anything from app to--instead of returning None--return an integer denoting the number of items successfully deleted/removed.


src/viam/app/data/client.py
```diff
@@ -290,10 +320,41 @@ async def bounding_box_labels_by_filter(self, filter: Optional[Filter] = None) -
...
-    async def data_capture_upload(self, metadata: UploadMetadata, sensor_contents: Optional[List[SensorData]] = None) -> None:
-        raise NotImplementedError()
+    async def data_capture_upload(self, metadata: UploadMetadata , sensor_contents: List[SensorData]) -> None:
+        """Upload binary or tabular sensor data.
+
+        Sync binary or tabular data collected on a robot through a specific component (i.e. a motor) along with the relevant metadata with
+        app.viam.com.
+
+        Args:
+            metadata (viam.app.proto.datasync.UploadMetaData): Metadata relating the data collected with the corresponding robot part. Must
```
stuqdog, line 330: I appreciate the amount of detail you're putting into these docstrings, but I'm pretty sad that this info isn't available if I want to look at the type itself. Given that upload metadata is used for both upload methods, but which _parts_ of that data we care about are different, I think this is a good case for asking for just the data we need and then building an `UploadMetadata` object ourselves.

src/viam/app/data/client.py
```diff
@@ -290,10 +320,41 @@ async def bounding_box_labels_by_filter(self, filter: Optional[Filter] = None) -
...
 
-    # TODO(RSDK-3637): Implement
     async def file_upload(self, metadata: UploadMetadata, file_contents: Optional[FileData]) -> None:
-        raise NotImplementedError()
+        """Upload arbitrary file data.
+
+        Sync file data that may be stored on a robot along with the relevant metadata to app.viam.com.
+
+        Args:
+            metadata (viam.app.proto.datasync.UploadMetaData): Metadata for the file to be uploaded. Must specify a part ID.
```
stuqdog, line 349: I think we don't care about part ID for file upload, but do care about file name and extension?

src/viam/app/data/client.py
```diff
@@ -290,10 +320,41 @@ async def bounding_box_labels_by_filter(self, filter: Optional[Filter] = None) -
...
+            file_contents (List[viam.app.proto.datasync.FileData]): A FileData object containing binary data to upload as a file.
```
stuqdog, line 350: `FileData` is just `bytes` under the hood, so we should probably ask for that rather than the list of `FileData` so we can provide a better documentation experience.

src/viam/app/data/client.py
stuqdog: But also see above about just asking for what we need and making an `UploadMetadata` under the hood.

src/viam/app/data/client.py
```diff
@@ -290,10 +320,41 @@ async def bounding_box_labels_by_filter(self, filter: Optional[Filter] = None) -
...
+    async def data_capture_upload(self, metadata: UploadMetadata , sensor_contents: List[SensorData]) -> None:
+        """Upload binary or tabular sensor data.
+
+        Sync binary or tabular data collected on a robot through a specific component (i.e. a motor) along with the relevant metadata with
+        app.viam.com.
+
+        Args:
+            metadata (viam.app.proto.datasync.UploadMetaData): Metadata relating the data collected with the corresponding robot part. Must
+                specify a part ID, component type, component name, method name, and data type.
+            sensor_contents (List[viam.app.proto.datasync.SensorData]): A list of the data to be uploaded. If the metadata specifies a
```
stuqdog, line 332: This is kind of finicky ("if _this_ field is set to _that_ value, then we upload `foo`, otherwise ..."). Ideally users shouldn't have to parse logic like that in docstrings. Probably would be better to just ask for people to pass the individual parts of `SensorData` (timestamps, and then data which is either `bytes` or `Mapping[str, Any]`) and then construct a proper `SensorData` object from that.

stuqdog (changes requested):
Done with a first pass! Looks pretty good, though I think we may want to revisit the shape of the API.

src/viam/app/data/client.py
8ashar: I agree I think this would be a good change. The function signatures might get a little wonkier, but the benefit would be ease of use. It seems like the only place to get the definition of UploadMetadata is either through the .proto file that defines them or the .pyi generated from it - neither of which are super easy to decipher.

src/viam/app/data/client.py
8ashar: I also thought that we wouldn't need a part ID for this method, but when I run the function with either an invalid part id (or just without one), I receive a GRPCError that says (no robot part found). I haven't been able to to find where this is coming from in app.. I've only been able to find the string "no robot part found" in tests in app, so maybe I'll reach out to Alexa again and see if something else is up.

src/viam/app/data/client.py
8ashar: agreed

src/viam/app/data/client.py
8ashar: I like this too I'll get it added.

src/viam/app/data/client.py
8ashar: Okay I found where the error is coming up, a valid id is definitely required and used in this request.

src/viam/app/data/client.py
```diff
@@ -290,10 +320,41 @@ async def bounding_box_labels_by_filter(self, filter: Optional[Filter] = None) -
...
+                be nonempty.
+
+        Raises:
+            GRPCError: If a required data field is not defined.
+        """
+        request = DataCaptureUploadRequest(metadata=metadata, sensor_contents=sensor_contents)
+        _: DataCaptureUploadResponse = await self._data_sync_client.DataCaptureUpload(request, metadata=self._metadata)
 
-    # TODO(RSDK-3637): Implement
     async def file_upload(self, metadata: UploadMetadata, file_contents: Optional[FileData]) -> None:
```
stuqdog, line 343: One other thought: I think maintaining this version of `file_upload` makes sense, but also it would be great if we had a method that instead took as an argument a path to a file, and then we could convert its contents into bytes under the hood and infer the file name and extension and just upload that. Seems like a much more pleasant user experience.

stuqdog (changes requested):

src/viam/app/data/client.py
8ashar: Do you think it might be awkward to ask for three lists that have to line up by index? the sensor_contents parameter is weird. If the user is uploading binary data, only a single SensorData object is required and allowed (i.e. one time_requested timestamp, one time_received timestamp, and then the binary data). On the other hand, if tabular binary data is being uploaded a list of SensorData files of any length > 0 is allowed, meaning we need to take in a list of time_requested timestamps and time_received timestamps and then the tabular data itself. Another caveat is that if it is tabular data, it has to ALL be tabular data. Do you think all of this is worth explaining in the docstrings? Or should we try something else. If we did this, I'm thinking the individual fields of the SensorData objects as parameters would look something like this:

foo(
    time_received: Union[TimeStamp, List[TimeStamp]],
    time_requested: Union[TimeStamp, List[TimeStamp]],
    data: Union[bytes, List[Mapping[str, Any]]]
).

Another thing to consider is that we still probably could avoid getting to verbose in the docstrings because there is pretty good error handling on the receiving end in app. The error messages make it pretty obvious if someone passes lots of binary data when they weren't supposed to, etc.

8ashar: Maybe we could break this down into two different methods, one for binary data and one for tabular data or something

src/viam/app/data/client.py
8ashar: I like this. Do you want to overload the function? Also quick question: by "this version" do you mean the new one where we create the metadata and filedata objects ourselves under the hood? Or the old version where the client passes them as already existing arguments.

src/viam/app/data/client.py
8ashar: ^ just to make it more clear when a list of timestamps is needed vs when just 1 is needed without having a ton of logic explained in the docstrings or even in the code

src/viam/app/data/client.py
```diff
@@ -290,10 +328,226 @@ async def bounding_box_labels_by_filter(self, filter: Optional[Filter] = None) -
...
+        """Upload arbitrary file data.
+
+        Sync file data that may be stored on a robot along with the relevant metadata to app.viam.com.
+
+        Args:
+            part_id (str): Part ID of the component associated with the file.
+            component_type (Optional[str]): Optional type of the component associated with the file (i.e., "movement_sensor").
+            component_name (Optional[str]): Optional name of the component associated with the file.
+            method_name (Optional[str]): Optional name of the method associated with the file.
+            file_name (Optional[str]): Optional name of the file.
```
stuqdog, line 477: Naively, I would think that a method named `file_upload` would require a file name/extension. What happens if they aren't provided? Can we clarify in the comments?

src/viam/app/data/client.py
```diff
@@ -290,10 +328,226 @@ async def bounding_box_labels_by_filter(self, filter: Optional[Filter] = None) -
...
+            method_parameters (Optional[str]): Optional dicitonary of the method parameters. No longer in active use.
```
stuqdog, line 478: See above on use of this param.

src/viam/app/data/client.py
```diff
@@ -290,10 +328,226 @@ async def bounding_box_labels_by_filter(self, filter: Optional[Filter] = None) -
...
+            file_extension (Optional[str]): Optional file extension.
+            tags (Optional[List[str]]): Optional list of tags to allow for tag-based filtering.
```
stuqdog, line 480: how does `tags` work here? since it's an upload and not fetching anything, it doesn't really make sense to me what `tag-based filtering` is.

src/viam/app/data/client.py
```diff
@@ -290,10 +328,226 @@ async def bounding_box_labels_by_filter(self, filter: Optional[Filter] = None) -
...
+        """Upload arbitrary file data.
+
+        Sync file data that may be stored on a robot along with the relevant metadata to app.viam.com.
+
+        Args:
+            part_id (str): Part ID of the component associated with the file.
+            component_type (Optional[str]): Optional type of the component associated with the file (i.e., "movement_sensor").
+            component_name (Optional[str]): Optional name of the component associated with the file.
+            method_name (Optional[str]): Optional name of the method associated with the file.
+            method_parameters (Optional[str]): Optional dictionary of the method parameters. No longer in active use.
```
stuqdog, line 518: See above.

src/viam/app/data/client.py
```diff
@@ -290,10 +328,226 @@ async def bounding_box_labels_by_filter(self, filter: Optional[Filter] = None) -
...
+            tags (Optional[List[str]]): Optional list of tags to allow for tag-based filtering.
```
stuqdog, line 519: See above re: `tags` docstring.

stuqdog (changes requested):
Looking good! Some questions/comments.

src/viam/app/data/client.py
```diff
@@ -290,10 +328,226 @@ async def bounding_box_labels_by_filter(self, filter: Optional[Filter] = None) -
...
 
-    # TODO(RSDK-3637): Implement
-    async def file_upload(self, metadata: UploadMetadata, file_contents: Optional[FileData]) -> None:
-        raise NotImplementedError()
+        Args:
+            part_id (str): Part ID of the component used to captuer the data.
+            component_type (str): Type of the component used to capture the data (i.e., "movement_sensor").
+            component_name (str): Name of the component used to capture the data.
+            method_name (str): Name of the method used to capture the data.
+            method_parameters (Optional[Mapping[str, Any]]): Optional dictionary of method parameters. No longer in active use.
```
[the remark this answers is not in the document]
8ashar: Yes I asked the data /ml team about this parameter and this is what they said:

"Hi! We’re no longer actively using this, but it was to pass in any method parameters to the method itself. Previously, this was used to pass in the mime_type into ReadImage. But it’s still relevant if in the future a collector is added for a method that captures differently depending on the parameters"

They suggested leaving it in as a possible parameter.

src/viam/app/data/client.py
```diff
@@ -290,10 +328,226 @@ async def bounding_box_labels_by_filter(self, filter: Optional[Filter] = None) -
...
+        )
+        metadata = UploadMetadata(
+            part_id=part_id,
+            component_type=component_type,
+            component_name=component_name,
+            method_name=method_name,
+            type=DataType.DATA_TYPE_BINARY_SENSOR,
+            file_name=None,  # Not used in app.
+            method_parameters=method_parameters if method_parameters else None,
+            file_extension=file_extension if file_extension else None,
```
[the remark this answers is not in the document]
8ashar: file_name is none because it doesn't seem to be used at all on the receiving end in app (i.e., it's not stored in app), but I left it in explicitly defined as None as a reminder that it is a parameter that exists in case something changes in the future. I left file_extension in as an option just because it is used in app, but I think we could totally get away with not including it. In app, really no processing is done on the file extension, but it is stored as meta data if it nonempty. What do you think?

src/viam/app/data/client.py
```diff
@@ -290,10 +328,226 @@ async def bounding_box_labels_by_filter(self, filter: Optional[Filter] = None) -
...
+
+        metadata = UploadMetadata(
+            part_id=part_id,
+            component_type=component_type,
+            component_name=component_name,
+            method_name=method_name,
+            type=DataType.DATA_TYPE_TABULAR_SENSOR,
+            file_name=None,  # Not used in app.
+            method_parameters=method_parameters if method_parameters else None,
+            file_extension=file_extension if file_extension else None,
```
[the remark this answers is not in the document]
8ashar: Same as above

src/viam/app/data/client.py
8ashar: If they are not provided, the data still gets uploaded and the name and extension fields will just use the empty string "" as their values.

src/viam/app/data/client.py
8ashar: I think it attaches tags to the data. You're right my wording here is confusing. I mean that in the future when trying to retrieve this data, you can filter by the tags added. I will rephrase.

src/viam/app/data/client.py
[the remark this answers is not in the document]
stuqdog: It seems really strange to include a file extension but not a file name, so I'm inclined to not include either. If app starts caring about it, we can add it back as an optional param with default without breaking anything.

stuqdog (approved):
LGTM!

njooma (approved):
LGTM. One last very minor ticket that it might be worth getting into this as well: https://viam.atlassian.net/browse/RSDK-3953

njooma:
If we want this PR in the upcoming release, then you will need to follow the hotfix steps in the release guide, since the current release candidate branch has already been cut.

stuqdog:
@njooma good call on `binary_data_by_ids`, I think that's being resolved in a separate PR.

8ashar:
@njooma @stuqdog I added it to this pr (the change to binary_data_by_ids())! It was in another but that one got wonky with its commit history and was being weird so I closed it and just added it here to make life easier. The spelling errors in the rest of the docstrings will live in another pr that is shortly away just to make conflict resolution easier.

PR #346, opened 2023-07-06
RSDK-3922 No hardcoding credential type in Python

jckras:
Credential type is currently hard-coded as robot-location-secret, but we’re starting to also use robot-secret. We should update to make this a configurable value. We made this change first in cpp-sdk but also needs to be done in python-sdk.

### DIFF
--- a/examples/module/client.py
+++ b/examples/module/client.py
@@ -9,8 +9,11 @@ from viam.rpc.dial import Credentials, DialOptions
 
 
 async def connect():
-    creds = Credentials(type="<your authentication type here>", payload="<your authentication payload here>")
-    opts = RobotClient.Options(refresh_interval=0, dial_options=DialOptions(credentials=creds), log_level=logging.DEBUG)
+    creds = Credentials(type="<your authentication type here>",
+                        payload="<your authentication payload here>")
+    opts = RobotClient.Options(refresh_interval=0,
+                               dial_options=DialOptions(credentials=creds),
+                               log_level=logging.DEBUG)
     return await RobotClient.at_address("<your robot uri here>", opts)
 
 
--- a/examples/module/client.py
+++ b/examples/module/client.py
@@ -31,16 +31,16 @@ async def main():
     resp = await gizmo.do_two(False)
     print("do_two result:", resp)
 
-    # Certain types of streams are not currently supported by the viam-rust-utils library that python uses internally.
-    # If you are not using WebRTC, you can uncomment the following lines.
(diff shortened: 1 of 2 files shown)


src/viam/rpc/dial.py
```diff
@@ -191,7 +191,7 @@ def __init__(self) -> None:
         self._lib.init_rust_runtime.argtypes = ()
         self._lib.init_rust_runtime.restype = ctypes.c_void_p
 
-        self._lib.dial.argtypes = (ctypes.c_char_p, ctypes.c_char_p, ctypes.c_bool, ctypes.c_void_p)
+        self._lib.dial.argtypes = (ctypes.c_char_p, ctypes.c_char_p, ctypes.c_char_p, ctypes.c_bool, ctypes.c_void_p)
```
stuqdog, line 194: I think this line is too long and it's giving us a linter error. You can run `make lint` to confirm :)

src/viam/rpc/dial.py
```diff
@@ -30,7 +30,7 @@ class Credentials:
     Currently only supports robot location secret.
     """
 
-    type: Literal["robot-location-secret"]
+    type: str
```
stuqdog, line 33: If the number of auth types starts growing significantly we can revisit, but for now I think we still want this to be a `Literal` (don't want to tell people they can pass _anything_ if we weren't previously). But, instead of just the one, we should instead have a `Union` of `robot-location-secret` and `robot-secret` literals.

stuqdog (changes requested):
A couple small things, looks good though!

examples/module/client.py
```diff
@@ -9,9 +9,9 @@
 
 
 async def connect():
-    creds = Credentials(type="<your authentication type here>", payload="<your authentication payload here>")
+    creds = Credentials(type="robot-location-secret", payload="znm8jwc5ossauwmy24x7b9ts7ptb750ajg8bzmxv9utkoic2")
```
stuqdog, line 12: We should not commit an actual robot uri or payload.

stuqdog (changes requested):
One small change needed, otherwise looks good!

examples/module/client.py
jckras: Ah thank you for catching that!

stuqdog (approved):
lgtm!

njooma (approved):
LGTM! Only thing -- Something is up with your autolinter, we have our max line length set to 140 (79 doesn't make sense when we all have bigger screens now). You should run `make lint format` and push those changes to this PR as well.

PR #349, opened 2023-07-10
[RSDK 3953] change binary_data_by_ids() return type

8ashar:
This PR updates binary_data_by_ids() so that instead of a list of protobuf messages, the method return a list of binary data--each corresponding to one of the BinaryID's provided. Also included in this PR is a series of spelling error fixes.


stuqdog:
The new changes here look great, thanks! But also this has all the commits from the `data_sync.proto` PR in addition to the small fix to `binary_data_by_ids()` return type. Is this intentional?

PR #353, opened 2023-07-12
RSDK-3836 - update documentation

purplenicole730:
The docs have been updated to be more consistent with the existing `examples` in the Python SDK as well as the official Viam docs.
The changes include:
- adding a simpler `Sensor` example and adding the `Arm` example as an additional example at the bottom of the `Modular Resources` section
- suggesting users to go to modular resources before trying custom remotes
- updating the module example's `README` to mention the new Arm modular component
- in the `README`, point readers to go to the Python docs if they want to see a more in-depth explanation on how to create a modular component from an existing resource
- basic wording fixes

NOTE: Some of the wording is based off of the actual Viam docs for consistency.
The custom resource `Gizmo` and `Summation` module examples were not added to the docs on purpose. It seemed too repetitive without a lot of payoff, as it would be wordy for something that can be more easily shown in GitHub the way it is now. The docs have been updated to point a user to the example, and the `README` has also been updated to point a reader to the docs for existing resources.


docs/examples/example.ipynb
```diff
@@ -327,58 +303,54 @@
...
     "```\n",
     "\n",
     "After the resource has a defined model and creator function, the resource must be registered to the `Registry`. This can be done using [`register_resource_creator()`](https://python.viam.dev/autoapi/viam/resource/registry/index.html#viam.resource.registry.Registry.register_resource_creator) and passing a [`ResourceCreatorRegistration`](https://python.viam.dev/autoapi/viam/resource/registry/index.html#viam.resource.registry.ResourceCreatorRegistration) object with the creator function as a parameter.\n",
     "\n",
-    "In the same `modular-arm/src` directory, create a new file called `__init__.py`. In it, call `Registry.register_resource_creator()` with the subtype of the resource that was subclassed (in this case, `Arm`), the model name, and a `ResourceCreatorRegistration` object with the `new` function defined in the modular resource.\n",
+    "In the same `src` directory, create a new file called `__init__.py`. In it, call `Registry.register_resource_creator()` with the subtype of the resource that was subclassed, the model name, and a `ResourceCreatorRegistration` object with the creator function defined in the modular resource.\n",
```
andf-viam, line 330:
```suggestion
    "In the same `src` directory, create a new file named `__init__.py`. In it, call `Registry.register_resource_creator()` with the subtype of the resource that was subclassed, the model name, and a `ResourceCreatorRegistration` object with the creator function defined in the modular resource.\n",
```

docs/examples/example.ipynb
```diff
@@ -327,58 +303,54 @@
...
     "\n",
     "\n",
     "```python\n",
```
andf-viam, line 333: Should I copy this code into my `__init__.py` file? If so, please introduce the code with this guidance. Perhaps:

```
"Copy the code below into your `__init__.py` file:\n",
```

docs/examples/example.ipynb
```diff
@@ -417,48 +389,164 @@
...
     "```\n",
     "\n",
-    "In this case, `<your-modular-directory-name>.<main-program-filename-without-extension>` would be replaced with `src.main`. \n",
-    "\n",
     "To make this shell script executable, run this in the Terminal:\n",
     "\n",
-    "```sudo chmod +x <FILEPATH>/exec.sh```\n",
+    "```sudo chmod +x <FILEPATH>/<FILENAME>```\n",
+    "\n",
+    "In this case, `<your-module-directory-name>.<main-program-filename-without-extension>` would be replaced with `src.main`. `<FILENAME>` would be `exec.sh`.\n",
```
andf-viam, line 396: The first sentence of this callout belongs beneath the `exec` above the `chmod` instead.

Is the `<FILENAME>` always going to be `exec.sh` ? If, I recommend removing the second sentence advising such, and updating the example directly:

```
"```sudo chmod +x <FILEPATH>/exec.sh```\n",
```

docs/examples/example.ipynb
```diff
@@ -417,48 +389,164 @@
...
    "attachments": {},
    "cell_type": "markdown",
    "metadata": {},
    "source": [
     "## Create custom remotes\n",
     "\n",
-    "While the main RDK is written in golang, you can create custom components in python and connect them to a robot as a `remote` component. This allows you to extend the functionality of a robot, or even create an entire robot exclusively in python.\n",
+    "**CAUTION** *[Modular resources](https://python.viam.dev/examples/example.html#create-custom-modules) are the preferred method of creating custom resource implementations for the Python SDK unless you are hosting viam-server on a non-Linux system or have another issue with compilation.*\n",
+    "\n",
+    "While the main RDK is written in golang, you can create custom components in python and connect them to a robot as a [`remote` component](https://docs.viam.com/extend/custom-components-remotes/). This allows you to extend the functionality of a robot, or even create an entire robot exclusively in python.\n",
```
andf-viam, line 549: Thanks for including this!

andf-viam (approved):
Awesome! I've noted some small optional language suggestions for your consideration. Thanks!

benjirewis (approved):
LGTM! Thanks 🧑‍🔧

docs/examples/example.ipynb
purplenicole730: It will not always be `exec.sh`. For example, the module example calls it `run.sh`. The naming is up to the user.

cheukt:
nit: pr name should be in form of `RSDK-XXXX blah`

cheukt (approved):

PR #358, opened 2023-07-17
RSDK-3050 - Add get_images

cheukt:
On server side, tries to encode all PIL images to jpgs, was that what you were suggesting? @bhaney 
otherwise pretty similar to get_image


src/viam/components/camera/camera.py
```diff
@@ -48,6 +49,21 @@ async def get_image(self, mime_type: str = "", *, timeout: Optional[float] = Non
...
 
+    @abc.abstractmethod
+    async def get_images(self, *, timeout: Optional[float] = None, **kwargs) -> Tuple[List[Union[Image, RawImage]], datetime]:
+        """Get simultaneous images from different sensors, along with associated metadata (just timestamp for now).
+        It's not for getting a time series of images from the same sensor.
+
+        Returns:
+            Tuple[List[Union[Image, RawImage]], datetime]:
+                - List[Union[Image, RawImage]]:
+                  The list of images returned from the camera system.
```
bhaney, line 60: This is missing the "source_name" associated with the image, which is a string stating the sensor's name that produced the image

src/viam/components/camera/client.py
```diff
@@ -34,11 +47,18 @@ def __init__(self, name: str, channel: Channel):
...
-            image = RawImage(response.image, response.mime_type)
-            return image
-        return Image.open(BytesIO(response.image), formats=LIBRARY_SUPPORTED_FORMATS)
+        return get_image_from_response(response.image, response_mime_type=response.mime_type, request_mime_type=request.mime_type)
+
+    async def get_images(self, *, timeout: Optional[float] = None) -> Tuple[List[Union[Image.Image, RawImage]], datetime]:
+        request = GetImagesRequest(name=self.name)
+        response: GetImagesResponse = await self.client.GetImages(request, timeout=timeout)
+        imgs = []
+        for img_data in response.images:
```
bhaney, line 56: you'll need to store the returned name of the sensor somewhere

src/viam/components/camera/service.py
```diff
@@ -61,6 +66,31 @@ async def GetImage(self, stream: Stream[GetImageRequest, GetImageResponse]) -> N
...
+        img_bytes_lst = []
+        for img in images:
+            try:
+                # encode all pil images as jpg as default
+                mimetype = CameraMimeType.JPEG
+                if isinstance(img, RawImage):
+                    mimetype, _ = CameraMimeType.from_lazy(img.mime_type)
+                fmt = mimetype.to_proto()
+                img_bytes = mimetype.encode_image(img)
+                img_bytes_lst.append(Image(source_name=name, format=fmt, image=img_bytes))
```
bhaney, line 86: the source_name might not be the camera name- it should be whatever string the server associated with the image.

src/viam/media/video.py
```diff
@@ -105,3 +106,37 @@ def is_supported(cls, mime_type: str) -> bool:
...
+            format (Format.ValueType): The mimetype in a proto enum.
+
+        Returns:
+            Self: The mimetype.
+        """
+        mimetypes = {
+            Format.FORMAT_RAW_RGBA: CameraMimeType.VIAM_RGBA,
+            Format.FORMAT_JPEG: CameraMimeType.JPEG,
+            Format.FORMAT_PNG: CameraMimeType.PNG,
+            Format.FORMAT_RAW_DEPTH: CameraMimeType.PCD,
```
bhaney, line 124: FORMAT_RAW_DEPTH is actually the MIME type "image/vnd.viam.dep", it is not the PCD format "pointcloud/pcd".

bhaney: https://github.com/viamrobotics/rdk/blob/main/utils/mime.go#L19

bhaney (changes requested):
Default encoding of PIL images to JPEG makes sense, just two things:

- need to add the viam raw depth format "image/vnd.viam.dep" to the python ENUM so that it can be matched to the format of FORMAT_RAW_DEPTH.
- `source_name` needs to be added to the server and client for each of the the returned images in the list
-

src/viam/media/video.py
cheukt: updated

njooma:
@bhaney @cheukt

I added 2 new types: `ViamImage` and `NamedImage` (which is a subclass of `ViamImage`). `ViamImage` will eventually be the return type of all our camera functions that return an image. `ViamImage` should have a number of benefits:

1. *Unify the return type.* Instead of `Union[RawImage, PIL.Image]`, it can now just be `ViamImage`.
2. *Default to lazy decoding.* `ViamImage` contains a computed property named `image`, which will decode the image based on the mime type (if possible), and also cache the result so subsequent calls do not require decoding. And if the user manually changes the mime type, the cache will be invalidated.

Because switching to `ViamImage` everywhere else will be a breaking change, we will do it separately.

`NamedImage` is a simple subclass of `ViamImage`. It has all the same benefits with an additional `name` attribute. This might be overkill (it might perhaps just be useful to have an optional `name` param on `ViamImage`, e.g. `name: Optional[str]`, but I think I like the optional less than the additional class definition).

I also changed the return type of the `get_images` function to return the full `ResponseMetadata` object, rather than just the timestamp. This will be useful in case we expand `ResponseMetadata` in the future, as it won't require a breaking change to add the additional fields. 

NB. I accidentally sorted all the imports in the `tests` directory, which is why so many of those files changed. But they're just import ordering and can safely be ignored.

examples/server/v1/components.py
```diff
@@ -332,6 +333,9 @@ def __del__(self):
     async def get_image(self, mime_type: str = "", **kwargs) -> Image.Image:
         return self.image.copy()
 
+    async def get_images(self, timeout: Optional[float] = None, **kwargs) -> Tuple[List[Union[Image.Image, RawImage]], datetime]:
```
cheukt, line 336: update signature here too

src/viam/components/camera/camera.py
```diff
@@ -48,6 +50,21 @@ async def get_image(self, mime_type: str = "", *, timeout: Optional[float] = Non
         """
         ...
 
+    @abc.abstractmethod
+    async def get_images(self, *, timeout: Optional[float] = None, **kwargs) -> Tuple[List[NamedImage], ResponseMetadata]:
+        """Get simultaneous images from different sensors, along with associated metadata.
+        This should not be used for getting a time series of images from the same sensor.
+
+        Returns:
+            Tuple[List[NamedImage], datetime]:
```
cheukt, line 59: datetime->ResponseMetadata

src/viam/components/camera/camera.py
```diff
@@ -48,6 +50,21 @@ async def get_image(self, mime_type: str = "", *, timeout: Optional[float] = Non
...
+                - List[NamedImage]:
+                  The list of images returned from the camera system.
+
+                - ResponseMetadata:
```
cheukt, line 63: @bhaney is the idea that timestamp should always be passed? the issue with returning response metadata is that it is optional in responsemetadata. this also generally differs from the response in the rdk/cpp sdk

cheukt (commented):
generally looks good, but a questions for bijan

src/viam/components/camera/camera.py
bhaney: timestamp is definitely optional - there may be situations where that info is not available

cheukt: cool

bhaney (approved):
Nice work with the lazy encoding/decoding, makes it easier to deal with all of the different kinds of file types we have, and makes for faster transfer

cheukt:
can't approve but LGTM too, if tests with the intel realsense pass

njooma (approved):

PR #359, opened 2023-07-18
[RSDK-3897] Update Data Docs

8ashar:
This PR makes a bunch of edits to the AppClient and DataClient docstrings and adds a section to the example Jupyter notebook showing how to use both classes.


Makefile
```diff
@@ -42,10 +42,13 @@ _test_docs:
 	pytest --nbmake "./docs"
 
 test_docs:
+	kill -9 `ps aux | grep "[d]ocs.examples._data_server" | awk '{print $$2}'`
```
stuqdog: This line seems to be causing tests to fail. I suspect adding `|| true` at the end (like the line below) should fix it?

src/viam/app/data/client.py
```diff
@@ -80,15 +80,15 @@ async def tabular_data_by_filter(
...
         dest: Optional[str] = None,
     ) -> List[Mapping[str, Any]]:
-        """Filter and download tabular data
+        """Filter and download tabular data.
 
         Args:
-            filter (viam.app.data.Filter): When supplied, the tabular data will be
-            filtered based on the provided constraints. Otherwise, all data is returned.
-            dest (str): When supplied, the tabular data will be saved to the provided file path.
+            filter (viam.proto.app.data.Filter): Optional `Filter` specifying tabular data to retrieve. No `Filter` implies all tabular
```
stuqdog, line 86: (q) have you confirmed that this change fixes the formatting issues that were showing up before?

docs/examples/_data_server.py
```diff
@@ -0,0 +1,135 @@
...
+    DataSyncServiceBase,
+    FileUploadRequest,
+    FileUploadResponse,
+)
+
+
+class DataServer(DataServiceBase, DataSyncServiceBase):
+    def __init__(self):
+        self.tabular_data_requested = False
+        self.tabular_response = [{"PowerPct": 0, "IsPowerred": False}, {"PowerPct": 0, "IsPowered": False}, {"Position": 0}]
```
stuqdog, line 51: (minor) s/IsPowerred/IsPowered

docs/examples/_data_server.py
```diff
@@ -0,0 +1,135 @@
...
+from viam.proto.app.datasync import (
+    DataCaptureUploadRequest,
+    DataCaptureUploadResponse,
+    DataSyncServiceBase,
+    FileUploadRequest,
+    FileUploadResponse,
+)
+
+
+class DataServer(DataServiceBase, DataSyncServiceBase):
```
stuqdog: I'm not entirely certain I understand what this class is doing. Can you clarify?

docs/examples/example.ipynb
```diff
@@ -7,9 +7,10 @@
    "source": [
     "# Example usage\n",
     "\n",
-    "The Viam SDK can be used in two ways:\n",
+    "The Viam SDK can be used in three ways:\n",
     "1. As a client to connect to a (remote or local) robot\n",
-    "2. Integrate custom resources to a robot\n",
+    "2. To create custom components and provide additional functionality to a robot\n",
```
stuqdog, line 12: This looks like it's reverting some changes that @purplenicole730 made with a recent documentation PR. is that intentional?

docs/examples/example.ipynb
```diff
@@ -898,6 +899,200 @@
...
+      "  }\n",
+      "}\n",
+      "]\n"
+     ]
+    }
+   ],
+   "source": [
+    "from viam.proto.app.data import Filter\n",
+    "\n",
+    "data = await data_client.tabular_data_by_filter(filter=Filter(robot_name=\"arduino\"))\n",
```
stuqdog, line 1008: Can we flesh out this `Filter` some more? Would love to show how all fields are set, or at a minimum the non-obvious fields (timestamps, tags).

stuqdog (changes requested):
First pass, some comments/questions but generally looking good!

docs/examples/_data_server.py
8ashar: Yes! This is another mock data service (similar to the one that had to be created in the testing ticket). This one is used for the example documentation. This server runs on a local port and pretends to be app and feeds our example data_client--the one included in the example notebook and so displayed on our docs page--fake data just to display. This is used because we can't actually authenticate and connect to app because we don't want to include a real location secret or robot uri in the docs pages or in the python repo.

Put differently, the example page writes a function that shows how to instantiate an AppClient. We don't actually call this function though, because we know it would fail since we aren't providing real location secrets or URIs. Instead, I depend on a fake mock AppClient that directly connects its DataClient to this server^ to give the reader the illusion of a data client. Any code that would 'reveal' this illusion is marked to be hidden and so won't show up on the html page, but it's still in the repo (and still runs and shouldn't cause any error at all).

That is all!

src/viam/app/data/client.py
8ashar: Working on this now! I haven't been able to get the relevant html files generated yet.

njooma (approved):

PR #364, opened 2023-07-20
Fix Docs Workflow

8ashar:
This PR removes the kill statements used when generating docs as described by the docs.yml file. Workflows get run in docker that are killed anyway, so we don't need to worry about linger processes.


stuqdog (approved):
lgtm!

PR #366, opened 2023-07-21
[RSDK-4113] add filter method

8ashar:
This PR adds a static helper method to the DataClient class that can be used to create a Filter object. The PR also updates the docs accordingly and includes implicit testing (i.e., there's no single unit test for the method, but the method is used to create a Filter and that filter is compared to /asserted against a separate, hard-coded Filter in other tests).


stuqdog (approved):
lgtm!

njooma (approved):

PR #372, opened 2023-07-26
[RSDK-3639] Create Location Client

8ashar:
This PR implements our LocationClient class, a class that exposes methods that make calls to the location service, and adds it as a property under out AppClient class. Right now, only the methods that've been labeled as high-priority have been implemented. The class also includes 3 shadow classes that essentially copy proto-defined classes but convert the Structs within those definitions to dictionaries and some Timestamps to Python datetimes. This specific PR also includes docstrings for the class. I'll make separate tickets and PR's for adding a snippet in the example notebook showing how to use the location_client and also some unit tests that depend on a mock location service.


8ashar:
One thing worth double checking is the format of the logs that I output as a result of passing a filepath 'dest' to get_robot_part_logs():
Screenshot 2023-07-26 at 3 13 33 PM

src/viam/app/location/client.py
```diff
@@ -0,0 +1,720 @@
...
+
+    class RobotPart:
+        """`RobotPart` shadow class.
+
+        Copy the attributes of a `viam.proto.app.RobotPart` object and convert and store its internal `Timestamp` objects as Python
+        `datetime` objects. It also converts the relevant `Struct` objects into dictionaries, mapping strings to items of type `Any`.
+        """
+
+        @classmethod
+        def robot_part_from_proto(cls, robot_part: RobotPartPB) -> Self:
```
njooma, line 86: rename this to simply `from_proto`

src/viam/app/location/client.py
```diff
@@ -0,0 +1,720 @@
...
+
+class LocationClient:
+    """gRPC client for method calls to app.
+
+    Constructor is used by `AppClient` to instantiate relevant service stub. Calls to `LocationClient`  methods should be made through
+    `AppClient`.
+    """
+
+    class RobotPart:
+        """`RobotPart` shadow class.
```
njooma, line 79: Probably restructure this to say:

> A class that mirrors the `RobotPart` proto message.

src/viam/app/client.py
```diff
@@ -43,12 +45,18 @@ async def create(cls, dial_options: DialOptions) -> Self:
...
+    _location_id: str
 
     @property
     def data_client(self) -> DataClient:
         """Insantiate and return a DataClient used to make `data` and `data_sync` method calls."""
         return DataClient(self._channel, self._metadata)
 
+    @property
+    def location_client(self) -> LocationClient:
+        """Insantiate and return a DataClient used to make `app` method calls."""
```
njooma, line 57:
```suggestion
        """Insantiate and return a LocationClient used to make `location` method calls to app."""
```

src/viam/app/location/client.py
```diff
@@ -0,0 +1,720 @@
...
+            """Create a `RobotPart` from the .proto defined `RobotPart`.
+
+            Args:
+                robot_part (viam.proto.app.RobotPart): The object to copy from.
+
+            Returns:
+                RobotPart: The `RobotPart`.
+            """
+            self = cls()
+            self.proto = robot_part
```
njooma, line 96: No need to save the proto

src/viam/app/location/client.py
```diff
@@ -0,0 +1,720 @@
...
+        robot_config: Optional[Mapping[str, Any]]
+        last_access: Optional[datetime]
+        user_supplied_info: Optional[Mapping[str, Any]]
+        main_part: bool
+        fqdn: str
+        local_fqdn: str
+        created_on: Optional[datetime]
+        secrets: List[SharedSecret]
+
+        proto: RobotPartPB
```
njooma, line 128: Make this a property that generates a new `RobotPartPB` every time, like

```python3
@property
def proto(self) -> RobotPartPB:
    return RobotPartPB(
        id = self.id,
        name = self.name,
        ...
    )
```

That way, if they change any of the attributes on the native `RobotPart`, it'll be reflected in the new `RobotPartPB`.

src/viam/app/location/client.py
```diff
@@ -0,0 +1,720 @@
...
+
+    class LogEntry:
+        """`LogEntry` shadow class.
+
+        Copy the attributes of a `viam.proto.app.LogEntry` object and convert and store its internal `Timestamp` object as a Python
+        `datetime` object. It also converts the relevant `Struct` objects into dictionaries, mapping strings to items of type `Any`.
+        """
+
+        @classmethod
+        def log_entry_from_proto(cls, log_entry: LogEntryPB) -> Self:
```
njooma, line 138: Same as above, rename to only `from_proto`

src/viam/app/location/client.py
```diff
@@ -0,0 +1,720 @@
...
+        main_part: bool
+        fqdn: str
+        local_fqdn: str
+        created_on: Optional[datetime]
+        secrets: List[SharedSecret]
+
+        proto: RobotPartPB
+
+    class LogEntry:
+        """`LogEntry` shadow class.
```
njooma, line 131: Same as above, say this mirrors the pb, not a shadow class

src/viam/app/location/client.py
```diff
@@ -0,0 +1,720 @@
...
+            """Create a `LogEntry` from the .proto defined `LogEntry`.
+
+            Args:
+                log_entry (viam.proto.app.LogEntry): The object to copy from.
+
+            Returns:
+                LogEntry: The `LogEntry`.
+            """
+            self = cls()
+            self.proto = log_entry
```
njooma, line 148: Same as above, don't save the proto

src/viam/app/location/client.py
```diff
@@ -0,0 +1,720 @@
...
+        host: str
+        level: str
+        time: Optional[datetime]
+        logger_name: str
+        message: str
+        caller: Optional[Mapping[str, Any]]
+        stack: str
+        fields: List[Mapping[str, Any]]
+
+        proto: LogEntryPB
```
njooma, line 168: Same as above, turn this into a property

src/viam/app/location/client.py
```diff
@@ -0,0 +1,720 @@
...
+
+    class Fragment:
+        """`Fragment` shadow class.
```
njooma, line 171: Ditto: mirrors

src/viam/app/location/client.py
```diff
@@ -0,0 +1,720 @@
...
+
+        Copy the attributes of a `viam.proto.app.Fragment` object and convert and store its internal `Timestamp` object as a Python
+        `datetime` object. It also converts the relevant `Struct` object into a dictionary, mapping strings to items of type `Any`.
+        """
+
+        @classmethod
+        def fragment_from_proto(cls, fragment: FragmentPB) -> Self:
```
njooma, line 178: Ditto: `from_proto`

src/viam/app/location/client.py
```diff
@@ -0,0 +1,720 @@
...
+        fragment: Optional[Mapping[str, Any]]
+        organization_owner: str
+        public: bool
+        created_on: Optional[datetime]
+        organization_name: str
+        robot_part_count: int
+        organization_count: int
+        only_used_by_owner: int
+
+        proto: FragmentPB
```
njooma, line 212: Ditto: `@property`

src/viam/app/location/client.py
```diff
@@ -0,0 +1,720 @@
...
+            """Create a `Fragment` from the .proto defined `Fragment`.
+
+            Args:
+                fragment (viam.proto.app.Fragment): The object to copy from.
+
+            Returns:
+                Fragment: The `Fragment`.
+            """
+            self = cls()
+            self.proto = fragment
```
njooma, line 188: Ditto: don't save

src/viam/app/location/client.py
```diff
@@ -0,0 +1,720 @@
...
+
+
+class LocationClient:
+    """gRPC client for method calls to app.
+
+    Constructor is used by `AppClient` to instantiate relevant service stub. Calls to `LocationClient`  methods should be made through
+    `AppClient`.
+    """
+
+    class RobotPart:
```
njooma, line 78: All of these class should live OUTSIDE of `LocationClient`, since they could be used anywhere/everywhere

src/viam/app/location/client.py
```diff
@@ -0,0 +1,720 @@
...
+        """Get a location.
+
+        Args:
+            location_id (Optional[str]): ID of the location to get. Defaults to the location ID provided at `LocationClient` instantiation.
+                If no default location ID was passed, a GRPCError will be thrown.
+
+        Raises:
+            GRPCError: If an invalid location ID is passed.
+
+        Location:
```
njooma, line 302: This should be `Returns:`

src/viam/app/location/client.py
```diff
@@ -0,0 +1,720 @@
...
+        """Delete a location.
+
+        Args:
+            location_id (str): ID of the location to delete. Must be specified.
+
+        Raises:
+            GRPCError: If an invalid location ID is passed.
+        """
+        request = DeleteLocationRequest(location_id=location_id)
+        _: DeleteLocationResponse = await self._location_client.DeleteLocation(request, metadata=self._metadata)
```
njooma, line 342: No need to set this value as an `_`, just raw `await` should be fine

src/viam/app/location/client.py
```diff
@@ -0,0 +1,720 @@
...
+        Args:
+            organization_id (str): ID of the organization to share the location with.
+            location_id (Optional[str]): ID of the location to be shared. Defaults to the current authorized location.
+
+        Raises:
+            GRPCError: Permission denied.
+        """
+        location_id = location_id if location_id else self._location_id
+        request = ShareLocationRequest(location_id=location_id, organization_id=organization_id)
+        _: ShareLocationResponse = await self._location_client.ShareLocation(request, metadata=self._metadata)
```
njooma, line 372: Same, no need to assign to `_`

src/viam/app/location/client.py
```diff
@@ -0,0 +1,720 @@
...
+        Args:
+            organization_id (str): ID of the organization to unshare the location with.
+            location_id (Optional[str]): ID of the location to be unshared. Defaults to the current authorized location.
+
+        Raises:
+            GRPCError: Permission denied.
+        """
+        location_id = location_id if location_id else self._location_id
+        request = UnshareLocationRequest(location_id=location_id, organization_id=organization_id)
+        _: UnshareLocationResponse = await self._location_client.UnshareLocation(request, metadata=self._metadata)
```
njooma, line 386: Ditto: no `_`

src/viam/app/location/client.py
```diff
@@ -0,0 +1,720 @@
...
+    async def location_auth(self):
+        raise NotImplementedError()
+
+    async def create_location_secret(self):
+        raise NotImplementedError()
+
+    async def delete_location_secret(self):
+        raise NotImplementedError()
+
+    async def get_robot(self, robot_id: str) -> Robot:
```
njooma, line 397: Why is this in the `location` client? This feels like it should just be in app?

src/viam/app/location/client.py
```diff
@@ -0,0 +1,720 @@
...
+        page_token = ""
+        logs = []
+
+        while True:
+            request = GetRobotPartLogsRequest(id=robot_part_id, errors_only=errors_only, filter=filter, page_token=page_token)
+            response: GetRobotPartLogsResponse = await self._location_client.GetRobotPartLogs(request, metadata=self._metadata)
+            if not response.next_page_token or len(response.next_page_token) == 0:
+                break
+            else:
+                logs += [LocationClient.LogEntry.log_entry_from_proto(log_entry=log_entry) for log_entry in response.logs]
```
njooma, line 490: If you're getting ALL the logs ever on a robot, this could be a huge list. Do we want to default to returning the entire list? Or should we use the pagination that's already built into the request?

njooma (changes requested):
Some changes:

1. Need tests
2. Why are robot/robot part/fragment/role/etc. methods here? These seem like app apis, not location APIs
3. Default to getting all logs seems like not a great idea. It could take a really really long time and also the memory footprint of the result could be very large. I think we paginate
4. Some basic formatting requests

src/viam/app/location/client.py
```diff
@@ -0,0 +1,711 @@
...
+
+    async def delete_robot_part_secret(self):
+        raise NotImplementedError()
+
+    async def list_robots(self, location_id: Optional[str] = None) -> List[Robot]:
+        """Get a list of all robots under the specified location.
+
+        Args:
+            location_id (Optional[str]): ID of the location to retrieve the robots from. Defaults to the location ID provided at
+                `LocationClient` instantiation.
```
stuqdog, line 581: what if no location ID is provided at `LocationClient` instantiation?

src/viam/app/location/client.py
```diff
@@ -0,0 +1,711 @@
...
+        """
+        request = DeleteRobotRequest(id=robot_id)
+        _: DeleteRobotResponse = await self._location_client.DeleteRobot(request, metadata=self._metadata)
+
+    async def list_fragments(self, organization_id: str, show_public: Optional[bool] = False) -> List[Fragment]:
+        """Get a list of fragments under the specified organization.
+
+        Args:
+            organization_id (str): ID of the organization under which to list the fragments.
+            show_public: Optional boolean specifiying whether or not to only include public fragments. Defaults to False.
```
stuqdog, line 649: I think it's probably safer to default to `True` here.

stuqdog (changes requested):
Nice, this is looking pretty good! Some requested semantic/documentation changes, and piggy backing on some of Naveed's points. I generally agree with his high level comments, and added some brief thoughts on splitting up the robot APIs into a separate client.

src/viam/app/location/client.py
```diff
@@ -0,0 +1,711 @@
...
+        raise NotImplementedError()
+
+    async def delete_organization_invite(self):
+        raise NotImplementedError()
+
+    async def resend_organization_invite(self):
+        raise NotImplementedError()
+
+    async def create_location(self, organization_id: str, name: str, parent_location_id: Optional[str] = None) -> Location:
+        """Create and name a location under the specified organization and parent location.
```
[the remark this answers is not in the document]
8ashar: As of right now, I am able to successfully call this method, as long as the org_id is from the org that the initial location I used to auth belongs to it. Should I assume this won't be the case then? Also I though if should work if we're authed as an org owner?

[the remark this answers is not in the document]
stuqdog: Ah interesting, I may have gotten myself confused then! Yeah, maybe we should just update `raises` to indicate that we raise if we pass an org_id that we don't have permissions in? 

I am curious how it will work if we try to create top level locations in the org after auth changes go in, since we'll no longer be an owner of the org. but we're figuring out solutions to that now so no need to dig into it.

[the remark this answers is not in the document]
8ashar: I'm also curious how this will play out when the auth changes take place. Not sure what to expect at this point.

8ashar:
@njooma @stuqdog Okay I made all the changes we've talked about. I've run successful local tests but have yet to write a formal testing suite- that'll come next, and then I'll make the necessary changes to our example notebook and other docs stuff. Thanks!

src/viam/app/app_client.py
```diff
@@ -0,0 +1,791 @@
...
+            List[viam.app.app_client.LogEntry]: The list of log entries.
+        """
+        num_pages = num_pages if num_pages is not None else 1
+        assert num_pages >= 0
+        errors_only = errors_only if errors_only else True
+        page_token = ""
+        logs = []
+
+        if num_pages == 0:
+            while True:
```
stuqdog, line 535: (minor) there's a bit of duplication in this code block. I think we could get rid of the `if num_pages == 0: ... else:` logic and reduce this to a single block by creating a `capture_all_logs = num_pages == 0` variable before the `while_true` and referencing that when deciding whether or not to break. Other approaches could work too! Not worth spending _too_ much time thinking on this, but worth considering.

src/viam/app/app_client.py
```diff
@@ -0,0 +1,791 @@
...
+                LOGGER.error(f"Failed to write robot part logs to file {dest}", exc_info=e)
+
+        return logs
+
+    async def _get_robot_part_logs(
+        self, robot_part_id: str, filter: str, errors_only: bool, page_token: str, logs: List[LogEntryPB]
+    ) -> str:
+        request = GetRobotPartLogsRequest(id=robot_part_id, errors_only=errors_only, filter=filter, page_token=page_token)
+        response: GetRobotPartLogsResponse = await self._location_client.GetRobotPartLogs(request, metadata=self._metadata)
+        logs += [LogEntry.from_proto(log) for log in response.logs]
```
stuqdog, line 573: I suspect this is python idiomatic so no need to change anything, but: let the record show that passing a mutable data structure to a function, mutating it, and returning some other value makes me sad :(

src/viam/app/app_client.py
```diff
@@ -0,0 +1,791 @@
...
+        """Mark the specified robot part for restart.
+
+        Args:
+            robot_part_id (str): ID of the robot part to mark for restart.
+
+        Raises:
+            GRPCError: If an invalid robot part ID is passed.
+        """
+        request = MarkPartForRestartRequest(part_id=robot_part_id)
+        _: MarkPartForRestartResponse = await self._location_client.MarkPartForRestart(request, metadata=self._metadata)
```
stuqdog, line 645: for consistency we should not type annotate the unused return here

src/viam/app/app_client.py
```diff
@@ -0,0 +1,791 @@
...
+        """Delete the specified robot.
+
+        Args:
+            robot_id (str): ID of the robot to delete.
+
+        Raises:
+            GRPCError: If an invalid robot ID is passed.
+        """
+        request = DeleteRobotRequest(id=robot_id)
+        _: DeleteRobotResponse = await self._location_client.DeleteRobot(request, metadata=self._metadata)
```
stuqdog, line 722: see above re: type annotating unused return

stuqdog (changes requested):
Looking good! Some comments/requested changes but mostly pretty minor.

src/viam/app/app_client.py
8ashar: Lolz

src/viam/app/app_client.py
```diff
@@ -0,0 +1,791 @@
...
+
+    async def list_fragments(self, organization_id: str, show_public: Optional[bool] = True) -> List[Fragment]:
+        """Get a list of fragments under the specified organization.
+
+        Args:
+            organization_id (str): ID of the organization under which to list the fragments.
+            show_public: Optional boolean specifiying whether or not to only include public fragments. Defaults to True.
```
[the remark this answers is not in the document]
8ashar: the public fragments actually aren't always shown! If show_public is true, we only see the public ones. If show_public is false, we only see the private ones.

src/viam/app/app_client.py
```diff
@@ -0,0 +1,791 @@
...
+        Returns:
+            viam.proto.app.Location: The location.
+        """
+        location_id = location_id if location_id else self._location_id
+        request = GetLocationRequest(location_id=location_id)
+        response: GetLocationResponse = await self._location_client.GetLocation(request, metadata=self._metadata)
+        return response.location
+
+    async def update_location(
+        self, location_id: str, name: Optional[str] = None, parent_location_id: Optional[str] = None, region: str = None  # Not used in app.
```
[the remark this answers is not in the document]
8ashar: I updated the docstring to just say that it currently isn't explicitly used in app. Lmk if you'd rather I just take it out.

src/viam/app/app_client.py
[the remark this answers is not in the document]
stuqdog: oh! that's surprising. can we update the comment on `show_public` to clarify that?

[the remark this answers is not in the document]
8ashar: Yup lemme do that

stuqdog (commented):
One comment on the `region` question. Otherwise, looks good! We should hold off on release until testing is done.

src/viam/app/app_client.py
```diff
@@ -356,7 +360,8 @@ async def update_location(
                 change).
             parent_location_id(Optional[str]): Optional new parent location to move the location under. Defaults to the empty string ""
                 (i.e., no new parent location is assigned).
-            name (Optional[str]): Optional new GCS region to associate the location with.
+            name (Optional[str]): Optional new GCS region to associate the location with. Currently, this parameter is not explicitly used
```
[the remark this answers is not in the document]
8ashar: oh oops I didn't even see that I had given them the same name on accident. I'll get rid of region for now though.

src/viam/app/app_client.py
```diff
@@ -0,0 +1,784 @@
...
+            location_id (Optional[str]): ID of the location to be unshared. Defaults to the current authorized location.
+
+        Raises:
+            GRPCError: Permission denied. A location can only be unshared by users who are owners of both organizations involved. An
+                `AppClient` cannot be an authorized owner of two organizations. Permission will always be denied when calling this method. A
+                GRPCError is also thrown when attempting to unshare a location with an organization it is not already shared with.
+        """
+        location_id = location_id if location_id else self._location_id
+        request = UnshareLocationRequest(location_id=location_id, organization_id=organization_id)
+        await self._location_client.UnshareLocation(request, metadata=self._metadata)
```
cheukt, line 430: I think these implementations should be removed for now since we don't plan on supporting it right now - we can always add them in a pinch if we need to

src/viam/app/app_client.py
```diff
@@ -0,0 +1,791 @@
...
+        Args:
+            channel (grpclib.client.Channel): connection to app.
+            metadata (Mapping[str, str]): Required authorization token to send requests to app.
+            location_id (Optional[str]): Default location ID.
+        """
+        self._metadata = metadata
+        self._location_client = AppServiceStub(channel)
+        self._location_id = location_id
+
+    _location_client: AppServiceStub
```
njooma, line 261: Rename this to be `_app_client`

src/viam/app/app_client.py
```diff
@@ -0,0 +1,791 @@
...
+    async def delete_organization_member(self):
+        raise NotImplementedError()
+
+    async def delete_organization_invite(self):
+        raise NotImplementedError()
+
+    async def resend_organization_invite(self):
+        raise NotImplementedError()
+
+    async def create_location(self, organization_id: str, name: str, parent_location_id: Optional[str] = None) -> Location:
```
njooma, line 314: This might be extremely opinionated, but I feel like the ordering of these args should be different. I think it should be `name` first, then `org` and `parent`. This new ordering is about the new location, and then where that new location lives. The current argument order is where the location lives, then the name, and then where the location lives again.

src/viam/app/app_client.py
```diff
@@ -0,0 +1,791 @@
...
+            GRPCError: If an invalid robot ID is passed.
+
+        Returns:
+            List[viam.app.app_client.RobotPart]: The list of robot parts.
+        """
+        request = GetRobotPartsRequest(robot_id=robot_id)
+        response: GetRobotPartsResponse = await self._location_client.GetRobotParts(request, metadata=self._metadata)
+        return [RobotPart.from_proto(robot_part=part) for part in response.parts]
+
+    async def get_robot_part(self, robot_part_id: Optional[str], dest: Optional[str] = None, indent: Optional[int] = None) -> RobotPart:
```
njooma, line 476: `robot_part_id` shouldn't be optional

src/viam/app/app_client.py
```diff
@@ -0,0 +1,791 @@
...
+
+        return RobotPart.from_proto(robot_part=response.part)
+
+    async def get_robot_part_logs(
+        self,
+        robot_part_id: str,
+        filter: Optional[str] = None,
+        dest: Optional[str] = None,
+        errors_only: Optional[bool] = None,
+        num_pages: Optional[int] = None,
```
njooma, line 509: Do we know how many entries are per page? 
Also, if this defaults to 1, then make it default to 1. Don't make it `Optional[int] = None`, but simply `int = 1`

src/viam/app/app_client.py
```diff
@@ -0,0 +1,791 @@
...
+                LOGGER.err(f"Failed to write config JSON to file {dest}", exc_info=e)
+
+        return RobotPart.from_proto(robot_part=response.part)
+
+    async def get_robot_part_logs(
+        self,
+        robot_part_id: str,
+        filter: Optional[str] = None,
+        dest: Optional[str] = None,
+        errors_only: Optional[bool] = None,
```
njooma, line 508: Docstring says this defaults to `True`. So have it default to `True` in the signature.

src/viam/app/app_client.py
```diff
@@ -0,0 +1,791 @@
...
+        errors_only = errors_only if errors_only else True
+        page_token = ""
+        logs = []
+
+        if num_pages == 0:
+            while True:
+                next_page_token = await self._get_robot_part_logs(
+                    robot_part_id=robot_part_id, filter=filter, errors_only=errors_only, page_token=page_token, logs=logs
+                )
+                if not next_page_token or len(next_page_token) == 0:
```
njooma: if `next_page_token` is a string, then check if empty string, not if `len == 0`

src/viam/app/app_client.py
```diff
@@ -0,0 +1,791 @@
...
+                    break
+                page_token = next_page_token
+        else:
+            while True:
+                next_page_token = await self._get_robot_part_logs(
+                    robot_part_id=robot_part_id, filter=filter, errors_only=errors_only, page_token=page_token, logs=logs
+                )
+                num_pages -= 1
+                if (not next_page_token or len(next_page_token) == 0) or num_pages == 0:
```
njooma: Same here, check to see if `next_page_token` is empty string, not its length

src/viam/app/app_client.py
```diff
@@ -0,0 +1,791 @@
...
+                for log in logs:
+                    time = log.time
+                    level = log.level.upper()
+                    logger_name = log.logger_name.split(".")[0]
+                    file_name = log.caller["File"]
+                    line_number = int(log.caller["Line"])
+                    message = log.message
+                    file.write(f"{time}\t\t{level}\t{logger_name}\t{file_name}:{line_number}\t{message}\n")
+            except Exception as e:
+                LOGGER.error(f"Failed to write robot part logs to file {dest}", exc_info=e)
```
njooma, line 556: Might be useful to have the robot part ID here as well, in case there are multiple log functions running

src/viam/app/app_client.py
njooma: I agree, I dislike this as well. I don't think this should be an inout param. I think this function should return a tuple of `(List[LogEntry], str)`, and the calling function should do what it wants with the list of logs

tests/mocks/services.py
```diff
@@ -497,3 +612,179 @@ async def FileUpload(self, stream: Stream[FileUploadRequest, FileUploadResponse]
         assert request_file_contents is not None
         self.binary_data = request_file_contents.file_contents.data
         await stream.send_message(FileUploadResponse())
+
+
+class MockLocation(AppServiceBase):
```
njooma, line 617: This probably shouldn't be called `MockLocation` anymore

njooma (changes requested):

src/viam/app/app_client.py
8ashar: @njooma it maxes out at 100 log entries per page

njooma: ok then maybe instead of saying `num_pages` we do a max count of logs, and not num pages

stuqdog (changes requested):
forgot to submit this review yesterday, sorry! I see that some changes have happened since I looked; I'll take another look and submit any additional comments.

stuqdog (approved):

PR #378, opened 2023-08-07
RSDK-4313 - Include data request metadata

stuqdog:
#### Major changes
Create `BinaryData` and `TabularData` types to encompass all meaningful data/metadata from associated calls

#### Minor changes
Some small refactors to tests


docs/examples/_server.py
```diff
@@ -169,19 +172,49 @@
...
         self.tabular_data_requested = True
         _ = await stream.recv_message()
-        n = len(self.tabular_response)
-        tabular_structs = [Struct()] * n
-        for i in range(n):
-            tabular_structs[i].update(self.tabular_response[i])
-        await stream.send_message(TabularDataByFilterResponse(data=[TabularData(data=struct) for struct in tabular_structs]))
+        tabular_structs = []
+        tabular_metadata = [data.metadata for data in self.tabular_response]
+        for idx, tabular_data in enumerate(self.tabular_response):
```
stuqdog, line 204: I opted against a list comprehension here because I thought it was a bit cluttered and awkward to read. Not a strong preference; happy to revert to a list comprehension if others prefer it.

src/viam/app/data_client.py
```diff
@@ -60,6 +62,54 @@ class DataClient:
     `ViamClient`.
     """
 
+    class TabularData:
```
stuqdog, line 65: Returning the actual proto type for tabular data requests seemed like a bad idea because it's awkward to navigate and has a struct (which we wanted to get rid of. But, `time_requested` and `time_received` are their own fields, and I thought returning a 4ple would be super ugly. So, I created a class to represent meaningful tabular data, and created an analogous class for binary data for consistency's sake.

Happy to revisit the type question if there's pushback against creating these shadow classes.

tests/mocks/services.py
```diff
@@ -504,11 +507,20 @@ async def TabularDataByFilter(self, stream: Stream[TabularDataByFilterRequest, T
...
             return
         self.filter = request.data_request.filter
-        n = len(self.tabular_response)
-        tabular_response_structs = [Struct()] * n
-        for i in range(n):
-            tabular_response_structs[i].update(self.tabular_response[i])
-        await stream.send_message(TabularDataByFilterResponse(data=[TabularData(data=struct) for struct in tabular_response_structs]))
+        tabular_response_structs = []
+        tabular_metadata = [data.metadata for data in self.tabular_response]
+        for idx, tabular_data in enumerate(self.tabular_response):
```
stuqdog, line 512: see above comment re: use of list comprehension here.

tests/test_data_client.py
```diff
@@ -26,16 +30,20 @@
...
+MIME_TYPE = "mime_type"
+MIME_TYPES = [MIME_TYPE]
+URI = "some.robot.uri"
 SECONDS_START = 978310861
 NANOS_START = 0
 SECONDS_END = 978310861
 NANOS_END = 0
+START_TS = Timestamp(seconds=SECONDS_START, nanos=NANOS_START)
+END_TS = Timestamp(seconds=SECONDS_END, nanos=NANOS_END)
+START_DATETIME = START_TS.ToDatetime()
```
stuqdog, line 42: (flyby) I liked the idea of programmatically showing that `START_DATETIME` and `START_TS` are equivalent.

tests/test_data_client.py
```diff
@@ -26,16 +30,20 @@
...
+END_DATETIME = END_TS.ToDatetime()
 TAGS = ["tag"]
 BBOX_LABELS = ["bbox_label"]
-FILTER = Filter(
+FILTER = DataClient.create_filter(
```
stuqdog, line 46: (flyby) we both had a hardcoded filter here and - in one method below - a filter created via `create_filter`. this seemed redundant so I moved to a default of only using a single filter that is created via our helper method.

src/viam/app/data_client.py
```diff
@@ -60,6 +62,54 @@ class DataClient:
...
+        """Class representing a piece of tabular data and associated metadata.
+
+        Args:
+            data (Mapping[str, Any]): the requested data.
+            metadata (viam.proto.app.data.CaptureMetadata): the metadata from the request.
+            time_requested (datetime): the time the data request was sent.
+            time_received (datetime): the time the requested data was received.
+        """
+
+        def __init__(self, data: Mapping[str, Any], metadata: CaptureMetadata, time_requested: datetime, time_received: datetime) -> None:
```
8ashar, line 75: Do we need to annotate the init function's return type? If we do, should it be "Self" or "TabularData" instead of "None"?

docs/examples/_server.py
```diff
@@ -169,19 +172,48 @@
 class MockData(DataServiceBase):
     def __init__(self):
         self.tabular_data_requested = False
-        self.tabular_response = [{"PowerPct": 0, "IsPowered": False}, {"PowerPct": 0, "IsPowered": False}, {"Position": 0}]
+        self.tabular_response = [
+            DataClient.TabularData(
```
jckras, line 176: (q) In general, what is the "data" being collected? Any data the robot is giving? And what is binary data vs tabular data?

docs/examples/_server.py
```diff
@@ -169,19 +172,48 @@
...
+        for idx, tabular_data in enumerate(self.tabular_response):
+            tabular_structs.append(
+                TabularData(
+                    data=dict_to_struct(tabular_data.data),
+                    metadata_index=idx,
+                    time_requested=datetime_to_timestamp(tabular_data.time_requested),
+                    time_received=datetime_to_timestamp(tabular_data.time_received)
+                )
+            )
+        await stream.send_message(TabularDataByFilterResponse(
```
jckras, line 213: (q) Where is this message being sent?

jckras (approved):
Added questions for my own understanding but lgtm

src/viam/app/data_client.py
```diff
@@ -60,6 +62,54 @@ class DataClient:
...
+
+    class BinaryData:
+        """Class representing a piece of binary data and associated metadata.
+
+        Args:
+            data (bytes): the requested data.
+            metadata (viam.proto.app.data.BinaryMetadata): the metadata from the request.
+        """
+
+        def __init__(self, data: bytes, metadata: BinaryMetadata) -> None:
```
8ashar, line 100: Same comment here about init method type annotation

8ashar (commented):
Thank you for your help on this!

src/viam/app/data_client.py
stuqdog: Yeah, good question! I don't know that we have to annotate, we seem to be a bit inconsistent about it in the SDK. Personally I'm a fan of type annotations so I'd prefer to keep it in, and when we do annotate `init` methods in python it seems to pretty consistently set the return type as `None`.

src/viam/app/data_client.py
stuqdog: See above response :)

src/viam/app/data_client.py
```diff
@@ -60,6 +62,54 @@ class DataClient:
...
+        metadata: CaptureMetadata
+        time_requested: datetime
+        time_received: datetime
+
+        def __str__(self) -> str:
+            return f"{self.data}\n{self.metadata}Time requested: {self.time_requested}\nTime received: {self.time_received}\n"
+
+        def __eq__(self, other: "DataClient.TabularData") -> bool:
+            return str(self) == str(other)
+
```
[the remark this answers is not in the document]
stuqdog: Thanks for commenting on that! I don't think implementing the custom `str` or `eq` methods for the others is necessary at this time. I added the custom `str` here because we want to print out meaningful information when saving the data to file, and I added the `eq` method because we need it for testing. Because those concerns aren't relevant for the other shadow classes at this time, I don't think there's any need to do worry about adding these methods.

Re: adding a `proto` property, I'd love to do so but these aren't really shadow types in quite the same way. That is, there's no 1:1 mapping from the `BinaryData` or `TabularData` types that I implemented to a single proto-defined type (there are parameters we still don't care about from the proto type, the proto response actually contains two lists of different types and we're grabbing pieces of data from each, etc.). So I don't think a `proto` property really makes sense in this case.

src/viam/app/data_client.py
```diff
@@ -60,6 +62,54 @@ class DataClient:
...
+        data: bytes
+        metadata: BinaryMetadata
+
+        def __str__(self) -> str:
+            return f"{self.data}\n{self.metadata}"
+
+        def __eq__(self, other: "DataClient.BinaryData") -> bool:
+            return str(self) == str(other)
+
```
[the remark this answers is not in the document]
stuqdog: Yeah, good question! See above re: `proto` property and it not being a 1:1 conversion :)

docs/examples/_server.py
stuqdog: Good question! The message is being sent back along the channel where the request was sent, so basically back to the calling client.

docs/examples/_server.py
stuqdog: Great question! The data is currently just all data being collected/stored by the robot in app. I don't think I could give a very good answer on the differences between binary and tabular data. I think tabular is going to be mapped data in a lot of ways (e.g., a mapping of function call to output), while binary data is data that's just stored as bytes (e.g., raw files, or a photo). The data team could definitely give a better answer though!

8ashar (approved):
lgtm!

njooma (approved):

PR #381, opened 2023-08-10
[RSDK-3189] Implement and Test Robot API's

8ashar:
This PR adds the methods defined by proto [here](https://github.com/viamrobotics/api/blob/main/proto/viam/app/v1/robot.proto) to our AppClient. The PR also includes a testing suites and the implementation of a mock app server.


njooma:
@8ashar any movement here?

8ashar:
> 

wdym?

8ashar:
@njooma ^

njooma:
As in is this still needed? Are you still working on it? Has it gone stale??

stuqdog:
@njooma based on some discussions with fleet and soleng, we think this isn't really needed at this time, and there is potential for some wonkiness if we allow users to make these calls. I asked Bashar to keep it in draft state so that the work still exists in case that changes.

cheukt:
would like to close this - don't think we'll be needing this anytime soon and we can always fish it out of the closed  PRs section

stuqdog:
> would like to close this - don't think we'll be needing this anytime soon and we can always fish it out of the closed PRs section, is that ok @stuqdog?

sorry missed this somehow! yep totally fine, closing now.

PR #382, opened 2023-08-11
Rename sensor to imager

npentrel:
Am I misunderstanding this or are we using sensor as a synonym

### DIFF
--- a/src/viam/components/camera/camera.py
+++ b/src/viam/components/camera/camera.py
@@ -52,8 +52,8 @@ class Camera(ComponentBase):
 
     @abc.abstractmethod
     async def get_images(self, *, timeout: Optional[float] = None, **kwargs) -> Tuple[List[NamedImage], ResponseMetadata]:
-        """Get simultaneous images from different sensors, along with associated metadata.
-        This should not be used for getting a time series of images from the same sensor.
+        """Get simultaneous images from different cameras, along with associated metadata.
+        This should not be used for getting a time series of images from the same camera.
 
         Returns:
             Tuple[List[NamedImage], ResponseMetadata]:
--- a/src/viam/components/camera/camera.py
+++ b/src/viam/components/camera/camera.py
@@ -52,8 +52,8 @@ class Camera(ComponentBase):
 
     @abc.abstractmethod
     async def get_images(self, *, timeout: Optional[float] = None, **kwargs) -> Tuple[List[NamedImage], ResponseMetadata]:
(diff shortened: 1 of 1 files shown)


njooma (approved):
This is fine by me but this text was copied from RDK so we should change there as well cc @bazile-clyde

bazile-clyde:
@npentrel I didn't write this but I suspect the reason the word "camera" wasn't used is because this method was implemented for the [Intel Real Sense](https://www.intelrealsense.com/depth-camera-d435/). 

That is one camera with multiple "imagers" as they call it. Each takes a separate image and we use the timestamps to confirm both images were taken at roughly the same time. 

To answer your question, no, "sensor" was not used synonymously with "camera". However, I'm not opposed to changing the word here, though, "camera" might be a bit overloaded since now it's referring to the imagers and the camera housing the imagers, the camera component.

npentrel:
should we use imgers then?

cheukt (approved):

njooma:
@npentrel has there been any movement on this? I fear it becoming stale...

npentrel:
let me find the pr on the rdk and ask there - and then I'll follow up here

bazile-clyde:
> Ok - @bhaney added the RDK code - what do you think Bijan - should we change this to imagers?

@npentrel He's OOO for about two weeks, and my team owns this code. I'm okay with leaving it as is or changing it to "imagers."

npentrel:
ok let's do imagers then

src/viam/components/camera/camera.py
```diff
@@ -52,8 +52,8 @@ async def get_image(self, mime_type: str = "", *, timeout: Optional[float] = Non
 
     @abc.abstractmethod
     async def get_images(self, *, timeout: Optional[float] = None, **kwargs) -> Tuple[List[NamedImage], ResponseMetadata]:
-        """Get simultaneous images from different sensors, along with associated metadata.
-        This should not be used for getting a time series of images from the same sensor.
+        """Get simultaneous images from different cameras, along with associated metadata.
+        This should not be used for getting a time series of images from the same camera.
```
npentrel, line 56:
```suggestion
        """Get simultaneous images from different imagers, along with associated metadata.
        This should not be used for getting a time series of images from the same imager.
```

npentrel:
https://github.com/viamrobotics/rdk/pull/2833

njooma (approved):

bazile-clyde (approved):
LGTM!

PR #383, opened 2023-08-11
RSDK-4135 - implement getgeometries

purplenicole730:


src/viam/components/power_sensor/client.py
```diff
@@ -55,3 +57,11 @@ async def do_command(self, command: Mapping[str, ValueTypes], *, timeout: Option
         request = DoCommandRequest(name=self.name, command=dict_to_struct(command))
         response: DoCommandResponse = await self.client.DoCommand(request, timeout=timeout)
         return struct_to_dict(response.result)
+
+    async def get_geometries(self, *, extra: Optional[Dict[str, Any]] = None, timeout: Optional[float] = None) -> List[Geometry]:
```
purplenicole730, line 61: This was only added since it was failing tests because of sensor also having get_geometries. But this method doesn't exist in the proto file of PowerSensor, so I'm not entirely sure how this should go.

src/viam/components/arm/arm.py
```diff
@@ -1,10 +1,10 @@
...
+from typing import Any, Dict, Final, List, Optional, Tuple
 
 from viam.resource.types import RESOURCE_NAMESPACE_RDK, RESOURCE_TYPE_COMPONENT, Subtype
 
 from ..component_base import ComponentBase
-from . import JointPositions, KinematicsFileFormat, Pose
+from . import Geometry, JointPositions, KinematicsFileFormat, Pose
 
 
 class Arm(ComponentBase):
```
maximpertsov, line 10: Q: it seems like we are adding `get_geometries` to every component interface - should we add this to `ComponentBase` then?

src/viam/components/power_sensor/client.py
cheukt: can you implement it and raise an error in the PowerSensor parent class?

src/viam/components/input/input.py
```diff
@@ -192,3 +194,13 @@ async def trigger_event(self, event: Event, *, extra: Optional[Dict[str, Any]] =
             event (Event): The event to trigger
         """
         raise NotSupportedError(f"Input controller named {self.name} does not support triggering events")
+
+    @abc.abstractmethod
+    async def get_geometries(self) -> List[Geometry]:
+        """
+        Get all geometries associated with the input, in their current configuration, in the frame of the input.
```
cheukt, line 201: input -> input controller

src/viam/components/generic/generic.py
```diff
@@ -66,3 +68,13 @@ def complex_command(self, arg1, arg2, arg3):
     """
 
     SUBTYPE: Final = Subtype(RESOURCE_NAMESPACE_RDK, RESOURCE_TYPE_COMPONENT, "generic")
+
+    @abc.abstractmethod
+    async def get_geometries(self) -> List[Geometry]:
+        """
+        Get all geometries associated with the arm, in their current configuration, in the frame of the arm.
```
cheukt, line 75: arm -> generic

src/viam/components/base/client.py
```diff
@@ -130,3 +130,11 @@ async def do_command(
...
         response: DoCommandResponse = await self.client.DoCommand(request, timeout=timeout)
         return struct_to_dict(response.result)
+
+    async def get_geometries(self, *, extra: Optional[Dict[str, Any]] = None, timeout: Optional[float] = None) -> List[Geometry]:
+        if extra is None:
+            extra = {}
+        request = GetGeometriesRequest(name=self.name, extra=dict_to_struct(extra))
+        response: GetGeometriesResponse = await self.client.GetGeometries(request, timeout=timeout)
+        geometries = [geometry for geometry in response.geometries]
+        return geometries
```
cheukt, line 140: consider refactoring this out into a separate function

cheukt (changes requested):

maximpertsov (approved):

src/viam/components/base/client.py
cheukt: in this case I meant the entire four lines or so (like in https://github.com/viamrobotics/viam-typescript-sdk/blob/main/src/utils.ts#L36-L58)

src/viam/utils.py
```diff
@@ -160,8 +159,12 @@ def datetime_to_timestamp(dt: datetime) -> Timestamp:
     return timestamp
 
 
-def container_to_list(container: RepeatedCompositeFieldContainer) -> List[Any]:
-    return [item for item in container]
+async def get_geometries(client, extra: Optional[Dict[str, Any]] = None, timeout: Optional[float] = None) -> List[Any]:
```
cheukt, line 162: can you create a protocol for a client that can get geometries like Stoppable and Reconfigurable?

src/viam/components/power_sensor/power_sensor.py
```diff
@@ -90,3 +90,6 @@ def add_reading(name: str, reading, returntype: List) -> None:
         add_reading("power", pow, [float])
 
         return readings
+
+    async def get_geometries(self) -> List[Geometry]:
+        raise MethodNotImplementedError("GetGeometries").grpc_error
```
cheukt, line 95: thinking about this some more, a better move is to just not subclass Sensor - making it just unimplemented is a bit unwieldy, and there's no reason why it has to subclass Sensor anyway

cheukt (changes requested):

PR #387, opened 2023-08-16
RSDK-4578: Python SDK Base Properties

martha-johnston:
Update python SDK to include new base property `wheel_circumference_meters`

Note to reviewers: I am not sure if there's something I have to do to specifically update this PR to use the most recent API changes? I'm seeing some errors which I believe are because it's not referencing the most recent version of API. 

Edit: It passed all the checks though so maybe that was just a local issue

### DIFF
--- a/examples/server/v1/components.py
+++ b/examples/server/v1/components.py
@@ -184,7 +184,7 @@ class ExampleBase(Base):
         self.angular_pwr = Vector3(x=0, y=0, z=0)
         self.linear_vel = Vector3(x=0, y=0, z=0)
         self.angular_vel = Vector3(x=0, y=0, z=0)
-        self.props = Base.Properties(1.0, 1.0)
+        self.props = Base.Properties(1.0, 1.0, 1.0)
         super().__init__(name)
 
     async def move_straight(self, distance: int, velocity: float, extra: Optional[Dict[str, Any]] = None, **kwargs):
--- a/src/viam/components/base/base.py
+++ b/src/viam/components/base/base.py
@@ -23,6 +23,7 @@ class Base(ComponentBase):
     class Properties:
         width_meters: float
         turning_radius_meters: float
+        wheel_circumference_meters: float
 
     @abc.abstractmethod
     async def move_straight(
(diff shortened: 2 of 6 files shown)


stuqdog:
@martha-johnston you are correct, it looks like we don't have the latest in yet. I think tests are passing because we don't have any tests that specifically would check for this, but I wouldn't expect a user to have success with it as-is.

@njooma looks like we haven't updated protos in python in a while, are you aware of any changes that need to be implemented before we can do so?

maximpertsov:
> @martha-johnston you are correct, it looks like we don't have the latest in yet. I think tests are passing because we don't have any tests that specifically would check for this, but I wouldn't expect a user to have success with it as-is.
> 
> @njooma looks like we haven't updated protos in python in a while, are you aware of any changes that need to be implemented before we can do so?

I pulled in proto changes for typescript recently and saw that [the motion services had breaking changes](https://github.com/viamrobotics/viam-typescript-sdk/pull/155)

njooma:
> @martha-johnston you are correct, it looks like we don't have the latest in yet. I think tests are passing because we don't have any tests that specifically would check for this, but I wouldn't expect a user to have success with it as-is.
> 
> 
> 
> @njooma looks like we haven't updated protos in python in a while, are you aware of any changes that need to be implemented before we can do so?

Yup, there are breaking data changes as well. Tagged you @stuqdog and @8ashar to snag the latest data updates: [comment](https://github.com/viamrobotics/viam-python-sdk/pull/373#issuecomment-1669772251)

martha-johnston:
Just want to follow up here, is there something I specifically need to do related to these breaking changes in this PR in order to update to the latest proto?

njooma:
@martha-johnston I updated the protos. If you rebase from main, you should be able to get all the new protos and accurately test your code.

martha-johnston:
@maximpertsov @cheukt should be good to go for review now! all the protos are updated and tests pass

maximpertsov (approved):

cheukt (approved):
LGTM

PR #391, opened 2023-08-18
unnormalize api/model name

cheukt:
to be consistent with rdk


maximpertsov (commented):
would it be worth adding a small test for this?

maximpertsov (approved):

stuqdog (approved):

PR #400, opened 2023-08-22
RSDK-4455 - Run session heartbeat loop in separate thread

cheukt:
instead of within the same thread (using the event loop), since that can be crowded out if thread is blocked

only one shared var, and we only need it to track 3 states: True, False, None. Made an enum to track


src/viam/sessions_client.py
```diff
@@ -44,24 +41,24 @@ class SessionsClient:
...
+    def __init__(self, channel: Channel, address: str, dial_options: Optional[DialOptions], *, disabled: bool = False):
         self.channel = channel
         self.client = RobotServiceStub(channel)
         self._disabled = disabled
-        self._lock = asyncio.Lock(**loop_kwargs())
+
+        self._address = address
+        self._dial_options = dial_options
+
+        # can only synchronize numbers and chars across processes, so 0 = Unknown, 1 = True, 2 = False
```
maximpertsov, line 54: let's make these values consts or enums to codify this

tests/test_sessions_client.py
maximpertsov, line 1: can we add your test scenario as a unit test here? if not it should be a QA scenario

maximpertsov (approved):
[my test suggestion](https://github.com/viamrobotics/viam-python-sdk/pull/400#discussion_r1303232781) is [opt] if it ends up being too cumbersome

src/viam/sessions_client.py
cheukt: done

tests/test_sessions_client.py
cheukt: added a test, though looks a bit ugly haha

maximpertsov (approved):
nice!

cheukt:
after talking to @dgottlieb a bit more, will try to make this a thread

src/viam/sessions_client.py
```diff
@@ -103,66 +109,58 @@ async def metadata(self) -> _MetadataLike:
...
+            self._current_id = response.id
 
         # tick once to ensure heartbeats are supported
-        await _heartbeat_tick(self.client, self._current_id, self._supported)
+        await self._heartbeat_tick(self.client)
+
+        with self._lock:
+            if self._supported == _SupportedState.TRUE:
+                # We send heartbeats slightly faster than the interval window to
+                # ensure that we don't fall outside of it and expire the session.
```
maximpertsov, line 123: completely unrelated and my bad, but 1/5th the time is not "slightly" faster :clown_face: 

```suggestion
                # We send heartbeats faster than the interval window to
                # ensure that we don't fall outside of it and expire the session.
```

alternatively we can just remove the comment altogether

src/viam/sessions_client.py
```diff
@@ -103,66 +109,58 @@ async def metadata(self) -> _MetadataLike:
...
+                wait = self._heartbeat_interval.total_seconds() / 5
+
+                self._thread = Thread(
+                    name="heartbeat-thread",
+                    target=asyncio.run,
+                    args=(self._heartbeat_process(wait),),
+                    daemon=True,
+                )
+                self._thread.start()
```
maximpertsov, line 132: Q: is it possible to get here with an active thread? (e.g. `self._thread is not None`). If so, do we need to do any cleanup?

src/viam/sessions_client.py
```diff
@@ -44,24 +46,32 @@ class SessionsClient:
...
         self.client = RobotServiceStub(channel)
         self._disabled = disabled
-        self._lock = asyncio.Lock(**loop_kwargs())
+
+        self._address = address
+        self._dial_options = dial_options
+
+        self._lock: Lock = Lock()
+        self._supported: _SupportedState = _SupportedState.UNKNOWN
+        self._thread: Optional[Thread] = None
```
njooma, line 61: It'd be nice to define these above with the other instance vars

src/viam/sessions_client.py
```diff
@@ -79,77 +89,78 @@ async def _recv_trailers(self, event: RecvTrailingMetadata):
...
+            if self._supported == _SupportedState.TRUE:
+                # We send heartbeats slightly faster than the interval window to
+                # ensure that we don't fall outside of it and expire the session.
+                wait = self._heartbeat_interval.total_seconds() / 5
+
+                self._thread = Thread(
+                    name="heartbeat-thread",
+                    target=asyncio.run,
+                    args=(self._heartbeat_process(wait),),
+                    daemon=True,
```
njooma, line 130: Are we sure this should be daemonized?

src/viam/sessions_client.py
```diff
@@ -79,77 +89,78 @@ async def _recv_trailers(self, event: RecvTrailingMetadata):
...
         else:
             LOGGER.debug("Sent heartbeat successfully")
 
+    async def _heartbeat_process(self, wait: float):
+        dial_options = self._dial_options if self._dial_options is not None else DialOptions()
+        dial_options.disable_webrtc = True
+
+        channel = await dial(address=self._address, options=dial_options)
+        client = RobotServiceStub(channel.channel)
+        while self._supported == _SupportedState.TRUE:
```
njooma, line 157: Does this check need to be in a lock?

src/viam/sessions_client.py
cheukt: main issue I ran into a non-daemonized thread is that it doesn't stop on a ctrl+c, and since the daemonic thread just dies on program exit, the cleanup can be handled like normal (since only one rust connection is created).

src/viam/sessions_client.py
cheukt: I think it is possible, added a reset if so

src/viam/sessions_client.py
cheukt: moved, also moved some of the class vars into this function since I don't think they should've been class vars to begin with

maximpertsov (approved):

njooma (approved):

PR #416, opened 2023-08-31
DATA-1813 Add method to get fromDataManagement value from extra map in python SDK

kaywux:
Discussed here: https://github.com/viamrobotics/rdk/pull/2839#discussion_r1310713150


benjirewis (approved):
Thanks!

PR #418, opened 2023-09-01
RSDK-3961: Add max_reconnect_attempts

hexbabe:
https://viam.atlassian.net/browse/RSDK-3961

> Part of the webhook functionality with ESP32s involves the use of set_power_mode(PowerMode.POWER_MODE_OFFLINE_DEEP). 
> 
> When an ESP32 is in deep sleep, it is essentially off , and will therefore not be able to respond to the SDK. This means the SDK connection will time out on a success, resulting in an error.
> 
> While the user does have control over the sdk client’s retry interval, they don’t have control over how many times a re-connect is attempted. This means that the user can’t determine how long (retry_interval x num_retries) this unnecessary reconnection attempt may live, and therefore can’t control the upper limit on network or compute utilization.
> 
> 
> In dial options, add max_reconnect_attempts

I added the above attribute and fixed the infinite loop in `_check_connection`

### DIFF
--- a/src/viam/robot/client.py
+++ b/src/viam/robot/client.py
@@ -262,7 +262,7 @@ class RobotClient:
 
             # Failure to grab resources could be for spurious, non-networking reasons. Try three times just to be safe.
             connection_error = None
-            for attempt in range(3):
+            for _ in range(3):
                 try:
                     _: ResourceNamesResponse = await self._client.ResourceNames(ResourceNamesRequest(), timeout=1)
                     connection_error = None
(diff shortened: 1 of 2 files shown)


src/viam/robot/client.py
```diff
@@ -262,7 +262,7 @@ async def _check_connection(self, check_every: int, reconnect_every: int):
 
             # Failure to grab resources could be for spurious, non-networking reasons. Try three times just to be safe.
             connection_error = None
-            for attempt in range(3):
+            for _ in range(3):
```
hexbabe, line 265: I added this to mirror the loop structures since this `attempt` iterator is unused anyways

hexbabe:
Actually, how should we test this? Is it necessary to get a micro-RDK set up with a ESP32 and recreate Matt's conditions

njooma (approved):
Minor comment that can be quickly addressed and then this can be merged!

PR #422, opened 2023-09-05
RSDK-3941 - Example module updates

cheukt:
bash script now installs python-venv if it doesn't exist. Script also exits on certain failures. Logs are more parseable as well


examples/complex_module/run.sh
```diff
@@ -5,9 +5,37 @@ cd `dirname $0`
...
+if ! python3 -m venv $VENV_NAME >/dev/null 2>&1; then
+    echo "Failed to create virtualenv."
+    if command -v apt-get >/dev/null; then
+        echo "Detected Debian/Ubuntu, attempting to install python3-venv automatically."
+        SUDO="sudo"
+        if ! command -v $SUDO >/dev/null; then
+            SUDO=""
+        fi
+		if ! apt info python3-venv >/dev/null 2>&1; then
+			echo "package info not found, trying apt update"
```
benjirewis, line 17:
```suggestion
			echo "Package info not found, trying apt update"
```
[super-nit]

examples/simple_module/src/main.py
```diff
@@ -36,7 +36,7 @@ def validate_config(cls, config: ComponentConfig) -> Sequence[str]:
                 raise Exception("Multiplier cannot be 0.")
         else:
             cls.multiplier = 1.0
-        return [""]
+        return []
```
benjirewis, line 39: Good catch; was this causing implicit deps bugs?

benjirewis (approved):

examples/complex_module/run.sh
```diff
@@ -5,9 +5,37 @@ cd `dirname $0`
...
+			$SUDO apt -qq update >/dev/null
+		fi
+        $SUDO apt install -qqy python3-venv >/dev/null 2>&1
+        if ! python3 -m venv $VENV_NAME >/dev/null 2>&1; then
+            echo "This module requires Python >=3.8 and pip to be installed." >&2
```
abe-winter, line 22: why 3.8?

examples/simple_module/run.sh
```diff
@@ -5,9 +5,37 @@ cd `dirname $0`
...
+			echo "package info not found, trying apt update"
+			$SUDO apt -qq update >/dev/null
+		fi
+        $SUDO apt install -qqy python3-venv >/dev/null 2>&1
+        if ! python3 -m venv $VENV_NAME >/dev/null 2>&1; then
+            echo "Failed to start module. This module requires Python >=3.8 and pip to be installed." >&2
+            exit 1
+        fi
+    else
+        echo "Failed to start module. This module requires Python >=3.8 and pip to be installed." >&2
```
abe-winter, line 26: Re this error message -- could also get here if python is present but virtualenv is not.

(But not a merge blocker. This script may not need to be comprehensive; guessing the setup logic will live in its own tool)

abe-winter (approved):

examples/simple_module/src/main.py
cheukt: yep

examples/complex_module/run.sh
cheukt: 3.8 is the lowest Python we support

examples/simple_module/run.sh
cheukt: updated the error message

njooma (approved):

PR #419, opened 2023-09-01
APP-2517: org API key support

ohEmily:
[APP-2517](https://viam.atlassian.net/browse/APP-2517)

Requires changes to rust-utils, [RSDK-4755](https://github.com/viamrobotics/rust-utils/pull/75)

This script (in this case, with the little hacks in place to connect to a robot on localhost):
```python
import asyncio
from viam import logging
from viam.robot.client import RobotClient
from viam.rpc.dial import Credentials, DialOptions

async def connect_robot_location_secret():
    creds = Credentials(
        type='robot-location-secret',
        payload='qymq89tpc8m5fe7o90wf3kxn3tb55eaqj1uchx0ul89puggu')
    opts = RobotClient.Options(
        refresh_interval=0,
        dial_options=DialOptions(credentials=creds)
    )
    return await RobotClient.at_address('robot-main.gmmaqnzytw.viam.cloud', opts)

async def connect_api_key():
    creds = Credentials(
        type='api-key',
        payload='t5pp84x7yl05uldjhllxzbuwjt0p6m49')
    opts = RobotClient.Options(
        refresh_interval=0,
        log_level=logging.DEBUG,
        dial_options=DialOptions(credentials=creds, auth_entity="4bccae38-c344-4183-a454-321b9566c721")
    )
    return await RobotClient.at_address('robot-main.gmmaqnzytw.viam.cloud', opts)

async def main():
    robot = await connect_robot_location_secret()
    print('Resources (robot location secret):')
    print(robot.resource_names)

    robot = await connect_api_key()
    print('Resources (API key):')
    print(robot.resource_names)

    # Don't forget to close the robot when you're done!
    await robot.close()

if __name__ == '__main__':
    asyncio.run(main())
```

Will output this:
```bash
⇒  python localhostRobotHost.py
2023-09-01 19:22:34,768		INFO	viam.rpc.dial (dial.py:245)	Connecting to socket: /tmp/proxy-YHLqeW0A.sock
Resources (robot location secret):
[namespace: "rdk"
type: "service"
subtype: "data_manager"
name: "builtin"
, namespace: "rdk"
type: "service"
subtype: "sensors"
name: "builtin"
, namespace: "rdk"
type: "service"
subtype: "motion"
name: "builtin"
, namespace: "rdk"
type: "component"
subtype: "servo"
name: "fakeservo"
]
2023-09-01 19:22:34,808		DEBUG	viam.rpc.dial (dial.py:188)	Creating new viam-rust-utils runtime
[...]

PR #424, opened 2023-09-06
RSDK-4528 - Allow API keys for cloud apis + allow users to specify app url

cheukt:
[RSDK-4528](https://viam.atlassian.net/browse/RSDK-4528)

location_id can apparently be none in app_client, so no changes needed there
allowing users to specify app url will make it easier for testing and stuff

[RSDK-4528]: https://viam.atlassian.net/browse/RSDK-4528?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/viam/app/viam_client.py
+++ b/src/viam/app/viam_client.py
@@ -27,7 +27,7 @@ class ViamClient:
 
             dial_options (viam.rpc.dial.DialOptions): Required information for authorization and connection to app. `creds` and
                 `auth_entity` fields are required.
-            app_url: (Optional[str]): URL of app. Uses https://app.viam.com if not specified.
+            app_url: (Optional[str]): URL of app. Uses app.viam.com if not specified.
 
         Raises:
             ValueError: If the input parameters are missing a required field or simply invalid.
@@ -47,7 +47,7 @@ class ViamClient:
         if dial_options.credentials.type == "robot-location-secret":
             self._location_id = dial_options.auth_entity.split(".")[1]
         if app_url is None:
-            app_url = "https://app.viam.com"
+            app_url = "app.viam.com"
         self._channel = await _dial_app(app_url)
         access_token = await _get_access_token(self._channel, dial_options.auth_entity, dial_options)
         self._metadata = {"authorization": f"Bearer {access_token}"}
--- a/src/viam/app/viam_client.py
+++ b/src/viam/app/viam_client.py
@@ -1,4 +1,4 @@
-from typing import Mapping
+from typing import Mapping, Optional
 
 from grpclib.client import Channel
 from typing_extensions import Self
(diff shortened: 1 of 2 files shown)


src/viam/app/viam_client.py
```diff
@@ -42,16 +43,20 @@ async def create_from_dial_options(cls, dial_options: DialOptions) -> Self:
             raise ValueError("dial_options.auth_entity cannot be None.")
 
         self = cls()
-        self._location_id = dial_options.auth_entity.split(".")[1]
-        self._channel = await _dial_app(dial_options)
+        self._location_id = None
+        if dial_options.credentials.type == "robot-location-secret":
+            self._location_id = dial_options.auth_entity.split(".")[1]
+        if app_url is None:
+            app_url = "https://app.viam.com"
```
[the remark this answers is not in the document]
cheukt: not particularly, but figured specifying https:// is slightly better than specifying the port number

src/viam/rpc/dial.py
```diff
@@ -30,7 +30,7 @@ class Credentials:
     Currently only supports robot location secret.
```
ohEmily, line 30: [minor] looks like we can remove this line. Yay!

src/viam/rpc/dial.py
```diff
@@ -30,7 +30,7 @@ class Credentials:
...
     """
 
-    type: Union[Literal["robot-location-secret"], Literal["robot-secret"]]
+    type: Union[Literal["robot-location-secret"], Literal["robot-secret"], Literal["api-key"]]
```
ohEmily, line 33: [q] Should we add validation on this field such that users would get an error if they supplied a bad literal/typo? I'm surprised I was able to supply the `api-key` string in the original org API key [PR](https://github.com/viamrobotics/viam-python-sdk/pull/419) without hitting any errors.

src/viam/rpc/dial.py
```diff
@@ -314,5 +314,5 @@ async def dial_direct(address: str, options: Optional[DialOptions] = None) -> Ch
     return await _dial_direct(address, options)
 
 
-async def _dial_app(options: DialOptions) -> Channel:
-    return await _dial_direct("app.viam.com:443")
+async def _dial_app(app_url: str) -> Channel:
+    return await _dial_direct(app_url)
```
ohEmily, line 318: [q] I see that above in this file we do:
```
    host, port = _host_port_from_url(address)
    if not port:
        port = 80 if insecure else 443
    server_hostname = host
```
Should we maybe have it default to `8080`? Were you able to test this both with `localhost:8080` as well as `http://localhost`? I am not sure if it would work for the latter option as-is but haven't tested. We could potentially match on the string `localhost` (rather than requiring the `http://` prefix) and hardcode 8080 in that case.

src/viam/rpc/dial.py
cheukt: it's just a type hint right now, so no error - the idea was that users could supply other cred types as the server changes without changes in the SDK, which I think is nicer - the server can/will reject if the type is incorrect anyway. I do think we could perhaps prepopulate these types, but out of scope for now

src/viam/rpc/dial.py
cheukt: `localhost:8080` definitely works, so does `http://localhost` if I update the default to be 8080. I think we kept it as `80` because that's default insecure port

src/viam/rpc/dial.py
cheukt: updated

src/viam/app/viam_client.py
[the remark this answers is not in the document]
cheukt: tho neither is necessary, I could make it just `app.viam.com` if that's better

src/viam/rpc/dial.py
ohEmily: ok! I'm not sure I follow why keeping it as `80` as desirable but I'm sure I'm lacking a lot of context here. Going to hit approve. Thank you!

ohEmily (approved):

PR #429, opened 2023-09-13
RSDK-4895: Make numpy an extra dependency

hexbabe:
https://viam.atlassian.net/browse/RSDK-4895

Changed pyproject.toml to not require numpy, adding it as an extra/dev dependency only; changed np imports to try/except and warn users on import of the mlmodel module

Tested locally with .whl file and [mlmodel] flag in venv


src/viam/services/mlmodel/client.py
```diff
@@ -5,8 +5,28 @@
...
+    from numpy.typing import NDArray
+except ImportError:
+    import warnings
+
+    warnings.warn(
+        (
+            "MLModel support in Viam Python SDK requires the installation of optional "
+            "dependencies: specifically, numpy.  Update your package using the extra [mlmodel]"
+            "e.g. `pip install viam-sdk[mlmode]` or the equivalent update in your dependency manager"
+        )  # will want to replace the e.g. suggestion with a link to docs when that is available
```
hexbabe, line 26: thoughts on the content of these comments?

pyproject.toml
```diff
@@ -17,7 +17,7 @@ include = ["LICENSE", "src/viam/rpc/libviam_rust_utils.*"]
   typing-extensions = "^4.7.1"
   Pillow = "^10.0.0"
   protobuf = "^4.23.4"
-  numpy = ">=1.21"
+  numpy = { version = ">=1.21", optional = true }
```
hexbabe, line 20: re-included this bit that was moved to dev dependencies because I wanted to follow the docs to make sure the extras flag works with PEP508/pip install stuff

source: https://python-poetry.org/docs/pyproject/#extras

src/viam/services/mlmodel/client.py
hexbabe: @andf-viam will likely be making updates to said docs

src/viam/utils.py
```diff
@@ -1,12 +1,25 @@
...
 import functools
-import numpy as np
+
+try:
+    import numpy as np
+    from numpy.typing import NDArray
+except ImportError:
+    import warnings
+
+    warnings.warn(
```
cheukt, line 11: this warning will print on import of the utils package, which will likely not involve the MLModel at all - I would consider moving the specific numpy utils out to a different file

hexbabe: yep true 😱

tests/test_mlmodel.py
```diff
@@ -64,6 +64,9 @@ def setup_class(cls):
         cls.manager = ResourceManager([cls.mlmodel])
         cls.service = MLModelRPCService(cls.manager)
 
+    #  ignore warning about our out-of-bound int casting (i.e. uint32 -> int16) because we don't store uint32s for int16 & uint16 tensor
```
hexbabe, line 67: numpy was giving 16 warnings re: these tests because we cast uint32 to uint16 & int16

but this shouldn't be a problem as per comment

pyproject.toml
```diff
@@ -60,3 +61,6 @@ line_length = 140
 [build-system]
 requires = [ "poetry-core>=1.0.0" ]
 build-backend = "poetry.core.masonry.api"
+
+[tool.poetry.extras]
+mlmodel = ["numpy"]
```
cheukt, line 66: I know we agreed on mlmodel earlier, but given we can't hide the directory anyway, I think numpy would be a better name given it'll be more generic. @njooma thoughts?

src/viam/services/mlmodel/utils.py
```diff
@@ -0,0 +1,103 @@
...
+    import numpy as np
+    from numpy.typing import NDArray
+except ImportError:
+    import warnings
+
+    warnings.warn(
+        (
+            "MLModel support in Viam Python SDK requires the installation of optional "
+            "dependencies: specifically, numpy.  Update your package using the extra [mlmodel]"
+            "e.g. `pip install viam-sdk[mlmodel]` or the equivalent update in your dependency manager"
```
cheukt, line 11: would be good to factor this string out so we aren't repeating it a few times

tests/test_utils.py
```diff
@@ -214,6 +213,9 @@ def test_struct_to_dict():
     assert struct_to_dict(struct) == expected
 
 
+#  ignore warning about our out-of-bound int casting (i.e. uint32 -> int16) because we don't store uint32s for int16 & uint16 tensor data
+# > 2^16-1 in the first place (inherently they are int16, we just cast them to uint32 for the grpc message)
+@pytest.mark.filterwarnings("ignore::DeprecationWarning")
 def test_flat_tensors_to_ndarrays():
```
cheukt, line 219: move these tests to a new file - test_mlmodel_utils.py

cheukt (changes requested):

src/viam/utils.py
```diff
@@ -351,3 +265,8 @@ def from_dm_from_extra(extra: Optional[Dict[str, Any]]) -> bool:
         return False
 
     return bool(extra["fromDataManagement"])
+
+
+_numpy_import_error_string = """MLModel support in Viam Python SDK requires the installation of optional
+dependencies: specifically, numpy.  Update your package using the extra [mlmodel]
+e.g. `pip install viam-sdk[mlmodel]` or the equivalent update in your dependency manager"""
```
andf-viam, line 272: If we require `numpy` for all MLModel usage in our Py SDK, then we must use `required` instead of `optional`. What do you think of:

```suggestion
_numpy_import_error_string = """MLModel support in the Viam Python SDK requires the installation of an
additional dependency: numpy.  Update your package using the extra [mlmodel]
e.g. `pip install viam-sdk[mlmodel]` or the equivalent update in your dependency manager."""
```

pyproject.toml
```diff
@@ -37,6 +37,7 @@ include = ["LICENSE", "src/viam/rpc/libviam_rust_utils.*"]
   tox = "^4.6.4"
   isort = "^5.12.0"
   pytest-watcher = "^0.3.4"
+  numpy = ">=1.21"
```
njooma, line 40: This can be removed since it's now included in the main dependency section

pyproject.toml
njooma: I think it should still be `mlmodel` because it should be the name of the enabled feature, not its dependencies. No we can't hide it, but we do not allow it to be imported without its dependencies. Plus, this gives us the option to add more dependencies to that feature

src/viam/utils.py
```diff
@@ -351,3 +265,8 @@ def from_dm_from_extra(extra: Optional[Dict[str, Any]]) -> bool:
         return False
 
     return bool(extra["fromDataManagement"])
+
+
+_numpy_import_error_string = """MLModel support in Viam Python SDK requires the installation of optional
```
njooma, line 270: This would probably live in the `services/mlmodel/__init__.py` file, and maybe with the name `_requirements_not_met_error_string` or something similar (basically just not specifying `numpy` by name, in case of future dependency changes)

njooma (changes requested):

pyproject.toml
cheukt: aight, my feelings aren't too strong, so this is fine for now

src/viam/utils.py
hexbabe: We refactor the string elsewhere but I replaced with your suggestions there

cheukt (approved):
LGTM

njooma (approved):

PR #436, opened 2023-09-18
Fix simple module

cheukt:
Validation shouldn't be setting variables on the class, and we generally shouldn't have instance-specific vars as class vars either.
Ran into this while investigating RSDK-4171


docs/examples/module_step2.py
```diff
@@ -26,9 +26,6 @@ async def get_readings(self, extra: Optional[Dict[str, Any]] = None, **kwargs) -
         wifi_signal = [x for x in content[2].split(" ") if x != ""]
         return {"link": wifi_signal[2], "level": wifi_signal[3], "noise": wifi_signal[4]}
 
-    async def get_geometries(self, *, extra: Optional[Dict[str, Any]] = None, timeout: Optional[float] = None) -> List[Geometry]:
```
purplenicole730, line 29: Does every `Sensor` need to define `get_geometries()` or is it unnecessary?

purplenicole730, line 29: Does every `Sensor` need to define `get_geometries()` or is it unnecessary?

purplenicole730 (approved):
LGTM, although I did have questions about `get_geometries()`. Is it that not every `Sensor` needs it defined, or that the simple module was missing it?

docs/examples/module_step2.py
cheukt: it's unnecessary because the `ComponentBase` parent class already implements it as unimplemented

PR #437, opened 2023-09-20
Bugfix: Check connection status before attempting to reconnect

maximpertsov:


maximpertsov:
Will add to this change as a patch to our last release branch

PR #441, opened 2023-09-21
DATA-1813 Fix KeyError when fromDataManagement is not in extra

kaywux:
Bug fix from https://github.com/viamrobotics/viam-python-sdk/pull/416

### DIFF
--- a/src/viam/utils.py
+++ b/src/viam/utils.py
@@ -258,8 +258,8 @@ async def to_thread(func: Callable[_P, _R], *args: _P.args, **kwargs: _P.kwargs)
 
 
 def from_dm_from_extra(extra: Optional[Dict[str, Any]]) -> bool:
-    """Used in modular filter components to get the 'fromDataManagement' value from an extra map."""
+    """Used in modular filter components to return whether 'fromDataManagement' was set in the extra map."""
     if extra is None:
         return False
 
-    return bool(extra["fromDataManagement"])
+    return "fromDataManagement" in extra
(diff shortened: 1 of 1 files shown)


src/viam/utils.py
```diff
@@ -258,8 +258,8 @@ async def to_thread(func: Callable[_P, _R], *args: _P.args, **kwargs: _P.kwargs)
 
 
 def from_dm_from_extra(extra: Optional[Dict[str, Any]]) -> bool:
-    """Used in modular filter components to get the 'fromDataManagement' value from an extra map."""
+    """Used in modular filter components to return whether 'fromDataManagement' was set in the extra map."""
     if extra is None:
         return False
 
-    return bool(extra["fromDataManagement"])
+    return "fromDataManagement" in extra
```
cheukt, line 265: isn't it possible that "fromDataManagement" is in extra but not true?

kaywux: The only time we pass "fromDM" is when it's explicitly set to true - but changed this to return the value.

benjirewis (approved):

src/viam/utils.py
```diff
@@ -259,7 +259,7 @@ async def to_thread(func: Callable[_P, _R], *args: _P.args, **kwargs: _P.kwargs)
 
 def from_dm_from_extra(extra: Optional[Dict[str, Any]]) -> bool:
     """Used in modular filter components to get the 'fromDataManagement' value from an extra map."""
-    if extra is None:
+    if extra is None or "fromDataManagement" not in extra:
         return False
 
     return bool(extra["fromDataManagement"])
```
cheukt, line 265:
```suggestion
    return bool(extra.get("fromDataManagement", False))
```
feels a little more natural to me, but either is fine

cheukt (approved):

src/viam/utils.py
cheukt: that sounds good, returning value is a bit more robust so thanks for making the change!

PR #443, opened 2023-09-22
RSDK-4978 - return file id from upload

cheukt:
also some flyby formatting

### DIFF
--- a/tests/test_data_sync_client.py
+++ b/tests/test_data_sync_client.py
@@ -42,7 +42,7 @@ class TestClient:
     async def test_binary_data_capture_upload(self, service: MockDataSync):
         async with ChannelFor([service]) as channel:
             client = DataClient(channel, DATA_SERVICE_METADATA)
-            response = await client.binary_data_capture_upload(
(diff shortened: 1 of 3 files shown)


tests/test_data_sync_client.py
```diff
@@ -45,35 +42,37 @@ class TestClient:
     async def test_binary_data_capture_upload(self, service: MockDataSync):
         async with ChannelFor([service]) as channel:
             client = DataClient(channel, DATA_SERVICE_METADATA)
-            await client.binary_data_capture_upload(
+            response = await client.binary_data_capture_upload(
```
maximpertsov, line 45: [minor] to match the naming convention in the existing tests below

```suggestion
            file_id = await client.binary_data_capture_upload(
```

maximpertsov (approved):
minor suggestion for clarity/consistency, otherwise LGTM

tests/test_data_sync_client.py
cheukt: updated for both tests

benjirewis (approved):

PR #446, opened 2023-09-27
RSDK-5174: Fix get_image mimetype handler in python sdk

hexbabe:
https://viam.atlassian.net/browse/RSDK-5174

The first conditional branch of `get_image_from_response` never gets hit because the default `mime_type` value in the invoker— `get_image` in client.py —  is `""` and not `None`.

### DIFF
--- a/src/viam/components/camera/client.py
+++ b/src/viam/components/camera/client.py
@@ -23,7 +23,7 @@ from viam.utils import ValueTypes, dict_to_struct, get_geometries, struct_to_dic
 from . import Camera, RawImage
 
 
-def get_image_from_response(data: bytes, response_mime_type: str, request_mime_type: Optional[str] = None) -> Union[Image.Image, RawImage]:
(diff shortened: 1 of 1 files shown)


cheukt (commented):
is there a corresponding test we can add?

hexbabe:
> is there a corresponding test we can add?

I think a test exists for this use case i.e. empty mime type, but we don't invoke the helper function, `get_image_from_response`, that actually handles the logic. We do it manually like this:

```
            # Test empty mime type. Empty mime type should default to JPEG for non-depth cameras
            request = GetImageRequest(name="camera")
            response: GetImageResponse = await client.GetImage(request)
            assert service._camera_mime_types["camera"] == CameraMimeType.JPEG
```

cheukt (approved):

hexbabe:
I also tested with an actual Python sdk client and it works now no args

maximpertsov (approved):

seanavery (approved):
LGTM!

PR #447, opened 2023-09-27
Fix reconnect logic

cheukt:
- sessions client was overwriting dial options in the robot client, so making an actual copy
- on reconnect, sessions client actually wants the new socket address, so got that instead of the address of the robot itself

### DIFF
--- a/src/viam/sessions_client.py
+++ b/src/viam/sessions_client.py
@@ -1,4 +1,5 @@
 import asyncio
+from copy import deepcopy
 from datetime import timedelta
 from enum import IntEnum
 from threading import Lock, Thread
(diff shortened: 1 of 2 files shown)


src/viam/sessions_client.py
```diff
@@ -44,12 +44,23 @@ class SessionsClient:
...
         self._disabled = disabled
+        if dial_options is not None:
+            self._dial_options = DialOptions(
+                disable_webrtc=True,
+                auth_entity=dial_options.auth_entity,
+                credentials=dial_options.credentials,
+                insecure=dial_options.insecure,
+                allow_insecure_downgrade=dial_options.allow_insecure_downgrade,
+                allow_insecure_with_creds_downgrade=dial_options.allow_insecure_with_creds_downgrade,
+            )
```
maximpertsov, line 61: Let's use deepcopy here (`from copy import deepcopy`), otherwise we'll need to remember to add new dial option fields here too.

```suggestion
        if dial_options is not None:
            self._dial_options = deepcopy(dial_options)
```

cheukt: done

src/viam/sessions_client.py
```diff
@@ -50,18 +51,8 @@ def __init__(self, channel: Channel, direct_dial_address: str, dial_options: Opt
...
-                credentials=dial_options.credentials,
-                insecure=dial_options.insecure,
-                allow_insecure_downgrade=dial_options.allow_insecure_downgrade,
-                allow_insecure_with_creds_downgrade=dial_options.allow_insecure_with_creds_downgrade,
-            )
-        else:
-            self._dial_options = DialOptions(disable_webrtc=True)
-
+        self._dial_options = deepcopy(dial_options) if dial_options is not None else DialOptions()
+        self._dial_options.disable_webrtc = True
```
maximpertsov, line 55: ah nice, yeah i forgot we always want webrtc disabled here - good catch

maximpertsov (approved):

PR #454, opened 2023-10-03
Set dependencies to the most minimum version

njooma:
By setting dependencies to their most minimum versions (but allowing for newer versions), we enable users with older versions of these packages to continue to use the SDK.

For example, @IanWhalen ran into an issue where the SDK required Pillow >=10, but another dependency of his required <10. Since the SDK did not use any new functionality from version Pillow v10, we should've kept the requirement at >=9.

This resets the minimum Pillow requirement to be >=9. It also sets the current deps to all be `>= VERSION`, so in the future we should manually update these numbers IFF we use new functionality introduced by the new version.


maximpertsov:
I noticed that CI still installs Pillow version 10:

```
# install packages step
  • Installing pillow (10.0.1)
```

Is there a way to force CI to install the minimum version to catch if accidentally use a feature that isn't available in the minimum version?

njooma:
@maximpertsov Good call out. I've added a new test suite to the matrix so that we now test against the minimum and maximum requirements: https://github.com/viamrobotics/viam-python-sdk/actions/runs/6422987966/job/17440578256

In future, we might want to look at what [Sentry have done](https://sentry.engineering/blog/how-we-run-our-python-tests-in-hundreds-of-environments-really-fast) if we want something more robust, but it seems like overkill right now.

maximpertsov (approved):

PR #455, opened 2023-10-04
RSDK-4977 - add ml training apis

stuqdog:
Adds ML Training client.

(Note to reviewers) I am currently running tests to make sure these all work in practice. I expect things to work out and didn't want to delay getting reviewers on it, but it's possible some small fixes will be pushed based on those tests.

### DIFF
--- a/src/viam/app/ml_training_client.py
+++ b/src/viam/app/ml_training_client.py
@@ -16,9 +16,7 @@ from viam.proto.app.mltraining import (
     TrainingStatus,
     TrainingJobMetadata,
 )
-from viam.proto.app.data import (
-    Filter,
-)
+from viam.proto.app.data import Filter
 
 LOGGER = logging.getLogger(__name__)
 
--- a/src/viam/app/ml_training_client.py
+++ b/src/viam/app/ml_training_client.py
@@ -11,8 +11,6 @@ from viam.proto.app.mltraining import (
     ListTrainingJobsResponse,
     MLTrainingServiceStub,
     ModelType,
-    SubmitTrainingJobRequest,
-    SubmitTrainingJobResponse,
     TrainingStatus,
     TrainingJobMetadata,
 )
--- a/src/viam/app/ml_training_client.py
+++ b/src/viam/app/ml_training_client.py
@@ -48,32 +48,7 @@ class MLTrainingClient:
         tags: List[str],
         filter: Optional[Filter] = None,
     ) -> str:
-        """Submits a training job to be processed by the training job manager.
-
-        Args:
-            org_id (str): the ID of the organization whose data will be used for training.
-            model_name (str): the training model name.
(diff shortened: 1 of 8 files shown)


src/viam/app/ml_training_client.py
```diff
@@ -0,0 +1,121 @@
...
+            model_version=model_version,
+            model_type=model_type,
+            tags=tags,
+            filter=filter,
+        )
+        response: SubmitTrainingJobResponse = await self._ml_training_client.SubmitTrainingJob(request, metadata=self._metadata)
+
+        return response.id
+
+    async def get_training_job(self, id: str) -> TrainingJobMetadata:
```
stuqdog, line 80: (note to reviewers) I think there's an argument to be made for creating our own shadow type for `TrainingJobMetadata`. The type ultimately breaks down into a struct of primitives, a request (which is just a a string and a `filter`, which we already have conversion functions for), and an enum (which is relatively straightforward). So, the type isn't terribly complicated, but it is a bit nested (i.e., it contains a request which contains a filter) and I could imagine users not loving interacting with it.

I opted for the simpler thing for now because 1) why do extra work if it's not necessary, and 2) my understanding is @njooma has a preference for using the `proto` types when we can reasonably get away with it. But, I'm curious to get y'all's thoughts. You can see what a manually constructed `TrainingJobMetadata` looks like in the test file below.

cheukt:
cc @mcvella

src/viam/app/ml_training_client.py
njooma: I can see how hoisting some of the fields to top-level would be useful. If we were to make a shadow class, would filter also be automatically converted to the python-native type?

src/viam/utils.py
```diff
@@ -263,3 +268,65 @@ def from_dm_from_extra(extra: Optional[Dict[str, Any]]) -> bool:
         return False
 
     return bool(extra.get("fromDataManagement", False))
+
+
+def create_filter(
```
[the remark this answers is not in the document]
njooma: I think it's fine here. If you really wanted, you could create a single-use mixin, but [beware the pitfalls](https://www.ianlewis.org/en/mixins-and-python)

njooma (approved):

src/viam/app/ml_training_client.py
stuqdog: I don't think we actually have a python-native `Filter` type, we just have a helper function for creating a `Filter`. So users would still be stuck with the proto type unless we want to now create a python-native `Filter`. This would be a bit of a reversal from our previous approach, but maybe worth considering given the now-expanded presence of `Filter`s?

njooma: In that case I think we can leave this as is -- no need to create a shadow type

cheukt (approved):
LGTM

PR #458, opened 2023-10-09
RSDK-5109 - add close function

purplenicole730:
[Jira ticket](https://viam.atlassian.net/browse/RSDK-5109)

Add a `close()` function to the Python SDK so that users can specify what happens when a resource closes. This function is not meant to be called directly, but rather to be called when a module or resource manager is removing a resource.

Changes
- Added a default close function in `ResourceBase` that returns nothing, but can be overwritten
- Modules also now call `close()` when removing a resource
- Unittests added to assert that both default (summation) and overwritten (gizmo) `close()` functions were called appropriately
- Resource manager now has a `close()` function, where it calls all its resources' `close()` functions
- Unittests added to assert that resources' closed functions were called when resource manager's close function called
- Added close functions to most of the examples with a comment explaining that it was completely optional

Closing the resource manager was also tested manually by printing out statements in the example's `server.py`


src/viam/resource/manager.py
```diff
@@ -105,3 +105,7 @@ def _resource_by_name_only(self, name: str) -> ResourceBase:
             if rname.name == name:
                 return resource
         raise ResourceNotFoundError("resource", name)
+
+    async def close(self):
+        for _, resource in self.resources.items():
+            await resource.close()
```
cheukt, line 111: what if a resource errors while closing?

src/viam/module/module.py
```diff
@@ -161,6 +161,7 @@ async def remove_resource(self, request: RemoveResourceRequest):
                 await resource.stop()
             else:
                 resource.stop()
+        await resource.close()
```
cheukt, line 164: what if a resource errors while closing?

cheukt (changes requested):
looks pretty good! can you update the docs/example modules to refer to this new functionality as well? In the example modules, I think just logging something will be good, just want to surface the function within the examples

src/viam/module/module.py
cheukt: also could we just put the close call inside server.remove_resource()?

njooma: Perhaps it should go in `ResourceManager.remove_resource` since we always want to make sure that a resource is closed if we're about to remove it. 

My suggestion would be to make `remove_resource` async, because it gives writers of modules a lot more flexibility in case they need to wait for resource cleanup

njooma (approved):
Looks good pending updates from @cheukt's comments

src/viam/module/module.py
purplenicole730: Note: Would making `remove_resource()` async mean a possible breaking change to anyone using modules?

src/viam/resource/manager.py
purplenicole730: Ah, I guess the move here is to raise an error, but continue with the close so that the robot isn't just sitting there not able to close.

src/viam/module/module.py
cheukt: would it? do users call `remove_resource()` directly or is it just internal?

purplenicole730: There aren't any examples of a user calling `remove_resource()`, but if users are able to call `add_resource()`... I guess it'd be a very specific/weird scenario to call it directly. I'll make the change without worry then!

src/viam/resource/manager.py
njooma: Maybe you could collect the errors (or at least which resources did error) and then raise/log those?

pseudocode below

```python
for _, resource in self.resource.items():
    try:
        await resource.close()
    catch e:
        LOGGER.error(f'Encountered error while closing resource named {resource.name}', e)
```

or even

```python
errors = {}
for _, resource in self.resource.items():
    try:
        await resource.close()
    catch e:
        errors[resource.name] = e
    finally:
        LOGGER.error(f'Encountered {len(errors)} errors while closing: {errors}')
```

purplenicole730: Yep, I did something similar to the second one!

src/viam/resource/manager.py
```diff
@@ -91,21 +91,23 @@ def get_resource(self, of_type: Type[ResourceType], name: ResourceName) -> Resou
...
+    async def remove_resource(self, name: ResourceName):
         """Remove the resource with the specified ```ResourceName```.
 
         Args:
             name (viam.proto.common.ResourceName): The ResourceName of the resource
         """
         with self._lock:
-            del self.resources[name]
+            try:
+                resource = self._resource_by_name_only(name.name)
```
cheukt, line 102: I don't think this is the best way to index into a resource - what if we have an arm called `arm1` and a base called `arm1`?

src/viam/rpc/server.py
```diff
@@ -113,7 +113,11 @@ async def serve(
                 await self._server.start(host, port)
                 LOGGER.info(f"Serving on {host}:{port}")
             await self._server.wait_closed()
-            await self.close()
+            for rn, resource in self.resources.items():
+                try:
+                    await resource.close()
+                except Exception as e:
+                    LOGGER.error(f"Error while closing {rn.name}:", e)
```
cheukt, line 120: I would leave this logic in the resource manager

docs/examples/module_step2_optional.py
```diff
@@ -48,6 +48,10 @@ def reconfigure(self, config: ComponentConfig, dependencies: Mapping[ResourceNam
             multiplier = 1.0
         self.multiplier = multiplier
 
+    def close(self):
+        # This is a completely optional function to include. This will be called when the resource is closed.
+        print(f"{self.name} is closed.")
```
cheukt, line 53: we should use the logger instead - print statements don't get propagated to the RDK

docs/examples/module_step2_optional.py
```diff
@@ -48,6 +48,10 @@ def reconfigure(self, config: ComponentConfig, dependencies: Mapping[ResourceNam
             multiplier = 1.0
         self.multiplier = multiplier
 
+    def close(self):
+        # This is a completely optional function to include. This will be called when the resource is closed.
```
cheukt, line 52:
```suggestion
        # This is a completely optional function to include. This will be called when the resource is removed from the config or the module is shutting down.
```
think this is slightly clearer as to when it gets called?

cheukt (changes requested):
a few smaller nits

src/viam/resource/manager.py
```diff
@@ -99,13 +101,27 @@ async def remove_resource(self, name: ResourceName):
...
+        with self._lock:
+            errors = {}
+            for rn in rns:
+                try:
+                    await self.remove_resource(rn)
+                except Exception as e:
+                    errors[rn] = e
+                    LOGGER.error(f"Error while closing {rn.name}:", e)
+            if errors != {}:
+                raise Exception(errors)
```
cheukt, line 123: I think it's ok to not raise even with errors, would just note in the docstring.

cheukt (approved):
LGTM

src/viam/rpc/server.py
```diff
@@ -113,9 +113,10 @@ async def serve(
                 await self._server.start(host, port)
                 LOGGER.info(f"Serving on {host}:{port}")
             await self._server.wait_closed()
-            for rn, resource in self.resources.items():
+            rns = [key for key in self.resources.keys()]
```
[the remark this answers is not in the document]
purplenicole730: Yes, `self.resources.keys()` itself raises issues. And already addressed the second point!

src/viam/resource/manager.py
purplenicole730: Sounds good. Thank you!

PR #459, opened 2023-10-10
RSDK-4788: Add Readings to Python SDK

martha-johnston:
fix imports (from sensor to common) and use `GetReadingsRequest/Response` for movement sensor and power sensor


src/viam/components/power_sensor/power_sensor.py
```diff
@@ -57,6 +57,7 @@ async def get_readings(self, *, extra: Optional[Dict[str, Any]] = None, timeout:
                is_ac: bool
                power: float
             }
+
```
martha-johnston, line 60: should the below be removed? in `sensor.py` the only thing in `get_readings` is ...

src/viam/components/movement_sensor/movement_sensor.py
```diff
@@ -147,6 +147,7 @@ async def get_readings(self, *, extra: Optional[Dict[str, Any]] = None, timeout:
                 compass: float,
                 orientation: Orientation,
             }
+
```
martha-johnston, line 150: should the below be removed? in `sensor.py` the only thing in `get_readings` is `...`. Should we not be calling everything manually now that we have the request/response to use?

src/viam/components/movement_sensor/client.py
```diff
@@ -3,7 +3,7 @@
 from grpclib.client import Channel
 
 from viam.components.movement_sensor.movement_sensor import MovementSensor
-from viam.proto.common import DoCommandRequest, DoCommandResponse, Geometry
+from viam.proto.common import DoCommandRequest, DoCommandResponse, Geometry, GetReadingsRequest, GetReadingsResponse
```
benjirewis, line 6: I see some import errors for these values in the test; I'm guessing `GetReadingsRe*` types were added to the `api` repo but they have yet to be automatically imported to the Python SDK.

src/viam/components/movement_sensor/movement_sensor.py
benjirewis: @njooma would definitely know. My guess is that now that you're not invoking `super().get_readings()`, you only need `...` here.

randhid: yes we should remove the bespoke implementation in each sdk, we don't need it anymore. It will now be the responsibility of the driver using these wrappers.

njooma: This was modeled after what's done in [rdk](https://github.com/viamrobotics/rdk/blob/12d225bb4da41d580576a44e8047bb5e9e88ce20/components/movementsensor/movementsensor.go#L106) where the base class for Movement Sensor defines the `Readings` function that all subclasses can use.

src/viam/components/movement_sensor/client.py
njooma: Ya exactly -- this has to be merged in after the API changes get merged

src/viam/components/movement_sensor/movement_sensor.py
randhid: @njooma, we added a common.proto `GetReadings` API, so we no longer have to do this in each SDK. Now we have a client and server-side implementation, and we're also changing the [go SDK](https://github.com/viamrobotics/rdk/pull/3047) implementation. So the way it's done now should be changed.

martha-johnston:
@njooma @benjirewis @clintpurser does anyone know if I should be running make buf locally and uploading that? or should those be generated through GitHub actions? seems like test aren't passing because it doesn't have the new api methods (but it should because api was merged)

src/viam/services/navigation/navigation.py
```diff
@@ -5,6 +5,7 @@
 
 from ..service_base import ServiceBase
 from . import GeoObstacle, GeoPoint, Mode, Waypoint
+from viam.proto.service.navigation import Path
```
benjirewis, line 8: @njooma here and elsewhere we were just importing `Path` from `viam.proto.service.navigation`. I see some other API types being imported from `.`; how does that work? We added `Path` (along with `GetPathsRequest` and `GetPathsResponse`) to `src/viam/proto/service/navigation/__init__.py`, but I don't think we were able to find `Path` in `.` here.

tests/test_motion_service.py
```diff
@@ -17,7 +17,8 @@
     "gantry": PoseInFrame(reference_frame="gantry", pose=Pose(x=2, y=3, z=4, o_x=3, o_y=4, o_z=5, theta=21)),
 }
 MOTION_CONFIGURATION = MotionConfiguration(
-    vision_services=[ResourceName(namespace="rdk", type="service", subtype="vision", name="viz1")],
+    obstacle_detectors=[ObstacleDetector(vision_service=ResourceName(namespace="rdk", type="service", subtype="vision", name="viz1"),
+                                         camera=ResourceName(namespace="rdk", type="component", subtype="camera", name="cam1"))],
```
benjirewis, line 21: I assume it's fine to just use a random camera here in the test?

src/viam/services/navigation/client.py
```diff
@@ -36,6 +39,11 @@ def __init__(self, name: str, channel: Channel):
         self.client = NavigationServiceStub(channel)
         super().__init__(name)
 
+    async def get_paths(self, *, timeout: Optional[float] = None) -> List[Path]:
```
benjirewis, line 42: [q] Why do we return API types to users? I thought we would convert the API type to some native Python class and return that to the user?

tests/test_motion_service.py
randhid: Anything that can return `GetObjectPointCluds` for an obstacle. which cameras do.

src/viam/services/navigation/client.py
randhid: I want to request that follow-up API cleanup that doesn't have to do with Martha's readings change be handled in a follow-up ticket. This design does look wrong, though ) I think it should return a Python class but I don't want to make a decision given we're doing this on the fly.

randhid:
@njooma @benjirewis @clintpurser  all good?

randhid (approved):

benjirewis (approved):
LGTM but probably wait for final approval from @njooma

src/viam/services/navigation/client.py
njooma: It's not bad to return a protobuf message. It's actually preferred IMO. This way we don't have to create a bunch of classes that mirror the proto, nor we do we have to keep those classes updated.

src/viam/services/navigation/navigation.py
njooma: The `.` syntax is only if the class is included in the `src/viam/services/navigation/__init__.py`, not the `.../proto/...` file.

src/viam/services/navigation/client.py
benjirewis: @njooma that makes sense. Per recent conversation in the SDKs channel, it seems like it's just language-dependent whether the protos are "good enough" to not need native-mirrored types.

martha-johnston:
@njooma I think I covered everything... I do want movement sensor to no longer be a subclass of sensor so that should be fixed now, I updated the docstring, and added get_readings to ExampleMovementSensor. also updated the nav __init__.py file so path could be imported along with GeoPoint, Waypoint, etc.

src/viam/services/navigation/__init__.py
```diff
@@ -1,5 +1,5 @@
 from viam.proto.common import GeoObstacle, GeoPoint
-from viam.proto.service.navigation import Mode, Waypoint
+from viam.proto.service.navigation import Mode, Waypoint, Path
```
njooma, line 2: This will work, but you should also add it in the `__all__` list below

njooma (approved):
Almost!! One minor change. Approving it because you should be able to merge it in once you add that additional change

tests/test_robot.py
```diff
@@ -60,7 +60,6 @@
     ResourceName(namespace=RESOURCE_NAMESPACE_RDK, type=RESOURCE_TYPE_COMPONENT, subtype="camera", name="camera1"),
     ResourceName(namespace=RESOURCE_NAMESPACE_RDK, type=RESOURCE_TYPE_COMPONENT, subtype="motor", name="motor1"),
     ResourceName(namespace=RESOURCE_NAMESPACE_RDK, type=RESOURCE_TYPE_COMPONENT, subtype="movement_sensor", name="movement_sensor1"),
-    ResourceName(namespace=RESOURCE_NAMESPACE_RDK, type=RESOURCE_TYPE_COMPONENT, subtype="sensor", name="movement_sensor1"),
```
martha-johnston, line 63: made the __all__ change, but this was causing the tests to fail... I removed this line because they were both called "movement_sensor1" but this was of subtype "sensor", and now the tests pass. Just want to make sure this checks out @njooma

njooma: Ohh nice find. There's another spot where this should be updated, but I made a [ticket](https://viam.atlassian.net/browse/RSDK-5461) for it for myself for later (not required for this)

njooma (approved):

PR #466, opened 2023-10-17
[DATA-2034] Add file extensions to binary data capture upload

njooma:
Add ability to provide file_extension to binary data capture upload

### DIFF
--- a/src/viam/app/data_client.py
+++ b/src/viam/app/data_client.py
@@ -437,6 +437,7 @@ class DataClient:
         method_parameters: Optional[Mapping[str, Any]] = None,
         tags: Optional[List[str]] = None,
         data_request_times: Optional[Tuple[datetime, datetime]] = None,
+        extension: Optional[str] = None,
     ) -> str:
         """Upload binary sensor data.
 
@@ -453,6 +454,7 @@ class DataClient:
             tags (Optional[List[str]]): Optional list of tags to allow for tag-based data filtering when retrieving data.
             data_request_times (Optional[Tuple[datetime.datetime, datetime.datetime]]): Optional tuple containing `datetime`s objects
                 denoting the times this data was requested[0] by the robot and received[1] from the appropriate sensor.
+            extension (str): The file extension of binary data including the period, e.g. .jpg, .png, .pcd
 
         Raises:
             GRPCError: If an invalid part ID is passed.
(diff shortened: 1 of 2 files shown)


src/viam/app/data_client.py
```diff
@@ -437,6 +437,7 @@ async def binary_data_capture_upload(
         method_parameters: Optional[Mapping[str, Any]] = None,
         tags: Optional[List[str]] = None,
         data_request_times: Optional[Tuple[datetime, datetime]] = None,
+        file_extension: Optional[str] = None,
```
jeremyrhyde, line 440: Can/does the file extension get translated to mime_type in the cloud?

jeremyrhyde (commented):
Thanks for doing this, the file extension will help us upload pcds, but does it also get translated to mime_type? The current issue we are facing is that the replay camera filters for objects with a mime_type of pcd

src/viam/app/data_client.py
njooma: According to @agreenb, [yes](https://viam.atlassian.net/browse/DATA-2034?focusedCommentId=22732)

> ... which we convert to mime type in the backend

agreenb: Yep! [here](https://github.com/viamrobotics/app/blob/aa701573f796faf5398b51b229c4edf68aa4a023/datasync/data_capture_upload.go#L238) and [here](https://github.com/viamrobotics/app/blob/aa701573f796faf5398b51b229c4edf68aa4a023/datasync/server.go#L135-L165), which uses a [.pcd](https://github.com/viamrobotics/app/blob/aa701573f796faf5398b51b229c4edf68aa4a023/datasync/server.go#L35) file extension to get the point cloud mime type.

agreenb (approved):
LGTM, thanks for the lightning fast fix and supporting if the user doesn't provide a prepended period!

src/viam/app/data_client.py
```diff
@@ -453,6 +454,7 @@ async def binary_data_capture_upload(
             tags (Optional[List[str]]): Optional list of tags to allow for tag-based data filtering when retrieving data.
             data_request_times (Optional[Tuple[datetime.datetime, datetime.datetime]]): Optional tuple containing `datetime`s objects
                 denoting the times this data was requested[0] by the robot and received[1] from the appropriate sensor.
+            file_extension (str): The file extension of binary data including the period, e.g. .jpg, .png, .pcd
```
agreenb, line 457: Could you include that this routes the file to its corresponding mime type in the backend?

njooma: @agreenb how's this for text:

> The backend will route the binary to its corresponding mime type based on this extension.

agreenb: Perfect, thanks!

PR #469, opened 2023-10-23
RSDK-4756 - add `Options` api-key helper function

stuqdog:
#### Major changes

Adds a simple helper function for creating connection `Options` for an api-key, with default values elsewhere.

### DIFF
--- a/src/viam/robot/client.py
+++ b/src/viam/robot/client.py
@@ -1,4 +1,5 @@
 import asyncio
+import re
 from dataclasses import dataclass
 from threading import RLock
 from typing import Any, Dict, List, Optional, Union
@@ -110,9 +111,17 @@ class RobotClient:
                 api_key (str): your API key
                 api_key_id (str): your API key ID
 
-            returns:
+            Raises:
+                ValueError: Raised if the `api_key_id` is not a valid uuid
+
+            Returns:
                 Self: the RobotClient options
             """
+            key_id_regex = re.compile(r"^[0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12}$")
+            match = key_id_regex.match(api_key_id)
+            if not match:
+                raise ValueError(f"{api_key_id} is not a valid UUID")
+
             credentials = Credentials(type="api-key", payload=api_key)
             dial_opts = DialOptions(credentials=credentials, auth_entity=api_key_id)
             self = cls()
--- a/src/viam/robot/client.py
+++ b/src/viam/robot/client.py
@@ -1,5 +1,4 @@
 import asyncio
-import re
 from dataclasses import dataclass
 from threading import RLock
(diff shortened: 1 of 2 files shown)


ohEmily:
From Ethan over Slack:
```
    opts = RobotClient.Options.with_api_key(api_key="1xn3n45qvmkmnad3dv7mgijufuo6gucg", api_key_id="48d18aff-d1f6-4927-bf5e-36c9c22e2233")
    return await RobotClient.at_address("viam-conn-test-remotes-main.tkofq1ck33.viam.cloud", opts)
```

stuqdog:
> LGTM!
> 
> Maybe one nice-to-have for which we might already have a helper in the codebase would be regex validation that the `api_key` isn't a UUID (or just check if there are dash characters). I suspect the #1 user error will be people mixing up the API key with the API key ID and this could help on that front.
> 
> Tagging in @jr22 in case you want to take a peek.

yeah that seems easy enough to do! can add :)

ohEmily:
Screenshot 2023-10-23 at 3 10 33 PM

(deleted this key in Atlas since I exposed it and it's technically connected to viam-dev.)

ohEmily:
(deleted this key in Atlas since I exposed it and it's technically connected to viam-dev.)

src/viam/robot/client.py
```diff
@@ -101,6 +102,32 @@ class Options:
         Whether sessions are disabled
         """
 
+        @classmethod
+        def with_api_key(cls, api_key: str, api_key_id: str) -> Self:
+            """
+            Create dial options with an API key for credentials and default values for other arguments.
```
njooma, line 108: These are not `dial options` but rather `RobotClient.Options`

njooma (commented):
Might be useful to actually have this as a convenience method on `DialOptions`, and have this method call that one

src/viam/robot/client.py
```diff
@@ -101,6 +102,32 @@ class Options:
...
+            Raises:
+                ValueError: Raised if the `api_key_id` is not a valid uuid
+
+            Returns:
+                Self: the RobotClient options
+            """
+            key_id_regex = re.compile(r"^[0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12}$")
+            match = key_id_regex.match(api_key_id)
+            if not match:
+                raise ValueError(f"{api_key_id} is not a valid UUID")
```
njooma, line 123: This might be easier using the UUID library in python:

```python3
import uuid

...

try:
    uuid.UUID(api_key_id)
except ValueError:
    raise ValueError(f"{api_key_id} is not a valid UUID")
```

jr22 (approved):
thanks for the tag @ohEmily-- took a peek and LGTM as well. thanks!

src/viam/robot/client.py
stuqdog: ooh thanks for the tip, didn't know about the `UUID` library!

njooma (approved):

PR #470, opened 2023-10-23
Update reqs for python 3.12

njooma:
[Python 3.12](https://docs.python.org/3.12/whatsnew/3.12.html#what-s-new-in-python-3-12) was released on October 2. This PR updates our requirements so that we can support it.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,7 +19,7 @@ jobs:
       image: python:${{ matrix.python-version }}
     strategy:
       matrix:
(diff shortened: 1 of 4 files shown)


maximpertsov (approved):

njooma:
This cannot be merged as it will fail for modules/servers, due to an issue with the [gRPC server not staying open](https://github.com/vmagamedov/grpclib/issues/182). I'm closing for now, and will reopen when the underlying gRPC issue is fixed

PR #471, opened 2023-10-25
RSDK-4907 - Add billing client

stuqdog:
#### Major changes
Adds billing client to python SDK.

### DIFF
--- a/src/viam/app/billing_client.py
+++ b/src/viam/app/billing_client.py
@@ -1,0 +1,1 @@
+import foo
--- a/src/viam/app/billing_client.py
+++ b/src/viam/app/billing_client.py
@@ -63,9 +63,8 @@ class BillingClient:
         request = GetInvoicePdfRequest(id=invoice_id, org_id=org_id)
         response: GetInvoicePdfResponse = await self._billing_client.GetInvoicePdf(request, metadata=self._metadata, timeout=timeout)
         data: bytes = response[0].chunk
-        file = open(dest, "wb")
-        file.write(data)
-        file.flush()
+        with open(dest, "wb") as file:
+            file.write(data)
 
     async def get_invoices_summary(self, org_id: str, timeout: Optional[float] = None) -> GetInvoicesSummaryResponse:
         """Access total outstanding balance plus invoice summaries for a given org.
--- a/src/viam/app/billing_client.py
+++ b/src/viam/app/billing_client.py
@@ -88,4 +88,3 @@ class BillingClient:
             viam.proto.app.billing.GetOrgBillingInformationResponse: The org billing information"""
         request = GetOrgBillingInformationRequest(org_id=org_id)
         return await self._billing_client.GetOrgBillingInformation(request, metadata=self._metadata, timeout=timeout)
-
--- a/src/viam/app/billing_client.py
+++ b/src/viam/app/billing_client.py
@@ -5,24 +5,14 @@ from grpclib.client import Channel
 from viam import logging
 from viam.proto.app.billing import (
         BillingServiceStub,
-        GetBillingSummaryRequest,
-        GetBillingSummaryResponse,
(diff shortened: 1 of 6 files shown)


tests/mocks/services.py
```diff
@@ -818,6 +839,64 @@ async def CancelTrainingJob(self, stream: Stream[CancelTrainingJobRequest, Cance
...
+        raise NotImplementedError()
+
+    async def GetInvoiceHistory(self, stream: Stream[GetInvoiceHistoryRequest, GetInvoiceHistoryResponse]) -> None:
+        raise NotImplementedError()
+
+    async def GetItemizedInvoice(self, stream: Stream[GetItemizedInvoiceRequest, GetItemizedInvoiceResponse]) -> None:
+        raise NotImplementedError()
+
+    async def GetUnpaidBalance(self, stream: Stream[GetUnpaidBalanceRequest, GetUnpaidBalanceResponse]) -> None:
+        raise NotImplementedError()
```
stuqdog, line 897: These methods are slated to be deprecated and so are not being implemented.

tests/test_billing_client.py
```diff
@@ -0,0 +1,112 @@
...
+    async def test_get_current_month_usage(self, service: MockBilling):
+        async with ChannelFor([service]) as channel:
+            org_id = "foo"
+            client = BillingClient(channel, BILLING_SERVICE_METADATA)
+            curr_month_usage = await client.get_current_month_usage(org_id=org_id)
+            assert curr_month_usage == CURR_MONTH_USAGE
+            assert service.org_id == org_id
+
+    @pytest.mark.asyncio
+    async def test_get_invoice_pdf(self, service: MockBilling):
```
stuqdog, line 93: Note to reviewers: the client code for `get_invoice_pdf` saves the pdf to a file. I didn't really want to mess around with having a test in CI that creates and modifies files, and the only other alternative I could think of (make the filepath optional, return the pdf's byte array instead of nothing, and then compare to that return value) I think would be a bad API change to the `get_invoice_pdf` method (the byte array is not particularly human parse-able or useful outside of being saved to a pdf). In local testing I was able to download and save invoice PDFs successfully so I figured that was good enough. If there's any disagreement with the above, let me know!

src/viam/app/billing_client.py
```diff
@@ -0,0 +1,90 @@
...
+        Args:
+            invoice_id (str): the ID of the invoice being requested
+            org_id (str): the ID of the org to request data from
+            dest (str): filepath to save the invoice to
+        """
+        request = GetInvoicePdfRequest(id=invoice_id, org_id=org_id)
+        response: GetInvoicePdfResponse = await self._billing_client.GetInvoicePdf(request, metadata=self._metadata, timeout=timeout)
+        data: bytes = response[0].chunk
+        file = open(dest, "wb")
+        file.write(data)
```
dgottlieb, line 67: This should be followed with a `file.flush`.

And probably an `os.fsync(file)`, but that's less likely to prevent a surprising behavior.

tests/test_billing_client.py
dgottlieb: Sounds reasonable to me

src/viam/app/billing_client.py
stuqdog: Ooh good callout, thanks! Also adding `file.flush()` to other app client methods that write to file.

dgottlieb (approved):

src/viam/app/billing_client.py
dgottlieb: Post-approve realization mentioned in person: We should be `file.close()`ing here, or `with open(...) as file:`. Which omits the need for a `flush` here.

stuqdog: Just switched to `with open`

dgottlieb (approved):
still lgtm

PR #473, opened 2023-10-26
await example close calls

purplenicole730:
Fix from this [PR](https://github.com/viamrobotics/viam-python-sdk/pull/458).
`close()` calls have to be awaited.


cheukt (approved):
LGTM

PR #478, opened 2023-11-02
RSDK-4529 - add missing app apis

stuqdog:


src/viam/app/app_client.py
```diff
@@ -374,11 +449,27 @@ async def list_organizations(self) -> List[Organization]:
...
+            org_id (Optional[str]): ID of the organization to query. If None, defaults to the
+            currently-authed org.
+
+        Raises:
+            GRPCError: If the provided org_id is invalid, or not currently authed to.
+
+        Returns:
+            viam.proto.app.Organization: The requested organization.
+        """
+        org_id = org_id if org_id else await self._get_organization_id()
```
maximpertsov, line 469: Similar issue as an above comment - is (`""`) a valid org_id? If so, we need to check for `not None` explicitly. If not, we might want to update the doc-string to something like `If None or blank/empty`.

```suggestion
        org_id = org_id if org_id is not None else await self._get_organization_id()
```

src/viam/app/app_client.py
```diff
@@ -350,6 +394,10 @@ def __init__(self, channel: Channel, metadata: Mapping[str, str], location_id: O
     _channel: Channel
     _organization_id: Optional[str] = None
 
+    # TODO(RSDK-5569): this method will no longer be reliable when we're not authing as an org.
+    # consider what assumptions we _can_ make (e.g., if `list_orgs` only returns a single org
+    # then we can probably still rely on it for org id?), and also update other APIs to ask for
+    # an org id (optionally?) instead of just assuming we'll get it from this method.
     async def _get_organization_id(self) -> str:
         return self._organization_id if self._organization_id else await self._request_organization_id()
```
maximpertsov, line 402: Probably a perverse case, but do we want to request a new org_id if the current org_id is an empty string (`""`)? If that is a valid org_id for some reason then we need to check for `not None` explicitly since empty strings are falsy in python.

```suggestion
        return self._organization_id if self._organization_id is not None else await self._request_organization_id()
```

src/viam/app/app_client.py
```diff
@@ -400,12 +491,50 @@ async def get_organization_namespace_availability(self, public_namespace: str) -
...
+        """Deletes an organization.
+
+        Args:
+            org_id (Optional[str]): ID of the organization to delete. If None, defaults to the
+            currently-authed org.
+
+        Raises:
+            GRPCError: if the organization is not authed to, or has locations in it.
+        """
+        organization_id = org_id if org_id else await self._get_organization_id()
```
maximpertsov, line 535: same-ish Q as noted above

```suggestion
        organization_id = org_id if org_id is not None else await self._get_organization_id()
```

src/viam/app/app_client.py
```diff
@@ -1200,6 +1394,50 @@ async def list_modules(self) -> List[Module]:
...
+            with the key.
+            name (Optional[str]): A name for the key. If None, defaults to the current timestamp.
+
+        Raises:
+            GRPCError: If the authorizations list is empty.
+
+        Returns:
+            Tuple[str, str]: The api key and api key ID.
+        """
+        name = name if name else str(datetime.now())
```
maximpertsov, line 1413: same-ish Q as noted above

```suggestion
        name = name if name is not None else str(datetime.now())
```

maximpertsov (changes requested):
some questions about the validity of blank orgs and keys

maximpertsov (approved):

src/viam/app/app_client.py
```diff
@@ -324,6 +347,27 @@ def proto(self) -> RobotPartHistoryEntryPB:
         )
 
 
+class APIKeyAuthorization:
+    """A class with the necessary authorization data for creating an API key.
+
+    Use this class when constructing API key authorizations to minimize the risk of malformed or missing data.
+    """
+
+    def __init__(self, role: str, resource_type: str, resource_id: str):
```
njooma, line 356: You might want to make these type hints `Union[Literal["owner"], Literal["operator"]]` and ``Union[Literal["organization"], Literal["location"], Literal["robot"]]`

src/viam/app/app_client.py
```diff
@@ -324,6 +347,27 @@ def proto(self) -> RobotPartHistoryEntryPB:
...
+        Must match `resource_id`.
+        resource_id (str): ID of the resource the role applies to (i.e., either an organization, location, or robot ID).
+        """
+        self._role = role
+        self._resource_type = resource_type
+        self._resource_id = resource_id
+
+    _role: str
+    _resource_type: str
+    _resource_id: str
```
njooma, line 368: Is there a reason for these being private?

src/viam/app/app_client.py
```diff
@@ -1043,13 +1224,12 @@ async def add_role(self, identity_id: str, role: str, resource_type: str, resour
         Raises:
             GRPCError: If either an invalid identity ID, role ID, resource type, or resource ID is passed.
         """
-        organization_id = await self._get_organization_id()
-        authorization = Authorization(
+        authorization = await self._create_authorization(
             identity_id=identity_id,
-            authorization_id=f"{resource_type}_{role}",
+            identity_type="",
+            role=role,
```
njooma, line 1230: Same note as above, perhaps make the parameter type hints be unions of literals

njooma (changes requested):

src/viam/app/app_client.py
stuqdog: In general I favor minimizing direct access to data members. In this case the class is kind of weird/bespoke compromise (it's an opinionated version of the API key authorization information for a single, specific use case) and so I favored minimizing the amount of interaction users have with it. It's not a super strong preference though, happy to revisit if you disagree!

PR #480, opened 2023-11-02
Fix simple module, update readme

cheukt:
updated client as well
as part of testing, found a small issue with our validation code so fixed that as well

also made all the logs info so it'll show up even if not in debug mode

@dannenberg think this is what you wanted

### DIFF
--- a/examples/simple_module/client.py
+++ b/examples/simple_module/client.py
@@ -7,9 +7,9 @@ from viam.components.sensor import Sensor
 
 
 async def connect():
-    # creds = Credentials(type="<your authentication type here>", payload="<your authentication payload here>")
-    opts = RobotClient.Options(refresh_interval=0, dial_options=DialOptions(insecure=True), log_level=logging.DEBUG)
-    return await RobotClient.at_address("localhost:8080", opts)
+    creds = Credentials(type="<your authentication type here>", payload="<your authentication payload here>")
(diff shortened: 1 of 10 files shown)


examples/simple_module/client.py
```diff
@@ -7,9 +7,9 @@
 
 
 async def connect():
-    creds = Credentials(type="<your authentication type here>", payload="<your authentication payload here>")
-    opts = RobotClient.Options(refresh_interval=0, dial_options=DialOptions(credentials=creds), log_level=logging.DEBUG)
-    return await RobotClient.at_address("<your robot uri here>", opts)
+    # creds = Credentials(type="<your authentication type here>", payload="<your authentication payload here>")
```
maximpertsov, line 10: assuming you meant to uncomment

```suggestion
    creds = Credentials(type="<your authentication type here>", payload="<your authentication payload here>")
```

maximpertsov: nvm i see we are just changing to insecure

purplenicole730: Might have to un-import `Credentials` now that we're not using it anymore.

purplenicole730 (approved):
LGTM

examples/simple_module/client.py
cheukt: ah that's unintentional, will revert that part

dannenberg (approved):
Thank you!! <3

PR #481, opened 2023-11-03
RSDK-5637  Remove mime type inference of PCD within get_image

hexbabe:
https://viam.atlassian.net/browse/RSDK-5637


kharijarrett (approved):
100% LGTM. GetImage is for 2D images only, this should never be allowed to return a pcd

bazile-clyde:
> 100% LGTM. GetImage is for 2D images only, this should never be allowed to return a pcd

For the record, I agree!

bazile-clyde (approved):

hexbabe:
hmm, now I’m wondering if any Python (lidar maybe?) modules depend on this functionality and if its breaking anything

cheukt (approved):

PR #483, opened 2023-11-06
DATA-1988 - Make file ext manadatory for binary_data_capture_upload, update docstrings

stuqdog:


dgottlieb (approved):
lgtm. I suppose we'll have to note this is backwards breaking?

maximpertsov (changes requested):
I slightly prefer keeping this as named argument but making it mandatory - we have a lot of position arguments and this will minimize breakage to existing code.

maximpertsov (approved):

PR #484, opened 2023-11-06
RSDK-4941 - use org api key in examples

stuqdog:


benjirewis (approved):

purplenicole730 (approved):
I wonder why logging was imported if it wasn't needed? Otherwise, LGTM!

stuqdog:
> I wonder why logging was imported if it wasn't needed? Otherwise, LGTM!

@purplenicole730 I think it was being used before in creating the options (`log_level=logging.DEBUG`) but the `with_api_key` method just goes with defaults.

purplenicole730:
> > I wonder why logging was imported if it wasn't needed? Otherwise, LGTM!
> 
> @purplenicole730 I think it was being used before in creating the options (`log_level=logging.DEBUG`) but the `with_api_key` method just goes with defaults.

Oh there it is! I forgot that we set `log_level` in the options and was more focused on the function change. Ignore me, haha 😁

PR #486, opened 2023-11-09
RSDK-4650 - Add streaming data capture upload

stuqdog:
Implements `streaming_data_capture_upload`

### DIFF
--- a/src/viam/app/data_client.py
+++ b/src/viam/app/data_client.py
@@ -540,8 +540,8 @@ class DataClient:
                 SensorData(
                     metadata=(
                         SensorMetadata(
-                            time_requested=datetime_to_timestamp(data_request_times[idx][0]) if data_request_times[idx] else None,
-                            time_received=datetime_to_timestamp(data_request_times[idx][1]) if data_request_times[idx] else None,
+                            time_requested=datetime_to_timestamp(data_request_times[idx][0]) if data_request_times else None,
+                            time_received=datetime_to_timestamp(data_request_times[idx][1]) if data_request_times else None,
                         )
                         if data_request_times[idx]
                         else None
--- a/src/viam/app/data_client.py
+++ b/src/viam/app/data_client.py
@@ -623,16 +623,6 @@ class DataClient:
                 await stream.recv_trailing_metadata()  # causes us to throw appropriate gRPC error
             return response.file_id
 
-    async def _file_upload(self, metadata: UploadMetadata, file_contents: FileData) -> FileUploadResponse:
-        request_metadata = FileUploadRequest(metadata=metadata)
-        request_file_contents = FileUploadRequest(file_contents=file_contents)
-        async with self._data_sync_client.FileUpload.open(metadata=self._metadata) as stream:
-            await stream.send_message(request_metadata)
-            await stream.send_message(request_file_contents, end=True)
-            response = await stream.recv_message()
-            assert response is not None
-            return response
-
     async def file_upload(
         self,
         part_id: str,
--- a/src/viam/app/data_client.py
+++ b/src/viam/app/data_client.py
@@ -579,6 +579,7 @@ class DataClient:
         method_name: Optional[str] = None,
         method_parameters: Optional[Mapping[str, Any]] = None,
         data_request_times: Optional[Tuple[datetime, datetime]] = None,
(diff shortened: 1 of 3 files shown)


src/viam/app/data_client.py
```diff
@@ -466,8 +469,8 @@ async def binary_data_capture_upload(
         sensor_contents = SensorData(
             metadata=(
                 SensorMetadata(
-                    time_requested=datetime_to_timestamp(data_request_times[0]) if data_request_times[0] else None,
```
stuqdog, line 469: (flyby) this would have caused a runtime error in cases where the (ostensibly optional) tuple was None.

src/viam/app/data_client.py
```diff
@@ -537,8 +540,8 @@ async def tabular_data_capture_upload(
                 SensorData(
                     metadata=(
                         SensorMetadata(
-                            time_requested=datetime_to_timestamp(data_request_times[idx][0]) if data_request_times[idx][0] else None,
```
stuqdog, line 540: (flyby) this would have caused a runtime error in cases where the (ostensibly optional) list of tuples was None.

src/viam/app/data_client.py
```diff
@@ -466,8 +469,8 @@ async def binary_data_capture_upload(
         sensor_contents = SensorData(
             metadata=(
                 SensorMetadata(
-                    time_requested=datetime_to_timestamp(data_request_times[0]) if data_request_times[0] else None,
-                    time_received=datetime_to_timestamp(data_request_times[1]) if data_request_times[1] else None,
+                    time_requested=datetime_to_timestamp(data_request_times[0]) if data_request_times else None,
+                    time_received=datetime_to_timestamp(data_request_times[1]) if data_request_times else None,
```
benjirewis, line 473: What if `data_request_times[0] and not data_request_times[1]`? Same question below.

src/viam/app/data_client.py
```diff
@@ -566,6 +569,63 @@ async def _data_capture_upload(self, metadata: UploadMetadata, sensor_contents:
...
+        """Uploads the metadata and contents of streaming binary data.
+
+        Args:
+            data (bytes): the data to be uploaded.
+            part_id (str): Part ID of the resource associated with the file.
+            file_ext (str): file extension type for the data. required for determining MIME type.
+            component_type (Optional[str]): Optional type of the component associated with the file (e.g., "movement_sensor").
+            component_name (Optional[str]): Optional name of the component associated with the file.
+            method_name (Optional[str]): Optional name of the method associated with the file.
+            method_parameters (Optional[str]): Optional dictionary of the method parameters. No longer in active use.
```
benjirewis, line 593: What does "No longer in active use" mean? Should we not even have this parameter in the SDK if it's "deprecated"?

benjirewis (commented):
Just some qs! Don't know much about this API.

purplenicole730 (approved):
LGTM! I'm not too familiar with this function, so it took me a while to get through. Nice tests as well!

src/viam/app/data_client.py
stuqdog: So `data_request_times` is annotated as an `Optional[Tuple[datetime, datetime]]`, not a `Tuple[Optional[datetime], Optional[datetime]]`, so there should either be a tuple with two `Some`s or a `None`. So I think the case you're describing is one that's inconsistent with how we expect input. Definitely open to discussing changing what we expect though!

src/viam/app/data_client.py
stuqdog: Yeah, good question! The `no longer in active use` language is cribbed from how `method_parameters` is annotated elsewhere in the SDK. I don't remember the details precisely, but I remember conversations with the data team basically being "We don't use this now, but we should still allow it to be passed because it does update on the backend."

src/viam/app/data_client.py
benjirewis: Ah no I misunderstood the typing; that sounds right to me thanks for the context!

src/viam/app/data_client.py
benjirewis: Cool; so sort just short of deprecated. Seems fine to me, just wanted to call out.

benjirewis (approved):

PR #489, opened 2023-11-15
[DATA-2065] Make data required in file upload

njooma:


benjirewis (approved):
@purplenicole730 probably has more context than I do, but seems good!

njooma:
@purplenicole730 is on vacay, so pinging @stuqdog instead

src/viam/app/data_client.py
```diff
@@ -659,25 +659,27 @@ async def streaming_data_capture_upload(
...
         request_metadata = StreamingDataCaptureUploadRequest(metadata=metadata)
+        stream: Stream[StreamingDataCaptureUploadRequest, StreamingDataCaptureUploadResponse]
         async with self._data_sync_client.StreamingDataCaptureUpload.open(metadata=self._metadata) as stream:
             await stream.send_message(request_metadata)
             await stream.send_message(StreamingDataCaptureUploadRequest(data=data), end=True)
-            response: StreamingDataCaptureUploadResponse = await stream.recv_message()
+            response = await stream.recv_message()
             if not response:
                 await stream.recv_trailing_metadata()  # causes us to throw appropriate gRPC error
+                raise TypeError("Response cannot be empty")
```
stuqdog, line 669: (q) Is it possible for us to actually reach this point? `if not response` is the case when the response isn't just empty, but a `None`, which I think should only happen if something went wrong such that we'd have a non-OK gRPC status and so throw an error at the preceding line. I guess either way there's no harm in having another safety rail here!

stuqdog (approved):
one question, but looks good to me!

src/viam/app/data_client.py
njooma: *WE* know we can't get here, but they *typechecker* doesn't know that and it was complaining

stuqdog: Ahh got it, thanks!

PR #491, opened 2023-11-22
[RSDK-5823] remove Board.model_attributes() from the SDK

penguinland:
This was removed from the RDK in https://github.com/viamrobotics/rdk/pull/3253. From the perspective of the Python SDK, no information is being communicated with this function: it always says that the board is remote and nothing else.

Strictly speaking, this is a breaking change because we don't know who out in the world is using this function. but my understanding is that the entire SDK is still in beta so breaking changes are okay. and anyone using it isn't getting any utility out of it, as previously stated.


njooma (approved):
LGTM % lint errors

PR #503, opened 2023-12-11
[RSDK-5810] remove impossible-to-implement functions from digital interrupts

penguinland:
Within digital interrupts, the `value()` method will get the number of interrupts that have occurred. It's a good thing to have; we'll keep it. However,
- `add_callback()` cannot be implemented without changing the GPRC interface, which will require a scope doc, which will require planning how to do this right. Get rid of it for now: digital interrupts aren't really part of the API yet.
- `add_post_processor()` doesn't exist outside of raspberry pi boards; get rid of it.
- `tick()` is only used in tests (to simulate the interrupt firing). Keep it in `MockDigitalInterrupt`, but get rid of it in the "real" code.

So, how do you use digital interrupts outside of the RDK server? Unfortunately, you kinda can't, and have never been able to. You can repeatedly check its value and do something if you notice the value has changed, but a proper GRPC interface for digital interrupts doesn't currently exist. We shouldn't pretend that it does.

### DIFF
--- a/examples/server/v1/components.py
+++ b/examples/server/v1/components.py
@@ -20,7 +20,6 @@ from viam.components.arm import Arm
 from viam.components.audio_input import AudioInput
 from viam.components.base import Base
 from viam.components.board import Board
-from viam.components.board.board import PostProcessor
 from viam.components.camera import Camera
 from viam.components.encoder import Encoder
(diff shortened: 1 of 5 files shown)


njooma (approved):
Love red PRs

PR #506, opened 2023-12-21
RSDK-5936: change return type for get readings

purplenicole730:
The return type for the `get_readings()` calls should be of `Mapping[str, viam.utils.ValueTypes]` rather than `Mapping[str, Any]`.

Tested with a fake movement sensor and power sensor.
```python
fakeMovementSensor get_readings return value:
{'linear_acceleration': x: 2.2
y: 4.5
z: 2
, 'angular_velocity': z: 1
, 'compass': 25.0, 'position': latitude: 40.7
longitude: -73.98
, 'linear_velocity': y: 5.4
, 'orientation': {'r': 1.0, 'i': 0.0, '_type': 'quat', 'k': 0.0, 'j': 0.0}, 'altitude': 50.5}
fake get_readings return value:
{'volts': 1.5, 'watts': 9.8, 'is_ac': True, 'amps': 2.2}
```


martha-johnston (commented):
lgtm assuming you have tested with a real (or fake) sensor not just the python tests. just want to make sure changing the type didn't mess up something somewhere else

martha-johnston (approved):

src/viam/utils.py
```diff
@@ -190,7 +190,7 @@ def sensor_readings_native_to_value(readings: Mapping[str, Any]) -> Mapping[str,
     return {key: primitive_to_value(value) for (key, value) in prim_readings.items()}
 
 
-def sensor_readings_value_to_native(readings: Mapping[str, Value]) -> Mapping[str, Any]:
+def sensor_readings_value_to_native(readings: Mapping[str, Value]) -> Mapping[str, ValueTypes]:
```
njooma, line 193: I see now why we had it as `Any`: `Vector3`, `GeoPoint`, and `Orientation` are not `ValueTypes`. So this new signature would not pass type checking. (It'll still work because python doesn't actually care about types.)

To make this return type more accurate, I suggest we create a new type alias: `SensorReading`, which is a `Union[ValueTypes, Orientation, GeoPoint, Vector3]`. So the final return type would be: `Mapping[str, SensorReading]`.

This way, if we ever support other sensor reading types (`EulerAngles`, `Quaternion`, etc.), we only have to update it here, and not for each method signature.

njooma (changes requested):

src/viam/utils.py
```diff
@@ -172,7 +175,7 @@ async def get_geometries(
     return [geometry for geometry in response.geometries]
 
 
-def sensor_readings_native_to_value(readings: Mapping[str, Any]) -> Mapping[str, Any]:
+def sensor_readings_native_to_value(readings: Mapping[str, Any]) -> Mapping[str, Value]:
```
njooma, line 178: Good catch

njooma (approved):

PR #510, opened 2023-12-27
RSDK-6034 Add GetProperties to SLAM and Nav

jeremyrhyde:
This PR adds the GetProperties endpoints for both the slam and navigation services in viam's python sdk. The new endpoint was added in [this](https://github.com/viamrobotics/api/commit/223673cee35c340ea2fb40b742680c5104abdafc) commit to the API.

**JIRA Ticket:** https://viam.atlassian.net/browse/RSDK-6034


src/viam/services/slam/slam.py
```diff
@@ -59,3 +59,14 @@ async def get_position(self, *, timeout: Optional[float]) -> Pose:
...
         """
         ...
+
+    @abc.abstractmethod
+    async def get_properties(self, *, timeout: Optional[float]) ->  Tuple[bool, MappingMode.ValueType]:
+        """
+        Get information regarding the current SLAM session.
+
+        Returns:
+            Tuple[bool, MappingMode.ValueType]: A tuple of a boolean values representing if the SLAM session is being run in
```
purplenicole730, line 69:
```suggestion
            Tuple[bool, MappingMode.ValueType]: A tuple of a boolean value representing if the SLAM session is being run in

File: tests/test_slam.py [hunk 2]
```diff
@@ -46,6 +48,12 @@ async def test_get_latest_map_info(self):
         time = await self.slam.get_latest_map_info()
         assert time == MockSLAM.LAST_UPDATE
 
+    @pytest.mark.asyncio
+    async def test_get_properties(self):
+        (cloud_slam, mapping_mode) = await self.slam.get_properties()
+        assert cloud_slam == MockSLAM.CLOUD_SLAM
+        assert mapping_mode == MockSLAM.MAPPING_MODE
```
purplenicole730, line 55: Does `MockSLAM` have a `CLOUD_SLAM` and `MAPPING_MODE` object? I don't see it in `tests/mocks/services.py > MockSLAM`.

purplenicole730 (changes requested):
Missing `MockSLAM` properties for your tests, unless they're defined somewhere else that I just missed?

The branch is called adding `get_properties()` to both slam AND nav, but this PR seems to only add the function to SLAM. Is Nav not getting this function in Python?

src/viam/services/slam/slam.py
purplenicole730: nit^

tests/test_slam.py
jeremyrhyde: Forgot to push that file, update should include it

tests/test_board.py
```diff
@@ -175,15 +175,11 @@ async def test_get_digital_interrupt_value(self, board: MockBoard, service: Boar
         async with ChannelFor([service]) as channel:
             client = BoardServiceStub(channel)
 
-            request = GetDigitalInterruptValueRequest(
-                board_name=board.name, digital_interrupt_name="dne"
-            )
+            request = GetDigitalInterruptValueRequest(board_name=board.name, digital_interrupt_name="dne")
```
jeremyrhyde, line 178: [note] not sure why this changes, was automatic with formater

jeremyrhyde (commented):
All tests affected by the scope of this PR pass however I am seeing unexpected errors involving the app client. Any suggestions for resolving these would be appreciated

randhid:
> All tests affected by the scope of this PR pass however I am seeing unexpected errors involving the app client. Any suggestions for resolving these would be appreciated

You've got to only pull in the relevant API changes, you're also pulling in other API changes if you look at the files that are added in your diff.

src/viam/services/navigation/navigation.py
```diff
@@ -49,3 +49,7 @@ async def get_mode(self, *, timeout: Optional[float]) -> Mode.ValueType:
     @abc.abstractmethod
     async def set_mode(self, mode: Mode.ValueType, *, timeout: Optional[float]):
         ...
+
+    @abc.abstractmethod
+    async def get_properties(self, *, timeout: Optional[float]) -> MapType:
```
purplenicole730, line 54: The return isn't `MapType.ValueType`?

purplenicole730 (approved):
LGTM, although maybe don't include the additions to the app generated files

njooma:
hope you don't mind @jeremyrhyde, I pushed an update that should fix the app test failures you were seeing. However, you'll still see a failure because you also need to update the documentation example server [here](https://github.com/viamrobotics/viam-python-sdk/blob/b2102576fcc5b669dd5273703a380e74507f5321/examples/server/v1/services.py#L23)

src/viam/services/navigation/navigation.py
jeremyrhyde: Great catch nicole!

njooma (approved):

stuqdog (approved):

PR #512, opened 2024-01-02
[RSDK-878] Typechecking

njooma:
Add typechecking to the python SDK. I chose [pyright](https://github.com/microsoft/pyright) over [mypy](https://mypy-lang.org/) for a few reasons:
1. pyright is the built-in type checker in VS Code
2. pyright is a bit more lenient, so easier to get started with
3. pyright seems to get updates a bit more frequently than mypy

We will likely still want to eventually move over to mypy, but this is a great starting place for us. 

Some notes on why I did what I did:
* `__eq__` functions
  * Technically equality functions should accept any object. So when we wrote our own `__eq__` methods that were typed to the type of self, it was an incorrect override
* `SUBTYPE` in resources
  * the `SUBTYPE` field is required for all resource types (anything that extends from `ResourceBase`). When we actually implement that in our concrete subtypes, we want it to be `Final`. However, since the original abstract definition is not `Final` (because it's abstract), the typechecker views it as an incompatible override. So I ignore it
* `**__` in method functions
  * Wherever an abstract resource type accepts `**kwargs` as parameters, those need to be included in the client to make a valid override. However, those would go unused. So I added `**__` to signify unused kwargs
* Component RPC services are now generic over component
  * You can genericize your RPC service so that when we call `self.get_resource`, it's correctly typed
* Board sub-components (e.g. GPIOPin) no longer subclass `ComponentBase`
  * They are not components and do not need the additional features that subclassing gives them, e.g. `from_robot`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,6 +46,7 @@ include = ["LICENSE", "src/viam/rpc/libviam_rust_utils.*"]
     {version = "<1.25.0", python = "<3.9"},
     {version = ">=1.25.0", python = ">=3.9"},
   ]
+  pyright = "^1.1.336"
 
 [tool.pytest.ini_options]
 addopts = "-ra"
@@ -72,3 +73,7 @@ build-backend = "poetry.core.masonry.api"
 
 [tool.poetry.extras]
 mlmodel = ["numpy"]
+
+[tool.pyright]
+include = [ "src" ]
+exclude = [ "**/gen", "**/proto" ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,30 +23,34 @@ include = ["LICENSE", "src/viam/rpc/libviam_rust_utils.*"]
   numpy = { version = ">=1.21,<2", optional = true }
 
   [tool.poetry.group.dev.dependencies]
-  pytest = "^7.4.2"
-  pytest-asyncio = "^0.21.1"
+  pytest = "^7.4.3"
+  pytest-asyncio = "^0.23.2"
   coverage = "^7.3.2"
   protoletariat = "^3.2.19"
   jupyter = "^1.0.0"
   flake8 = "^6.1.0"
-  myst-nb = "^0.17.2"
-  sphinx-autoapi = "^2.1.1"
-  sphinx-rtd-theme = "^1.3.0"
+  myst-nb = [
+    {version = "<1.0.0", python = "<3.9"},
+    {version = ">=1.0.0", python = ">=3.9"},
+  ]
+  sphinx-autoapi = [
+    {version = "<3.0.0", python = "<3.9"},
+    {version = ">=3.0.0", python = ">=3.9"},
+  ]
+  sphinx-rtd-theme = "^2.0.0"
   autopep8 = "^2.0.4"
   black = {extras = ["jupyter"], version = "^23.12.0"}
-  nbmake = "^1.4.5"
-  types-pillow = "^10.0.0.3"
+  nbmake = "^1.4.6"
+  types-pillow = "^10.1.0.2"
   mypy-protobuf = "^3.5.0"
-  tox = "^4.11.3"
+  tox = "^4.11.4"
   isort = "^5.12.0"
   pytest-watcher = "^0.3.4"
-  # TODO(RSDK-5450) - Update myst-nb so this can be unpinned.
-  astroid = "<=2.15.6"
   numpy = [
     {version = "<1.25.0", python = "<3.9"},
-    {version = ">=1.25.0", python = ">=3.9"},
+    {version = ">=1.26.2", python = ">=3.9"},
   ]
-  pyright = "^1.1.336"
+  pyright = "^1.1.339"
 
 [tool.pytest.ini_options]
 addopts = "-ra"
--- a/src/viam/__init__.py
+++ b/src/viam/__init__.py
@@ -43,14 +43,20 @@ sys.excepthook = _log_exceptions
 ##################
 def _rname_str(self: _ResourceName) -> str:
     return f"{self.namespace}:{self.type}:{self.subtype}/{self.name}"
+
+
 _ResourceName.__str__ = _rname_str  # noqa E305
 
 
 def _rname_repr(self: _ResourceName) -> str:
     return f"<viam.proto.common.ResourceName {str(self)} at {hex(id(self))}>"
+
+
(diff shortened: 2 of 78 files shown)


njooma:
@randhid @penguinland FYI this PR removes the subclassing from GPIOPin, AnalogReader, DigitalIterrupt

src/viam/app/billing_client.py
```diff
@@ -60,11 +60,12 @@ async def get_invoice_pdf(self, invoice_id: str, org_id: str, dest: str, timeout
             org_id (str): the ID of the org to request data from
             dest (str): filepath to save the invoice to
         """
-        request = GetInvoicePdfRequest(id=invoice_id, org_id=org_id)
-        response: GetInvoicePdfResponse = await self._billing_client.GetInvoicePdf(request, metadata=self._metadata, timeout=timeout)
-        data: bytes = response[0].chunk
-        with open(dest, "wb") as file:
-            file.write(data)
+        stream: Stream[GetInvoicePdfRequest, GetInvoicePdfResponse]
+        async with self._billing_client.GetInvoicePdf.open(timeout=timeout, metadata=self._metadata) as stream:
```
njooma, line 64: type checking surfaced this issue, where we weren't streaming the response properly

pyproject.toml
```diff
@@ -23,29 +23,34 @@ include = ["LICENSE", "src/viam/rpc/libviam_rust_utils.*"]
...
-  sphinx-autoapi = "^2.1.1"
-  sphinx-rtd-theme = "^1.3.0"
+  myst-nb = [
+    {version = "<1.0.0", python = "<3.9"},
+    {version = ">=1.0.0", python = ">=3.9"},
+  ]
+  sphinx-autoapi = [
+    {version = "<3.0.0", python = "<3.9"},
+    {version = ">=3.0.0", python = ">=3.9"},
+  ]
```
njooma, line 39: Updating deps only support >3.9

src/viam/logging.py
```diff
@@ -44,6 +45,8 @@ def getLogger(name: str) -> logging.Logger:
 
 
 def addHandlers(logger: logging.Logger):
+    logger.handlers.clear()
```
njooma, line 47: This is to fix an issue where logs were getting doubled if the level was changed

src/viam/services/slam/client.py
```diff
@@ -41,15 +41,15 @@ async def get_position(self, *, timeout: Optional[float] = None) -> Pose:
         response: GetPositionResponse = await self.client.GetPosition(request, timeout=timeout)
         return response.pose
 
-    async def get_point_cloud_map(self, *, timeout: Optional[float] = None) -> List[GetPointCloudMapResponse]:
+    async def get_point_cloud_map(self, *, timeout: Optional[float] = None) -> List[bytes]:
```
njooma, line 44: type checking surfaced these incorrect return types

src/viam/app/app_client.py
```diff
@@ -1369,10 +1369,11 @@ async def upload_module_file(self, module_file_info: Optional[ModuleFileInfo], f
...
             await stream.send_message(request_module_file_info)
             await stream.send_message(request_file, end=True)
-            response = await stream.recv_message()
-            if not response:
-                await stream.recv_trailing_metadata()  # causes us to throw appropriate gRPC error.
-            return response.url
+            response: Union[UploadModuleFileRequest, None] = await stream.recv_message()
+            if response is not None:
+                return response.url
+            await stream.recv_trailing_metadata()  # causes us to throw appropriate gRPC error.
```
stuqdog, line 1375: It looks like this got inverted from how it was before? we want to `recv_trailing_metadata()` to throw an error, when `response` is `None`.

src/viam/services/vision/client.py
```diff
@@ -127,7 +135,13 @@ async def get_object_point_clouds(
         response: GetObjectPointCloudsResponse = await self.client.GetObjectPointClouds(request, timeout=timeout)
         return list(response.objects)
 
-    async def do_command(self, command: Mapping[str, ValueTypes], *, timeout: Optional[float] = None) -> Mapping[str, ValueTypes]:
+    async def do_command(
+        self,
+        command: Mapping[str, ValueTypes],
+        *,
+        timeout: Optional[float] = None,
+        **__,
```
stuqdog, line 143: (q) Why is it that this is necessary for `do_command` but not, e.g., `get_object_point_clouds` above?

src/viam/__init__.py
```diff
@@ -41,6 +41,22 @@ def _log_exceptions(exctype, value, traceback):
 ##################
 # MONKEY PATCHES #
 ##################
-_ResourceName.__str__ = lambda self: f"{self.namespace}:{self.type}:{self.subtype}/{self.name}"
-_ResourceName.__repr__ = lambda self: f"<viam.proto.common.ResourceName {str(self)} at {hex(id(self))}>"
-_ResourceName.__hash__ = lambda self: hash(str(self))
```
stuqdog, line 46: (q) out of curiosity, what was wrong with the lambda approach?

stuqdog (changes requested):
Nice! A couple questions and one small change, otherwise lgtm!

src/viam/app/app_client.py
njooma: Not sure why I changed it, but isn't this functionally the same?

src/viam/services/vision/client.py
njooma: Because the abstract super method for `get_object_point_clouds` doesn't include `**kwargs` in the input list, whereas `do_command` does

src/viam/__init__.py
njooma: Nothing was wrong with it, but it was not possible to add types to the lambdas. I actually preferred the lambdas because it's so concise, but alas

purplenicole730 (approved):
LGTM. It's cool to see the errors that type checking picked up!

src/viam/app/app_client.py
stuqdog: 🤦 yes you're absolutely right, read it too quickly and thought the `if` statements were equivalent.

I do have a slight preference for either leaving it the way it was or updating the equivalent logic in `data_client.py` and `audo_input/client.py` just to all be consistent, but I don't feel strongly enough to hold off release.

stuqdog (approved):
lgtm!

PR #515, opened 2024-01-04
RSDK-6121: Implement getAccuracy

susmitaSanyal:
getAccuracy API changes. Scop doc: https://docs.google.com/document/d/1dbVI9njfMWXWg8B_wfRU9vCzi6pGQsgdwgqqjMB-Uhc/edit#heading=h.tcicyojyqi6c

getAccuracy now returns: 
map<string, float> accuracy 
float position_hdop
float position_vdop
int position_nmea_gga_fix
float compass_degrees_error


tests/mocks/components.py
```diff
@@ -525,6 +525,7 @@ async def home(self, *, extra: Optional[Dict[str, Any]] = None, timeout: Optiona
         self.homed = True
         self.extra = extra
         self.timeout = timeout
+        print("called home")
```
purplenicole730, line 528: Was this meant to be left in?

tests/test_movement_sensor.py
```diff
@@ -54,7 +54,14 @@
...
     compass_heading_supported=False,
 )
-ACCURACY = {"foo": 0.1, "bar": 2, "baz": 3.14}
+ACCURACY = MovementSensor.Accuracy(
+    accuracy={"foo": 0.1, "bar": 2, "baz": 3.14},
+    position_hdop=0.0,
+    position_vdop=0.0,
+    position_nmea_gga_fix=0,
+    compass_degrees_error=0.0,
+
```
purplenicole730, line 63: nit: erase empty line

purplenicole730 (approved):
LGTM!

tests/mocks/components.py
stuqdog: Yeah we should probably get rid of this, good catch @purplenicole730 !

stuqdog (approved):
Nice, looks good! Worth noting though, the `GetAccuracyResponse` is a pretty simple type (mostly just easily understood primitives). Personally I'm a big fan of creating shadow types for all things proto but this hasn't been a consistent norm in the python SDK. I suspect @njooma would have a preference for not creating the `Accuracy` shadow type because 1) it creates more maintenance work for us, including the possibility of failing to remain accurate if a `GetAccuracyResponse` introduces new fields in the future, and 2) as stated above, the values we're dealing with are all pretty straightforward and easily understood.

Would recommend the two minor fixes that @purplenicole730  caught and getting buy-in from @njooma on the type shadowing, otherwise looks great to me!

njooma (changes requested):
@stuqdog is correct, I'm not a big fan of shadow types. My suggestion would be to create a `TypeAlias` that maps to `GetAccuracyResponse` similar to how we did it for [`Camera.Properties`](https://github.com/viamrobotics/viam-python-sdk/blob/main/src/viam/components/camera/camera.py#L34)

src/viam/components/movement_sensor/service.py
```diff
@@ -114,8 +114,14 @@ async def GetAccuracy(self, stream: Stream[GetAccuracyRequest, GetAccuracyRespon
         name = request.name
         sensor = self.get_resource(name)
         timeout = stream.deadline.time_remaining() if stream.deadline else None
-        accuracy = await sensor.get_accuracy(extra=struct_to_dict(request.extra), timeout=timeout, metadata=stream.metadata)
-        response = GetAccuracyResponse(accuracy=accuracy)
+        accuracy = await sensor.get_accuracy(timeout=timeout, metadata=stream.metadata)
```
njooma, line 117: We should still be sending extra parameters here

src/viam/components/movement_sensor/service.py
```diff
@@ -114,8 +114,14 @@ async def GetAccuracy(self, stream: Stream[GetAccuracyRequest, GetAccuracyRespon
...
+        response = GetAccuracyResponse(
```
njooma, line 118: You shouldn't need to create a `GetAccuracyResponse`, you should be able to use the `accuracy` return value from `await sensor.get_accuracy(...)`

njooma (changes requested):

src/viam/components/movement_sensor/service.py
```diff
@@ -32,7 +32,7 @@
 from viam.utils import dict_to_struct, sensor_readings_native_to_value, struct_to_dict
 
 
-class MovementSensorRPCService(MovementSensorServiceBase, ResourceRPCServiceBase[MovementSensor]):
+class MovementSensorRPCService(MovementSensorServiceBase, ResourceRPCServiceBase):
```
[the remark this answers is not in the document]
susmitaSanyal: I ran lint and it changed it. adding it back.

src/viam/components/movement_sensor/client.py
```diff
@@ -134,12 +133,8 @@ async def get_accuracy(
...
         timeout: Optional[float] = None,
         **__,
-    ) -> Mapping[str, float]:
-        if extra is None:
-            extra = {}
-        request = GetAccuracyRequest(name=self.name, extra=dict_to_struct(extra))
-        response: GetAccuracyResponse = await self.client.GetAccuracy(request, timeout=timeout)
-        return response.accuracy
+    ) -> MovementSensor.Accuracy:
+        return await self.client.GetAccuracy(GetAccuracyRequest(name=self.name), timeout=timeout)
```
njooma, line 137: Sorry, didn't see on my first pass through but we've lost the `extra` param here as well

njooma (changes requested):

njooma (approved):

PR #521, opened 2024-01-16
Fix update-protos workflow

njooma:
The update protos workflow has been failing because of stale info. The issue is, there is no existing branch for the stale info to be coming from (the branch is auto-deleted once the PR is merged). 

I went ahead and changed the flow to use only git actions that we have more control over, rather than a third-party action.


.github/workflows/update_protos.yml
```diff
@@ -8,12 +8,9 @@ on:
 
 jobs:
   update-protos:
-    if: github.repository_owner == 'viamrobotics'
```
stuqdog, line 11: (q) why do we no longer want to check for `repository_owner == 'viamrobotics'`?

stuqdog (approved):

purplenicole730 (approved):

.github/workflows/update_protos.yml
njooma: Ooops that was an accident while I was testing. Added back in

PR #526, opened 2024-01-19
RSDK-6375 Add generic service wrappers

benjirewis:
[RSDK-6375](https://viam.atlassian.net/browse/RSDK-6375)

Adds client and service wrappers for the new generic service protos. Adds tests and tests utilities for generic service.

### DIFF
--- a/src/viam/services/generic/__init__.py
+++ b/src/viam/services/generic/__init__.py
@@ -1,0 +1,18 @@
+import viam.gen.service.generic.v1.generic_pb2  # Need this import for Generic service descriptors to resolve
+from viam.resource.registry import Registry, ResourceRegistration
+
+from .client import GenericClient
+from .generic import Generic
+from .service import GenericRPCService
+
+__all__ = [
+    "Generic",
+]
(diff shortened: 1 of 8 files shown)


tests/test_motor.py
```diff
@@ -156,9 +156,9 @@ async def test_set_power(self, motor: MockMotor, service: MotorRPCService):
         async with ChannelFor([service]) as channel:
             client = MotorServiceStub(channel)
             request = SetPowerRequest(name=motor.name, power_pct=13)
-            await client.SetPower(request, timeout=1.23)
+            await client.SetPower(request, timeout=2.34)
```
benjirewis, line 159: Discussed offline: bumped because of flaky behavior on 3.12.

purplenicole730 (approved):
LGTM! It took me longer than I care to admit to understand why you were adding generic service wrappers when we "already had" generic [component] service wrappers. 😅
Good catch on renaming `MockGeneric`!

njooma (approved):

PR #533, opened 2024-01-31
RSDK-6509 - Delete motion.MoveOnMapNew

nicksanford:
[Ticket](https://viam.atlassian.net/browse/RSDK-6509)


src/viam/proto/service/motion/__init__.py
```diff
@@ -17,8 +17,6 @@
     MotionConfiguration,
     MoveOnGlobeRequest,
     MoveOnGlobeResponse,
-    MoveOnMapNewRequest,
-    MoveOnMapNewResponse,
```
njooma, line 21: This is an autogenerated file, so I hope you didn't do this manually. But given that I don't see a corresponding change in the proto files, I think this was a manual update. 

My suggestion would be to let this file continue to be autogenerated, and delete `MoveOnMapNew` from the other files

purplenicole730 (approved):
Other than Naveed's comment and not changing a proto file manually, LGTM. I checked and it seems `MoveOnMapNew` is not anywhere else.

njooma:
Merged with https://github.com/viamrobotics/viam-python-sdk/pull/527

PR #537, opened 2024-02-16
[RDK-6642] Add examples to Camera component

ale7714:
[RSDK-6642](https://viam.atlassian.net/browse/RSDK-6642) 

Add code samples to the Camera component. The docs team will use this PR as an example on how to move the existing examples in docs.viam to the SDK.

### DIFF
--- a/src/viam/components/camera/camera.py
+++ b/src/viam/components/camera/camera.py
@@ -43,6 +43,17 @@ class Camera(ComponentBase):
         NOTE: If the mime type is ``image/vnd.viam.dep`` you can use :func:`viam.media.video.ViamImage.bytes_to_depth_array`
         to convert the data to a standard representation.
 
+        ::
+
+            my_camera = Camera.from_robot(robot=robot, name="my_camera")
+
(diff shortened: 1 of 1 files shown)


src/viam/components/camera/camera.py
```diff
@@ -43,6 +43,17 @@ async def get_image(
         NOTE: If the mime type is ``image/vnd.viam.dep`` you can use :func:`viam.media.video.ViamImage.bytes_to_depth_array`
         to convert the data to a standard representation.
 
+        ::
+
+            my_camera = Camera.from_robot(robot=robot, name="my_camera")
+
+            # Assume "frame" has a mime_type of "image/vnd.viam.dep"
+            frame = await my_camera.get_image()
```
njooma, line 51: Perhaps add this as the mime type as part of the request? I think that would showcase the method better

```python3
frame = await my_camera.get_image(mime_type = CameraMimeType.VIAM_RAW_DEPTH)
```

njooma (approved):
LGTM, one small comment but not blocking/no need to re-request review if you update

PR #538, opened 2024-02-16
RSDK-5731: add dataset apis

purplenicole730:
[Jira Ticket](https://viam.atlassian.net/browse/RSDK-5731)

Changes:
- Added dataset APIs to /data client (`CreateDataset`, `ListDatasetsByIDs`, `ListDatasetsByOrganizationID`, `RenameDataset`, `DeleteDataset`, `AddBinaryDataToDatasetByIDs`, and `RemoveBinaryDataFromDatasetByIDs`).
- Made a new `Dataset` class in `DataClient` class.
- Made a new `MockDataset` in `mocks/services.py`.
- Created a new `test_dataset.py`.
- Added tests for new DataClient functions.

### DIFF
--- a/src/viam/app/data_client.py
+++ b/src/viam/app/data_client.py
@@ -3,6 +3,7 @@ from datetime import datetime
 from pathlib import Path
 from typing import Any, List, Mapping, Optional, Tuple
 
+from google.protobuf.internal.containers import RepeatedCompositeFieldContainer
 from google.protobuf.struct_pb2 import Struct
 from grpclib.client import Channel, Stream
 
@@ -47,6 +48,7 @@ from viam.proto.app.data import (
 from viam.proto.app.dataset import (
     CreateDatasetRequest,
     CreateDatasetResponse,
+    Dataset,
     DatasetServiceStub,
     DeleteDatasetRequest,
     ListDatasetsByIDsRequest,
@@ -111,6 +113,7 @@ class DataClient:
                 return str(self) == str(other)
             return False
 
+    # TODO (RSDK-6684): Revisit if this shadow type is necessary
     class BinaryData:
         """Class representing a piece of binary data and associated metadata.
 
@@ -493,22 +496,6 @@ class DataClient:
     async def configure_database_user(self, organization_id: str, password: str) -> None:
         raise NotImplementedError()
 
-    class Dataset:
-        """Class representing a dataset.
-
-        Args:
-            id (str): the id of the dataset.
-            name (str): the name of the dataset.
-            organization_id (str): the organization id of the dataset.
-            time_created (datetime): the time the dataset was created.
-        """
-
-        def __init__(self, id: str, name: str, organization_id: str, time_created: datetime) -> None:
-            self.id = id
-            self.name = name
-            self.organization_id = organization_id
-            self.time_created = time_created
-
     async def create_dataset(self, name: str, organization_id: str) -> str:
         """Create a new dataset.
 
@@ -523,7 +510,7 @@ class DataClient:
         response: CreateDatasetResponse = await self._dataset_client.CreateDataset(request, metadata=self._metadata)
         return response.id
 
-    async def list_dataset_by_ids(self, ids: List[str]) -> List[Dataset]:
+    async def list_dataset_by_ids(self, ids: List[str]) -> RepeatedCompositeFieldContainer[Dataset]:
         """Get a list of datasets using their IDs.
 
         Args:
@@ -535,12 +522,9 @@ class DataClient:
         request = ListDatasetsByIDsRequest(ids=ids)
         response: ListDatasetsByIDsResponse = await self._dataset_client.ListDatasetsByIDs(request, metadata=self._metadata)
 
-        datasets = []
-        for dataset in response.datasets:
-            datasets.append(DataClient.Dataset(dataset.id, dataset.name, dataset.organization_id, dataset.time_created.ToDatetime()))
-        return datasets
+        return response.datasets
 
-    async def list_datasets_by_organization_id(self, organization_id: str) -> List[Dataset]:
+    async def list_datasets_by_organization_id(self, organization_id: str) -> RepeatedCompositeFieldContainer[Dataset]:
         """Get the datasets in an organization.
 
         Args:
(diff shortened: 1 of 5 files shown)


src/viam/app/data_client.py
```diff
@@ -478,6 +493,111 @@ async def get_database_connection(self, organization_id: str) -> str:
...
+            name (str): the name of the dataset.
+            organization_id (str): the organization id of the dataset.
+            time_created (datetime): the time the dataset was created.
+        """
+
+        def __init__(self, id: str, name: str, organization_id: str, time_created: datetime) -> None:
+            self.id = id
+            self.name = name
+            self.organization_id = organization_id
+            self.time_created = time_created
```
stuqdog, line 510: Given the simplicity of converting the proto `Dataset` type to this representation, I suspect it's gonna be easier for us to maintain (and more consistent with the rest of the SDK) to do away with this shadow type entirely and just return ask users to interact with raw proto `Dataset`s.

As an aside (certainly not necessary for this PR!) we should probably get rid of the `BinaryData` shadow class in here since there's no conversion at all, we're okay with users interacting with raw proto in the python SDK, and we run the risk of falling out of date if the proto representation ever changes.

src/viam/app/data_client.py
```diff
@@ -478,6 +493,111 @@ async def get_database_connection(self, organization_id: str) -> str:
...
+        """Rename a dataset specified by the dataset ID.
+
+        Args:
+            id (str): The ID of the dataset.
+            name (str): The new name of the dataset.
+        """
+        request = RenameDatasetRequest(id=id, name=name)
+        await self._dataset_client.RenameDataset(request, metadata=self._metadata)
+
+    async def delete_dataset(self, id: str):
```
stuqdog, line 572: (minor) can we add type annotation for the `None` return here, just for consistency's sake?

src/viam/app/data_client.py
```diff
@@ -806,6 +926,7 @@ def create_filter(
         end_time: Optional[datetime] = None,
         tags: Optional[List[str]] = None,
         bbox_labels: Optional[List[str]] = None,
+        dataset_id: Optional[str] = None,
     ) -> Filter:
         warnings.warn("DataClient.create_filter is deprecated. Use AppClient.create_filter instead.", DeprecationWarning, stacklevel=2)
```
stuqdog, line 931: (opt) A bit of a flyby, but can we correct this message? The proper create_filter exists in `utils`, not `AppClient`.

tests/test_dataset.py
```diff
@@ -0,0 +1,63 @@
...
+    @pytest.mark.asyncio
+    async def test_list_datasets_by_ids(self, service: MockDataset):
+        async with ChannelFor([service]) as channel:
+            client = DataClient(channel, DATA_SERVICE_METADATA)
+            datasets = await client.list_dataset_by_ids([ID])
+            assert service.ids == [ID]
+            assert datasets[0].id == DATASETS[0].id
+            assert datasets[0].name == DATASETS[0].name
+            assert datasets[0].organization_id == DATASETS[0].organization_id
+            assert datasets[0].time_created == DATASETS[0].time_created
```
stuqdog, line 52: (q) can this not be simplified down to `assert datasets == DATASETS`? Same with the following test.

stuqdog (changes requested):
Nice, looks good! A couple small comments and a suggestion to remove the shadow type, otherwise this looks great.

src/viam/app/data_client.py
purplenicole730: Got it, will remove the shadow type. Thanks for the explanation!
I'll make a ticket for removing `BinaryData` since I don't want to remove it without investigation.

stuqdog: Sounds good, thanks!

src/viam/app/data_client.py
purplenicole730: Great catch!

tests/test_dataset.py
stuqdog: Ahh, that makes sense. We'd need to implement `def __eq__` for the shadow type in order to allow direct comparison. But since we're using proto it's all good!

src/viam/app/data_client.py
```diff
@@ -3,6 +3,7 @@
 from pathlib import Path
 from typing import Any, List, Mapping, Optional, Tuple
 
+from google.protobuf.internal.containers import RepeatedCompositeFieldContainer
```
stuqdog, line 6: I don't think we want this `import`, see below comments.

src/viam/app/data_client.py
```diff
@@ -523,7 +510,7 @@ async def create_dataset(self, name: str, organization_id: str) -> str:
         response: CreateDatasetResponse = await self._dataset_client.CreateDataset(request, metadata=self._metadata)
         return response.id
 
-    async def list_dataset_by_ids(self, ids: List[str]) -> List[Dataset]:
+    async def list_dataset_by_ids(self, ids: List[str]) -> RepeatedCompositeFieldContainer[Dataset]:
```
stuqdog, line 513: I think we still want to return `List` here and below. We can accomplish that pretty easily by, e.g., `return list(response.datasets)` instead of `return response.datasets)`

stuqdog (approved):
Nice! One change I'd request (go back to `return`ing a list rather than the protobuf equivalent), no need to re-request review after that's done. Otherwise looks great!

src/viam/app/data_client.py
njooma: You could actually return `Sequence` here and no conversion would be necessary 👀

src/viam/app/data_client.py
```diff
@@ -478,6 +495,89 @@ async def get_database_connection(self, organization_id: str) -> str:
...
+        """Delete a dataset.
+
+        Args:
+            id (str): The ID of the dataset.
+        """
+        request = DeleteDatasetRequest(id=id)
+        await self._dataset_client.DeleteDataset(request, metadata=self._metadata)
+
+    async def add_binary_data_to_dataset_by_ids(self, binary_ids: List[BinaryID], dataset_id: str) -> None:
+        """Add VIAM_DATASET_{id} tag to binary data.
```
njooma, line 562: Perhaps for this method and the one below it (remove tag), the docstring should start more high level --

> Add the `BinaryData` to the provided `DataSet`.
>
> This `BinaryData` will be tagged with the `VIAM_DATASET_{id}` label

njooma (approved):
Some nits, but LGTM

PR #540, opened 2024-02-23
DOCS-1871: Update camera, encoder, gantry examples

npentrel:
Python methods are copied from docs.viam.com and should therefore be correct. You do not need to test each method but feel free to scan over them. It is possible there are mistakes.

### DIFF
--- a/src/viam/components/component_base.py
+++ b/src/viam/components/component_base.py
@@ -47,6 +47,14 @@ class ComponentBase(abc.ABC, ResourceBase):
         """
         Get all geometries associated with the Component, in their current configuration, in the frame of the Component.
 
+        ::
+
+            geometries = await component.get_geometries()
+            
+            if geometries:
+                # Get the center of the first geometry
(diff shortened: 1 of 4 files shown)


src/viam/components/component_base.py
```diff
@@ -47,6 +47,14 @@ async def get_geometries(self, *, extra: Optional[Dict[str, Any]] = None, timeou
...
         Get all geometries associated with the Component, in their current configuration, in the frame of the Component.
 
+        ::
+
+            geometries = await component.get_geometries()
+            
+            if geometries:
+                # Get the center of the first geometry
+                print(f"Pose of the first geometry's centerpoint: {geometries[0].center}")
+
```
npentrel, line 57:
```suggestion
        ::

            geometries = await component.get_geometries()

            if geometries:
                # Get the center of the first geometry
                print(f"Pose of the first geometry's centerpoint: {geometries[0].center}")

```

stuqdog (approved):
looks reasonable to me!

purplenicole730 (approved):

PR #541, opened 2024-02-23
DOCS-1870: Add methods for arm, board, and base

npentrel:


npentrel:
@benjirewis - no particular rush on this but just a reminder that this is ready for review. Thank you!

benjirewis (approved):
Apologies for delay, sometimes I ignore these SDK reviews, as I'm not sure how critical my review is anymore 😆 . This LGTM, but I'd wait on @njooma's review.

src/viam/components/base/base.py
```diff
@@ -92,6 +109,34 @@ async def set_power(
...
 
+        ::
+
+            my_base = Base.from_robot(robot=robot, name="my_base")
+
+            # Make your wheeled base move forward. Set linear power to 75%.
+            print("move forward")
+            await my_base.set_power(
+                linear=Vector3(x=0, y=-.75, z=0),
+                angular=Vector3(x=0, y=0, z=0))
```
njooma, line 120: For these, you could pass in an empty `Vector3()` rather than initializing all the values to `0` to make things a bit cleaner, but initializing to `0` might be good for docs as it shows the explicit values

njooma (approved):
Just one comment, but it's nonblocking

src/viam/components/base/base.py
npentrel: I don't have a strong opinion but I do like to show it explicitly for readability as you say

PR #548, opened 2024-03-01
DOCS-1869: Add robot client examples

skyleilani:


src/viam/robot/client.py
```diff
@@ -108,6 +108,18 @@ def with_api_key(cls, api_key: str, api_key_id: str, **kwargs) -> Self:
             """
             Create RobotClient.Options with an API key for credentials and default values for other arguments.
 
+            ::
+
+                # Replace "<API-KEY>" (including brackets) with your machine's API key
+                api_key = '<API-KEY>'
+                # Replace "<API-KEY-ID>" (including brackets) with your machine's API key
+                # ID
```
npentrel, line 116: they have  a longer line limit here so we can move these on the same line
```suggestion
                # Replace "<API-KEY-ID>" (including brackets) with your machine's API key ID
```

src/viam/robot/client.py
```diff
@@ -127,6 +138,23 @@ def with_api_key(cls, api_key: str, api_key_id: str, **kwargs) -> Self:
...
         """Create a robot client that is connected to the robot at the provided address.
 
+        ::
+
+            async def connect():
+            opts = RobotClient.Options.with_api_key(
+                # Replace "<API-KEY>" (including brackets) with your machine's API key
+                api_key='<API-KEY>',
+                # Replace "<API-KEY-ID>" (including brackets) with your machine's
+                # API key ID
```
npentrel, line 148:
```suggestion
                # Replace "<API-KEY-ID>" (including brackets) with your machine's API key ID
```

npentrel (approved):
just two whitespace improvements, otherwise good to go I think, thank you for these!

src/viam/robot/client.py
```diff
@@ -108,6 +108,17 @@ def with_api_key(cls, api_key: str, api_key_id: str, **kwargs) -> Self:
...
+            ::
+
+                # Replace "<API-KEY>" (including brackets) with your machine's API key
+                api_key = '<API-KEY>'
+                # Replace "<API-KEY-ID>" (including brackets) with your machine's API key ID
+                api_key_id = '<API-KEY-ID>'
+
+                opts = RobotClient.Options.with_api_key(api_key, api_key_id)
+
+                robot = await RobotClient.at_address('ADDRESS FROM THE VIAM APP', opts)
```
njooma, line 120: My only comment is that for almost all of these docs, you display user-definable variables as `<SOME-VAR-NAME>`, but for all the addresses you've just done `"ADDRESS"` or `"ADDRESS FROM THE VIAM APP"`. Should those also be in angle brackets (`<>`) for consistency?

njooma (approved):
Just a comment, but it's non-blocking

PR #550, opened 2024-03-01
DOCS-1877: Add motion, nav, slam method examples

skyleilani:
Additional comments added to navigation.py

### DIFF
--- a/src/viam/services/motion/client.py
+++ b/src/viam/services/motion/client.py
@@ -288,7 +288,7 @@ class MotionClient(ServiceClientBase, ReconfigurableResourceRPCClientBase):
 
         ::
 
-            # Assuming a `move_on_globe()` started started the execution
+            # Assuming a `move_on_globe()` started the execution
             # Stop the base component which was instructed to move by `move_on_globe()`
             # or `move_on_map()`
             my_base_resource_name = Base.get_resource_name("my_base")


src/viam/services/motion/client.py
```diff
@@ -121,6 +139,24 @@ async def move_on_globe(
...
+
+            # Get the ResourceNames of the base and movement sensor
+            my_base_resource_name = Base.get_resource_name("my_base")
+            mvmnt_sensor_resource_name = MovementSensor.get_resource_name(
+                "my_movement_sensor")
+            #  Define a destination GeoPoint at the GPS coordinates [0, 0]
+            my_destination = movement_sensor.GeoPoint(latitude=0, longitude=0)
+
+            # Move the base component to the designated geographic location, as reported by
+            # the movement sensor
```
npentrel, line 154: I think the max line length is 120 or 140 so can merge these too:
```suggestion
            # Move the base component to the designated geographic location, as reported by the movement sensor
```

src/viam/services/motion/client.py
```diff
@@ -188,6 +224,24 @@ async def move_on_map(
...
+        ::
+
+            motion = MotionClient.from_robot(robot=robot, name="builtin")
+
+            # Get the ResourceNames of the base and SLAM service
+            my_base_resource_name = Base.get_resource_name("my_base")
+            my_slam_service_name = SLAMClient.get_resource_name("my_slam_service")
+
+            # Define a destination pose with respect to the origin of the map from the SLAM
+            # service "my_slam_service"
```
npentrel, line 236:
```suggestion
            # Define a destination pose with respect to the origin of the map from the SLAM service "my_slam_service"
```

src/viam/services/navigation/navigation.py
```diff
@@ -22,36 +22,202 @@ class Navigation(ServiceBase):
...
+        There are two options for modes: MODE_MANUAL or MODE_WAYPOINT.
+
+            MODE_WAYPOINT: Start to look for added waypoints and begin autonomous navigation.
+            MODE_MANUAL: Stop autonomous navigation between waypoints and allow the base to be controlled manually.
+
+        ::
+            my_nav = NavigationClient.from_robot(robot=robot, name="my_nav_service")
+
+            # Set the Mode the service is operating in to MODE_WAYPOINT and begin
+            # navigation
```
npentrel, line 194:
```suggestion
            # Set the Mode the service is operating in to MODE_WAYPOINT and begin navigation
```

src/viam/services/slam/slam.py
```diff
@@ -22,6 +22,14 @@ class SLAM(ServiceBase):
     async def get_internal_state(self, *, timeout: Optional[float]) -> List[bytes]:
         """Get the internal state of the SLAM algorithm required to continue mapping/localization.
 
+        ::
+
+            slam = SLAMClient.from_robot(robot=robot, name="my_slam_service")
+
+            # Get the internal state of the SLAM algorithm required to continue
+            # mapping/localization.
```
npentrel, line 30:
```suggestion
            # Get the internal state of the SLAM algorithm required to continue mapping/localization.
```

src/viam/services/slam/slam.py
```diff
@@ -44,6 +59,13 @@ async def get_position(self, *, timeout: Optional[float]) -> Pose:
         """
         Get current position of the specified component in the SLAM Map.
 
+        ::
+            slam_svc = SLAMClient.from_robot(robot=robot, name="my_slam_service")
+
+            # Get the current position of the specified source component in the SLAM map as
+            # a Pose.
```
npentrel, line 66:
```suggestion
            # Get the current position of the specified source component in the SLAM map as a Pose.
```

npentrel (commented):
Mostly good - a few formatting issues. Please try building the docs to make sure everything looks right. You can build them with:

```
pip3 install poetry
make install
make documentation
cd docs/_build/html
python3 -m http.server 9000
``

File: src/viam/services/motion/client.py [hunk 6]
```diff
@@ -232,6 +284,24 @@ async def stop_plan(
     ):
         """Stop a component being moved by an in progress ``move_on_globe()`` or ``move_on_map()`` call.
 
+        ::
+
+            motion = MotionClient.from_robot(robot=robot, name="builtin")
+            my_base_resource_name = Base.get_resource_name("my_base")
+
+            # Assuming a move_on_globe started started the execution
```
njooma, line 292: Is it possible to link to the method here, rather than commenting out what could've started this process? It's kind of confusing to have another method's usage in the docs of this method

njooma: Like [down here](https://github.com/viamrobotics/viam-python-sdk/pull/550/files#diff-3f50c1b6e65f9668bfb199b019f5969113233f18cee420aa0ade2142218bc2e2R437-R438), you don't write a full `connect` function

src/viam/services/motion/client.py
```diff
@@ -188,6 +223,23 @@ async def move_on_map(
 
         You can monitor the progress of the ``move_on_map()`` call by querying ``get_plan()`` and ``list_plan_statuses()``.
 
+        ::
+
+            motion = MotionClient.from_robot(robot=robot, name="builtin")
+
+            # Get the ResourceNames of the base and SLAM service
```
stuqdog, line 230: (minor) should we specify "base _component_" since we're specifying SLAM is a a `service`?

src/viam/services/navigation/navigation.py
```diff
@@ -22,36 +22,208 @@ class Navigation(ServiceBase):
...
+            # Get a list containing each path stored by the navigation service
+            paths = await my_nav.get_paths()
+
+        Args:
+            timeout (Optional[float]): An option to set how long to wait (in seconds)
+                before calling a time-out and closing the underlying RPC call.
+
+        Returns:
+            (List[navigation.Path]): An array comprised of Paths, each path being a user-provided destination,
+                or, Waypoint and the set of geopoints that the robot expects to travel through to get there.
```
stuqdog, line 42: Can we rewrite this slightly? I'm a little confused by it. Something like `...each path being a either user-provided destination, or a Waypoint and the set of geopoints...` maybe?

stuqdog (approved):
A couple small things (no need to re-request review from me for them), otherwise LGTM, modulo comments from Naomi and Naveed :)

src/viam/services/motion/client.py
njooma: Not sure if adding an explanation that connect is in code sample is necessary, since *technically* people can use Viam without ever going to the web app

src/viam/services/navigation/navigation.py
stuqdog: Looks great, thanks!

npentrel (approved):

src/viam/services/motion/client.py
njooma: Sorry, but I meant the opposite! I like how you don't include the `connect` function in the example because that function is not part of the example. The assumption below is that the user already has the robot, we're just showing them how to use that specific function.

I would like the same thing to happen here. You have a lot of extra code for getting a movement sensor, setting a destination, and calling `move_on_globe`, when those aren't part of this example. In my opinion, it would be better if you simply linked to that function in the example. Something like:

```python
# Assuming [MotionClient.move_on_globe](LINK TO FUNCTION) started the execution
```

and then remove the rest of the commented-out code of obtaining a movement sensor, setting the destination, and calling the move_on_globe function.

src/viam/services/motion/client.py
```diff
@@ -232,6 +286,14 @@ async def stop_plan(
     ):
         """Stop a component being moved by an in progress ``move_on_globe()`` or ``move_on_map()`` call.
 
+        ::
+
+            # Assuming a `move_on_globe()` started started the execution
```
npentrel, line 291:
```suggestion
            # Assuming a `move_on_globe()` started the execution
```

njooma (approved):

PR #553, opened 2024-03-07
Build arm7l whl

biotinker:
Thanks to how pip does its version management, pip install viam-sdk does not work on an arm7l board. This means that for example, any python modules which require the viam SDK are uninstallable.
arm6l is forward compatible to arm7l. An arm7l board can use the arm6l SDK just fine, it just isn't installable by pip without manually downloading and renaming.

This PR repackages the arm6l buld as arm7l.

https://viaminc.slack.com/archives/C039G724TKP/p1709829556782589

### DIFF
--- a/.github/workflows/release-candidate.yml
+++ b/.github/workflows/release-candidate.yml
@@ -131,6 +131,9 @@ jobs:
           - arch: linux_armv6l
             ext: so
             whl: linux_armv6l
+          - arch: linux_armv6l
+            ext: so
+            whl: linux_armv7l
     steps:
       - name: Checkout Code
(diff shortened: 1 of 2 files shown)


njooma (approved):
This just works? Like armv6l and armv7l are simpatico?

biotinker:
Arm6l is forwards compatible with arm7l. Anything built for arm6l will run fine on an arm7l cpu as-is.

Hypothetically there are some efficiency gains that could be had if we create a specialized arm7l build, but absent doing the work to do that this allows us to use the SDK on arm7l boards.

PR #556, opened 2024-03-08
[RSDK-6871] add edited map to slam client

JohnN193:
ticket: https://viam.atlassian.net/browse/RSDK-6871

this pr adds the ReturnEditedMap bool to the slam typescript client. Will not merge until after the API PR is merged and this PR is updated accordingly
api change PR: https://github.com/viamrobotics/api/pull/456


Makefile
```diff
@@ -22,7 +22,7 @@ typecheck:
 
 _buf: clean
 	rm -rf src/viam/gen
-	buf generate buf.build/viamrobotics/api
+	buf generate https://github.com/JohnN193/api.git\#branch=edit-maps
```
JohnN193, line 25: will revert before merging after api PR is in

njooma (changes requested):
Similar to the TS SDK PR, I would add this param to the test suite

JohnN193:
Added a test for the parameter!

src/viam/services/slam/slam.py
```diff
@@ -36,7 +36,7 @@ async def get_internal_state(self, *, timeout: Optional[float]) -> List[bytes]:
         ...
 
     @abc.abstractmethod
-    async def get_point_cloud_map(self, *, timeout: Optional[float]) -> List[bytes]:
+    async def get_point_cloud_map(self, return_edited_map: Optional[bool] = None, *, timeout: Optional[float]) -> List[bytes]:
```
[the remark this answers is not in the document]
JohnN193: `None` should be treated the same way as `False` so I'll switch to that! The api from the scope had it listed as optional so I thought I would have to include that here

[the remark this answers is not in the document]
JohnN193: updated, lmk if you meant something else by "updating the docstring"

src/viam/services/slam/client.py
```diff
@@ -38,8 +38,8 @@ async def get_position(self, *, timeout: Optional[float] = None) -> Pose:
         response: GetPositionResponse = await self.client.GetPosition(request, timeout=timeout)
         return response.pose
 
-    async def get_point_cloud_map(self, *, timeout: Optional[float] = None) -> List[bytes]:
-        request = GetPointCloudMapRequest(name=self.name)
+    async def get_point_cloud_map(self, return_edited_map: Optional[bool] = None, *, timeout: Optional[float] = None) -> List[bytes]:
```
njooma, line 41: This should also be just a `bool` defaulted to `False`, rather than an `Optional[bool] = None` (similar to the change you made in `slam.py`)

njooma (changes requested):

src/viam/services/slam/client.py
JohnN193: ah my bad, got the one in the test but I forgot this one

src/viam/services/slam/slam.py
```diff
@@ -47,6 +47,12 @@ async def get_point_cloud_map(self, return_edited_map: Optional[bool] = None, *,
             # Get the point cloud map in standard PCD format.
             pcd_map = await slam_svc.get_point_cloud_map()
 
+        Args:
+            timeout (Optional[float]): An option to set how long to wait (in seconds)
+                before calling a time-out and closing the underlying RPC call.
```
stuqdog, line 52: We can get rid of the `timeout` docstring, just the `return_edited_map` is fine :)

stuqdog (approved):
Don't think we need (or want) the docstring for the `timeout` arg, otherwise lgtm!

njooma (approved):

PR #557, opened 2024-03-08
RSDK-6184: better module logs over gRPC

purplenicole730:
[Jira ticket](https://viam.atlassian.net/browse/RSDK-6184)

Introducing a new module handler class to handle logs when there is a module. When connecting to a `RobotClient`, the module class will call `viam.logging`'s new `setParent()` function, which starts module logging.

Logs then will go from looking like:
```
2024-03-04T12:08:02.742-0500	INFO	robot_server.process.my-module_/Users/nicolejung/viam/viam-python-sdk/examples/complex_module/run.sh.StdOut	pexec/managed_process.go:244	
\_ 2024-03-04 12:08:02,742		INFO	src.gizmo.my_gizmo (my_gizmo.py:79)	gizmo1 is closed.
```
to:
```
2024-03-12T12:59:16.238-0400	INFO	robot_server.my_gizmo	my_gizmo.py:79	gizmo1 is closed.
```

Unideal:
1. When a module log fails, the resulting stdout/stderr log is now from `ModuleLogger` instead.
_We have established that this edge case is okay as the original filename/lineno will still be in the message, and there is a prior error log explaining that module logging failed_
2. How should we test this? `Log` doesn't return anything to check against.
_We decided not to test._


src/viam/logging.py
```diff
@@ -1,11 +1,43 @@
...
+        try:
+            assert self.parent is not None
+            asyncio.create_task(
+                self.parent.log(record.name, record.levelname, time, message, stack), name=f"{viam._TASK_PREFIX}-LOG-{time}"
+            )
+        except Exception as err:
+            self.logger.handlers.clear()
+            addHandlers(self.logger, True)
+            self.logger.error(f"Module handler failed for {record.name} - {err}")
+            self.logger.log(record.levelno, message)
```
purplenicole730, line 39: I'm not sure how to make the logs appear like they're still from the original source. For example:
```
\_ 2024-03-08 17:16:11,222		INFO	src.gizmo.my_gizmo (logging.py:41)	my_gizmo.py:79	gizmo1 is closed.
```
Here, it says that the log is from `logging.py:41`. For now, I made it so that the message includes the filename and number it's from

src/viam/logging.py
```diff
@@ -1,11 +1,43 @@
...
+    def __init__(self, logger: logging.Logger):
+        self.logger = logger
+        self.parent = _MODULE_PARENT
+
+        logger.handlers.clear()
+        super().__init__()
+
+    def emit(self, record: LogRecord):
+        assert isinstance(record, logging.LogRecord)
+        stack = f"exc_info: {record.exc_info}, exc_text: {record.exc_text}, stack_info: {record.stack_info}"
```
benjirewis, line 26: Nice! These won't show up in the RDK yet but they will soon.

src/viam/logging.py
```diff
@@ -1,11 +1,43 @@
...
+        message = f"{record.filename}:{record.lineno}\t{record.msg}"
+        time = datetime.fromtimestamp(record.created)
+
+        try:
+            assert self.parent is not None
+            asyncio.create_task(
```
benjirewis, line 32: I don't have a great sense of `asyncio` semantics. Is it possible to ever "leak" this task in the event that `self.parent.log` hangs forever?

src/viam/logging.py
```diff
@@ -1,11 +1,43 @@
...
+                self.parent.log(record.name, record.levelname, time, message, stack), name=f"{viam._TASK_PREFIX}-LOG-{time}"
+            )
+        except Exception as err:
+            self.logger.handlers.clear()
+            addHandlers(self.logger, True)
```
benjirewis, line 37: Why are we clearing and re-adding handlers in the event of a failure to log?

src/viam/logging.py
benjirewis: This log already looks solid when you test it with the RDK! I just ran the RDK with the complex module example code and a `custom_run.sh` as we discussed offline; follow up with me if you want to test it the same way I did.

```
2024-03-10T22:26:43.009-0400	INFO	robot_server.src.gizmo.my_gizmo	my_gizmo.py:83	gizmo1 is closed.	{"log_ts": "2024-03-10T22:26:43.007Z"}
```

src/viam/logging.py
```diff
@@ -53,13 +85,34 @@ def addHandlers(logger: logging.Logger):
...
+    format = ColorFormatter("%(asctime)s\t\t" + "%(levelname)s\t" + "%(name)s (%(filename)s:%(lineno)d)\t" + "%(message)s\t")
+    handler.setFormatter(format)
+    handler.setLevel(LOG_LEVEL)
+
+    logger.addHandler(handler)
+
+
+def setParent(parent):
+    global _MODULE_PARENT
+    _MODULE_PARENT = parent
```
benjirewis, line 113: Nice! I like the use of the global and just applying it to all existing loggers. What type are `parent` and `_MODULE_PARENT`/should we add type annotations for them?

src/viam/robot/client.py
```diff
@@ -753,6 +755,28 @@ async def stop_all(self, extra: Dict[ResourceName, Dict[str, Any]] = {}):
...
+        Args:
+            name (str): The logger's name.
+            level (str): The level of the log.
+            time (str): The log creation time.
+            log (str): The log message.
+            caller (Dict[str, int]): The filename and number when the log was created.
+            stack_info (str): The stack information of the log.
+            fields (Dict[str, ValueTypes]): The extra fields of the log
+        """
+        entry = LogEntry(level=level, time=datetime_to_timestamp(time), logger_name=name, message=log, stack=stack)
```
benjirewis, line 776: This is almost perfect; I worry about the value of `logger_name`. 

Right now, for a log from `my_gizmo.py`, `logger_name` here ends up being `src.gizmo.my_gizmo`. The eventual RDK log looks like: `2024-03-10T22:26:43.009-0400	INFO	robot_server.src.gizmo.my_gizmo`. As you can see, we use the `logger_name` on the RDK side as a suffix to `robot_server` to give the user an idea of where the log came from. In Go, I used the name of the module; that would be something like `"complex_module"` in this case. Is there any way to set `logger_name` to the name of the module here? That way, we could get an eventual RDK log that looks like: `2024-03-10T22:26:43.009-0400	INFO	robot_server.complex_module`.

benjirewis (commented):
Pretty much perfect! Tested your branch out locally and module logging seems to be working 🎉. Just some questions and nits. I'm not sure if you'll want local Python unit tests in `test_module.py`; that's up to you and @njooma I think.

src/viam/logging.py
purplenicole730: I realized that the clear is redundant since `addHandlers` clears them for you. But if module logging errors, then we want the logger to have the regular handlers so that it can log using stderr and stdout. I wanted to readd the `ModuleHandler` after that so that the logger doesn't just never try module logging again.

src/viam/logging.py
purplenicole730: Yeah, I was struggling with this too. The type here is a `RobotClient`. But I can't import it because the class also imports `logging`. Thanks for reminding me to flag this for further discussion!

src/viam/robot/client.py
purplenicole730: I'll try!

src/viam/logging.py
benjirewis: Ah gotcha that makes sense; thanks for context! Can we leave a comment here explaining that and remove the redundant `clear()` call?

src/viam/logging.py
purplenicole730: The log you see is if the log succeeded. But when it fails, we still need to log, which is what this `except` is doing. The log line I posted above was while testing that failed logs still show up using `stdout` and `stderr`. But because it's calling the log from `logging.py`, it comes up as `logging.py:41` as the filename and line number

src/viam/logging.py
```diff
@@ -53,13 +85,34 @@ def addHandlers(logger: logging.Logger):
...
+    handler = ModuleHandler(logger)
+
+    format = ColorFormatter("%(asctime)s\t\t" + "%(levelname)s\t" + "%(name)s (%(filename)s:%(lineno)d)\t" + "%(message)s\t")
+    handler.setFormatter(format)
+    handler.setLevel(LOG_LEVEL)
+
+    logger.addHandler(handler)
+
+
+def setParent(parent):
```
purplenicole730, line 111: For this line and when we initialize (`_MODULE_PARENT = None`), parent is of type `RobotClient`. The problem with clarifying the type is that importing `RobotClient` introduces a circular import error, as `RobotClient` imports `viam.logging`.
Even if we put module handler in its own file, `logging.py` would import from `modulehandler`, which imports `RobotClient`, which imports `logging.py`, and etc.

src/viam/robot/client.py
purplenicole730: Modules don't have `name` in Python. How do you do it in Go so that the module and/or loggers know the module name?

benjirewis: Discussed offline; @dgottlieb how do you feel about Python module logs showing up as `2024-03-10T22:26:43.009-0400	INFO	robot_server.src.gizmo.my_gizmo`? Modules aren't really aware of their own name. In Go, we knew the module name because the module author _named_ the logger and we encouraged them to use the name of their module.

src/viam/logging.py
benjirewis: Gotcha... Not annotating seems fine; also just wanted to clarify what that type was for my own understanding. Thank you!

src/viam/logging.py
benjirewis: Gotcha... Let me think about that one. I think it's definitely an edge case that we fail to log over gRPC once gRPC logging has started, so unclear how "bad" it is that we report two different line numbers.

src/viam/robot/client.py
dgottlieb: I agree this is an interesting quirk. I'm fine with how this is right now.

If you _really_ wanted to try something (that may have different negative consequences), you could try trimming `src.gizmo.my_gizmo` down to just the bit after the last dot, i.e: `my_gizmo`.

src/viam/module/module.py
```diff
@@ -132,6 +134,7 @@ async def add_resource(self, request: AddResourceRequest):
         model = Model.from_string(config.model, ignore_errors=True)
         creator = Registry.lookup_resource_creator(subtype, model)
         resource = creator(config, dependencies)
+        logging.getLogger(request.config.name)
```
njooma, line 137: What is this for?

src/viam/logging.py
```diff
@@ -1,11 +1,43 @@
...
+import viam
+
 LOG_LEVEL = INFO
 LOGGERS: Dict[str, logging.Logger] = {}
+_MODULE_PARENT = None
+
+
+class ModuleHandler(logging.Handler):
+    def __init__(self, logger: logging.Logger):
+        self.logger = logger
```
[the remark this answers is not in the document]
njooma: I would create a new default logger for when module logging fails, rather than passing and saving the logger

src/viam/logging.py
njooma: theoretically i think. we should be able to wrap it in a `wait_for`:

```python3
coro = asyncio.wait_for(self._parent.log(record.name, record.levelname, time, message, stack), 10)
asyncio.create_task(coro, name=f"{viam._TASK_PREFIX}-LOG-{record.created}")
```

src/viam/logging.py
njooma: You can actually do it by doing the following

```python3
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    from .robot.client import RobotClient

def setParent(parent: "RobotClient"):
    ...
```

src/viam/logging.py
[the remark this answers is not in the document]
njooma: Also, please annotate the types of `parent` and `logger`

src/viam/robot/client.py
benjirewis: Agree with @dgottlieb (and I think you suggested that option offline to me already @purplenicole730); if it's non-trivial to trim then I think what you have now should work!

src/viam/module/module.py
purplenicole730: Thanks for catching, it's an artifact from testing and trying something else.

src/viam/logging.py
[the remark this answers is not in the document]
purplenicole730: This was my first instinct as well, and creating a `self._backup_logger` (there might be some old commits showing this). But then I started thinking about how if I created this, then calling a handler function that touches all loggers would then start making backups of the backup loggers, as well as changing the handlers of the backup. I'm actually happy to start thinking about this again, as maybe the problem lied in the execution.

src/viam/logging.py
```diff
@@ -1,11 +1,43 @@
...
-from logging import DEBUG, ERROR, FATAL, INFO, WARN, WARNING  # noqa: F401
+from datetime import datetime
+from logging import DEBUG, ERROR, FATAL, INFO, WARN, WARNING, LogRecord  # noqa: F401
 from typing import Dict
 
+import viam
+
 LOG_LEVEL = INFO
 LOGGERS: Dict[str, logging.Logger] = {}
+_MODULE_PARENT = None
```
stuqdog, line 13: (minor) should we type-annotate this? It takes a bit of digging around to really understand what `_MODULE_PARENT` actually _is_.

src/viam/logging.py
```diff
@@ -43,7 +75,7 @@ def getLogger(name: str) -> logging.Logger:
     return logger
 
 
-def addHandlers(logger: logging.Logger):
+def addHandlers(logger: logging.Logger, default=False):
```
stuqdog, line 78: `default` is pretty ambiguous here, I'm not sure what we mean by it exactly. Can we we rename to `use_default_handler` or something like that?

src/viam/logging.py
stuqdog: update: I see this was discussed below! If @njooma's suggestion works here then would love to see annotation; if not, no worries!

stuqdog (approved):
A couple minor things, otherwise lgtm (modulo others' comments)!

src/viam/robot/client.py
purplenicole730: Modules now look like this:
```
2024-03-12T12:59:16.238-0400	INFO	robot_server.my_gizmo	my_gizmo.py:79	gizmo1 is closed.
```

src/viam/robot/client.py
```diff
@@ -753,6 +755,28 @@ async def stop_all(self, extra: Dict[ResourceName, Dict[str, Any]] = {}):
...
+        Create a LogEntry object from the log to send to RDK.
+
+        Args:
+            name (str): The logger's name.
+            level (str): The level of the log.
+            time (str): The log creation time.
+            log (str): The log message.
+            caller (Dict[str, int]): The filename and number when the log was created.
+            stack_info (str): The stack information of the log.
+            fields (Dict[str, ValueTypes]): The extra fields of the log
```
njooma, line 774: These args need to be updated to reflect the current method signature

njooma (approved):
Almost!

src/viam/logging.py
```diff
@@ -43,31 +79,52 @@ def getLogger(name: str) -> logging.Logger:
...
+
+def _addHandlers(loggers: Iterable[logging.Logger], use_default_handlers=False):
+    format = _ColorFormatter("%(asctime)s\t\t" + "%(levelname)s\t" + "%(name)s (%(filename)s:%(lineno)d)\t" + "%(message)s\t")
+
+    handlers: List[logging.Handler] = []
+
+    if _MODULE_PARENT is not None and not use_default_handlers:
+        mod_handler = _ModuleHandler(_MODULE_PARENT)
+        mod_handler.setFormatter(format)
+        mod_handler.setLevel(LOG_LEVEL)
```
benjirewis, line 94: If I run the Python complex module example with your changes with `-debug` on the rdk (`go run web/cmd/server/main.go -config ../viam-configs/python_module.json -debug`), I get a _lot_ of heartbeat logs from the `RobotClient` logger. Also, when I shut down the RDK in that situation, the Python module process seems to raise some error: `grpclib.exceptions.StreamTerminatedError: Connection lost`. Any idea what's going on there?

benjirewis (commented):
Code changes LGTM 🎉 ! Just some concerns about `-debug`.

src/viam/logging.py
purplenicole730: I don't see why the `RobotClient` logger would have _more_ heartbeat logs than it used to. They should be the same amount as before (assuming that those heartbeat logs were being printed before).
If anything, these logs would be easier to read now that they're not surrounded by all the stdout/stderr lines

purplenicole730: Update: these heartbeat logs were NOT being printed out before.

benjirewis (approved):
Awesome! Errors from running with `-debug` seem to be gone. There are still quite a few heartbeat logs, so I'm curious if anything comes from your investigation into the frequency of those (logged by multiple loggers?). Otherwise, LGTM 🎉

dgottlieb (approved):
Just in case this was waiting on me

benjirewis (approved):
LGTM once again!

PR #559, opened 2024-03-12
RSDK-6885 - Add fields to GetPropertiesResponse

kkufieta:
Ticket: https://viam.atlassian.net/browse/RSDK-6885

API PR: https://github.com/viamrobotics/api/pull/458

Done:
* Run `make format`
* Add fields to `GetPropertiesResponse`:
  * `internal_state_file_type`
  * `sensor_info`


src/viam/services/slam/client.py
```diff
@@ -48,10 +48,15 @@ async def get_internal_state(self, *, timeout: Optional[float] = None) -> List[b
...
 
-    async def get_properties(self, *, timeout: Optional[float] = None) -> Tuple[bool, MappingMode.ValueType]:
+    async def get_properties(self, *, timeout: Optional[float] = None) -> Tuple[bool, MappingMode.ValueType, str, List[SensorInfo]]:
         request = GetPropertiesRequest(name=self.name)
         response: GetPropertiesResponse = await self.client.GetProperties(request, timeout=timeout)
-        return (response.cloud_slam, response.mapping_mode)
+
+        sensor_info = []
+        for si in response.sensor_info:
+            sensor_info.append(SensorInfo(name=si.name, type=si.type))
```
njooma, line 57: Is `response.sensor_info` not already a sequence of `SensorInfo`? If so, you don't need to recast it, you could simply do `list(response.sensor_info)`

Makefile
```diff
@@ -22,7 +22,7 @@ typecheck:
 
 _buf: clean
 	rm -rf src/viam/gen
-	buf generate buf.build/viamrobotics/api
+	buf generate https://github.com/kkufieta/api.git\#branch=kk/rsdk-6885-change-slam-proto
```
kkufieta, line 25: WIll update once API PR is merged

src/viam/services/slam/slam.py
```diff
@@ -83,7 +83,8 @@ async def get_properties(self, *, timeout: Optional[float]) -> Tuple[bool, Mappi
             slam_properties = await slam_svc.get_properties()
 
         Returns:
-            Tuple[bool, MappingMode.ValueType]: A tuple of a boolean value representing if the SLAM session is being run in
-            the cloud and the mapping mode of said session
+            Tuple[bool, MappingMode.ValueType, str, List[SensorInfo]]: A tuple containing: A boolean value representing if the SLAM session
+            is being run in the cloud, the mapping mode of said session, a string with the internal state file type of the SLAM algorithm,
+            and a list containing information about the sensors used for the SLAM session.
         """
```
purplenicole730, line 89: There doesn't seem to be a standard with how we introduce tuples in doc strings, but maybe it would be better to use this format used in [arm's get_kinematics](https://github.com/viamrobotics/viam-python-sdk/blob/e4dd87b490782f07ce6c3dc7b159d11aaed60333/src/viam/components/arm/arm.py#L194). Not a necessity, but wanted to provide an alternative in case there was a preference.

purplenicole730 (approved):
Agree with Naveed's comment, but otherwise, LGTM!

src/viam/services/slam/slam.py
kkufieta: Updated it! Let me know what you think.

src/viam/services/slam/slam.py
```diff
@@ -71,7 +71,7 @@ async def get_position(self, *, timeout: Optional[float]) -> Pose:
         ...
 
     @abc.abstractmethod
-    async def get_properties(self, *, timeout: Optional[float]) -> Tuple[bool, MappingMode.ValueType]:
+    async def get_properties(self, *, timeout: Optional[float]) -> Tuple[bool, MappingMode.ValueType, str, List[SensorInfo]]:
```
njooma, line 74: Ok so this is technically a breaking change, since anyone who is unpacking the tuple will now get an error

```python3
x, y = await slam.get_properties().  # ValueError: too many values to unpack (expected 2)
```
Also, this tuple is getting kinda long. Since we're basically just returning the `GetPropertiesResponse` but destructured, I think we should actually use a `TypeAlias` here and simply return the response as is. We do that already in a few places (see [camera](https://github.com/viamrobotics/viam-python-sdk/blob/9ec4ab4993101cfe793d239f2003b1c49a15982a/src/viam/components/camera/camera.py#L38)).

It could look something like:

```python3
# Other imports
# ...

if sys.version_info >= (3, 10):
    from typing import TypeAlias
else:
    from typing_extensions import TypeAlias


class SLAM(ServiceBase):
    ...

    Properties: "TypeAlias" = GetPropertiesResponse

    ...

    async def get_properties(self, *, timeout: Optional[float]) -> SLAM.Properties:
         ...
```

Then, in the client, we can simply return the response as is in both the client and service.

njooma (changes requested):

src/viam/services/slam/slam.py
kkufieta: Done - let me know if it looks correct!

tests/mocks/services.py
```diff
@@ -621,9 +626,9 @@ async def get_position(self, *, timeout: Optional[float] = None) -> Pose:
         self.timeout = timeout
         return self.POSITION
 
-    async def get_properties(self, *, timeout: Optional[float] = None) -> Tuple[bool, MappingMode.ValueType]:
+    async def get_properties(self, *, timeout: Optional[float] = None) -> Tuple[bool, MappingMode.ValueType, str, List[SensorInfo]]:
```
njooma, line 629: These tests need to be updated for the new return type

tests/test_slam.py
```diff
@@ -49,9 +49,11 @@ async def test_do(self):
 
     @pytest.mark.asyncio
     async def test_get_properties(self):
-        (cloud_slam, mapping_mode) = await self.slam.get_properties()
+        (cloud_slam, mapping_mode, internal_state_file_type, sensor_info) = await self.slam.get_properties()
         assert cloud_slam == MockSLAM.CLOUD_SLAM
         assert mapping_mode == MockSLAM.MAPPING_MODE
+        assert internal_state_file_type == MockSLAM.INTERNAL_STATE_FILE_TYPE
+        assert sensor_info == MockSLAM.SENSOR_INFO
```
njooma, line 56: These tests need to be updated for the new return type

njooma (changes requested):

njooma (approved):

PR #567, opened 2024-03-27
RSDK-4089: use viam image

purplenicole730:


src/viam/media/utils.py
```diff
@@ -0,0 +1,74 @@
...
+    Alternatively, image dimensions can be set manually as well (i.e.: image.width = 100)
+
+    Args:
+        image (ViamImage): The image to get dimensions of.
+
+    Raises:
+        NotSupportedError: Raised if given an image that is not of MIME type `image/vnd.viam.dep`.
+    """
+    if image.mime_type not in [CameraMimeType.JPEG, CameraMimeType.PNG, CameraMimeType.VIAM_RGBA]:
+        NotSupportedError("Type must be `image/jpeg`, `image/png`, or `image/vnd.viam.rgba` to use get_image_dimensions()")
```
stuqdog, line 70: do we need to `raise` this error?

src/viam/media/utils.py
```diff
@@ -0,0 +1,74 @@
...
+    image.width = width
+    image.height = height
+    depth_arr = array("H", image.data[24:])
+    depth_arr.byteswap()
+
+    depth_arr_2d = [[depth_arr[row * width + col] for col in range(width)] for row in range(height)]
+    return depth_arr_2d
+
+
+def get_image_dimensions(image: ViamImage) -> None:
```
stuqdog, line 56: (minor) When I see a method named `get_foo` I typically expect that it will return a `foo`, I think it's a bit strange that this returns `None`. But it'd be weird to name it `set_image_dimensions` too because it's taking no `image_dimension` arguments! Can we rename to `determine_image_dimensions` or something like that?

stuqdog (commented):
Nice, a couple small things but otherwise looks good to me!

I'm leaving this un-approved for now however because I believe (based on earlier conversation with @njooma) that we don't want to merge this into `main` but rather into a separate release branch 3 or 4 cycles from now, so that we give people plenty of time to prepare for the breaking change.

src/viam/media/utils.py
purplenicole730: Yes, this is a very good point!

src/viam/media/utils.py
purplenicole730: I wrote this whole paragraph why I thought having this error was necessary. And then realized you quite literally meant "hey, you forgot a `raise` keyword". 🤦

purplenicole730: Thank you for reminding me I should be writing tests to make sure an error is raised if someone tries to call this function with the wrong mime type!

src/viam/media/video.py
```diff
@@ -159,10 +56,10 @@ class ViamImage:
     Provides the raw data and the mime type, as well as lazily loading and caching the PIL.Image representation.
```
njooma, line 56: This comment needs to be updated to remove the `PIL` reference

src/viam/media/utils.py
```diff
@@ -0,0 +1,75 @@
+from array import array
```
njooma, line 1: This file I think should be called `pil.py` within a `utils` directory IMO. Since this file only contains utils for PIL

src/viam/media/utils.py
```diff
@@ -0,0 +1,75 @@
...
+from viam.errors import NotSupportedError
+from viam.media.video import CameraMimeType, ViamImage
+
+from .viam_rgba_plugin import RGBA_FORMAT_LABEL
+
+# Formats that are supported by PIL
+LIBRARY_SUPPORTED_FORMATS = ["JPEG", "PNG", RGBA_FORMAT_LABEL]
+
+
+def viam_to_pil_image(image: ViamImage) -> Image.Image:
```
njooma, line 16: We should have a reverse `pil_to_viam_image` function as well

src/viam/media/utils.py
```diff
@@ -0,0 +1,75 @@
...
+    image.width = width
+    image.height = height
+    depth_arr = array("H", image.data[24:])
+    depth_arr.byteswap()
+
+    depth_arr_2d = [[depth_arr[row * width + col] for col in range(width)] for row in range(height)]
+    return depth_arr_2d
+
+
+def determine_image_dimensions(image: ViamImage) -> None:
```
njooma, line 56: This should not be a public function. Setting/getting the dimensions of a `ViamImage` should not be left to the user, but done internally.

src/viam/media/video.py
```diff
@@ -183,50 +80,24 @@ def mime_type(self, value: CameraMimeType):
         if value == self.mime_type:
```
njooma, line 80: We don't need this setter anymore

src/viam/media/video.py
```diff
@@ -159,10 +56,10 @@ class ViamImage:
     Provides the raw data and the mime type, as well as lazily loading and caching the PIL.Image representation.
     """
 
+    _height: Optional[int]
+    _width: Optional[int]
     _data: bytes
     _mime_type: CameraMimeType
-    _image: Optional[Image.Image] = None
-    _image_decoded = False
```
njooma, line 165: We actually should probably keep these around.

We should attempt to have `width` and `height` available without the user calling `determine_image_dimension`. If we try to set these on `init`, then we actually move backwards in performance (since determining dimensions requires the use of PIL right now). So in order to keep our performance in line with what it used to be, we should only convert to PIL to get dimensions ONCE, and then keep that PIL.Image around so we don't have to always to the decoding.

src/viam/media/video.py
```diff
@@ -183,50 +80,24 @@ def mime_type(self, value: CameraMimeType):
...
-
-    def bytes_to_depth_array(self) -> List[List[int]]:
-        """Decode the data of an image that has the custom depth MIME type ``image/vnd.viam.dep`` into
-        a standard representation.
-
-        Raises:
-            NotSupportedError: Raised if given an image that is not of MIME type `image/vnd.viam.dep`.
+    def width(self) -> Union[int, None]:
+        """The width of the image"""
+        return self._width
```
njooma, line 87: Continuing from my earlier comment around keeping the `_image` and `_image_decoded` properties, the `width` and `height` getters should now be something like:

```python3
# First, check if we already have a value
if self._width is not None:
    return self._width
# If we don't have a value, check to see if we have already tried to decode the image
# If we have not, then try to decode the image now
if not self._image_decoded:
    try:
        self._image = Image.open(BytesIO(self.data), formats=LIBRARY_SUPPORTED_FORMATS)
    except UnidentifiedImageError:
        self._image = None
    self._image_decoded = True
# If we have decoded the image and the image is not none, then set the width so we don't have to do this again
if self._image_decoded and self._image is not None:
    self._width = self._image.width
return self._width
```

src/viam/services/vision/vision.py
```diff
@@ -75,12 +76,18 @@ async def get_detections(
             # Get an image from the camera
             img = await cam1.get_image()
 
+            # Get image dimensions
+            determine_image_dimensions(img)
+
```
njooma, line 81: Remove this -- the user should never be calling this

src/viam/services/vision/vision.py
```diff
@@ -141,6 +148,9 @@ async def get_classifications(
...
```
njooma, line 153: Same as above -- the user should never be calling this

tests/test_media.py
```diff
@@ -2,90 +2,119 @@
...
-        assert img._image_decoded is False
-        assert img.image is not None
-        assert img.image.tobytes() == i.tobytes()
-        assert img._image_decoded is True
-        img.close()
+        assert img._mime_type == CameraMimeType.PNG
+        pil_img = viam_to_pil_image(img)
+        assert pil_img.tobytes() == i.tobytes()
 
     def test_mime_type_update(self):
```
njooma, line 23: We shouldn't allow users to update mime types anymore

tests/test_media.py
```diff
@@ -2,90 +2,119 @@
...
+        img3.width
+    with pytest.raises(AttributeError):
+        img3.height
+    determine_image_dimensions(img3)
+    assert img3.width == 100
+    assert img3.height == 100
+
+    img4 = ViamImage(b"data", CameraMimeType.PCD)
+    with pytest.raises(NotSupportedError):
+        determine_image_dimensions(img4)
```
njooma, line 120: You should at MimeType tests to make sure that when you pass in mime types we support and ones we don't, that everything still works (e.g. that `MimeType.from_string("SOME_RANDOM_STRING")` doesn't error)

njooma (changes requested):
Great work so far! This is looking good

tests/test_media.py
purplenicole730: Would there be a `ViamImage` with an unsupported mime type? We purposely removed `CameraMimeType.UNSUPPORTED`, and `ViamImage` takes a `CameraMimeType` object.

src/viam/media/utils.py
purplenicole730: Are you suggesting to make a `src/viam/utils/` directory with `pil.py` in it? Would this directory also include the already existing`src/viam/utils.py`? That would be a bigger (and breaking) change as well, since so many places call from it.
I realized `src/viam/utils.py` has everything with no real organization; should I just put the contents of `pil.py` in there?

tests/test_media.py
njooma: All good

src/viam/components/camera/camera.py
```diff
@@ -40,15 +37,14 @@ class Camera(ComponentBase):
...
     async def get_image(
         self, mime_type: str = "", *, extra: Optional[Dict[str, Any]] = None, timeout: Optional[float] = None, **kwargs
-    ) -> Union[Image, RawImage]:
-        """Get the next image from the camera as an Image or RawImage.
+    ) -> ViamImage:
+        """Get the next image from the camera as a ViamImage.
         Be sure to close the image when finished.
 
-        NOTE: If the mime type is ``image/vnd.viam.dep`` you can use :func:`viam.media.video.ViamImage.bytes_to_depth_array`
+        NOTE: If the mime type is ``image/vnd.viam.dep`` you can use :func:`viam.media.utils.bytes_to_depth_array`
```
njooma, line 44: Revert this comment change, as `bytes_to_depth_array` is on `ViamImage`

njooma (approved):
Minor update to the comments, but otherwise this is great

PR #568, opened 2024-03-28
RSDK-7014: Configure python to flush to stdout despite not being connected to a tty.

dgottlieb:
I'm not seeing any existing tests that simulate the bringing up a python module as a subprocess. I'm not inclined to be the first person to add that test...but I can if it's necessary.

### DIFF
--- a/src/viam/module/module.py
+++ b/src/viam/module/module.py
@@ -1,4 +1,5 @@
-import io, sys
+import io
+import sys
 from inspect import iscoroutinefunction
 from threading import Lock
 from typing import List, Mapping, Optional, Sequence, Tuple
--- a/src/viam/module/module.py
+++ b/src/viam/module/module.py
@@ -64,6 +64,16 @@ class Module:
         return cls(address, log_level=log_level)
 
     def __init__(self, address: str, *, log_level: int = logging.INFO) -> None:
+        if 'reconfigure' in dir(sys.stdout):
+            """Modules are launched from the viam-server, and a module's stdout is not connected to
+            a tty. Python reacts to that environment by disabling line buffering. Which means
+            `print` statements are not immediately flushed to the viam-server. This is confusing for
+            customers and can interfere with debugging. We reconfigure stdout and stderr here to
+            better align python with other languages/viam SDKs and user expectations.
+
+            Note: `reconfigure` only exists in python3.7+. """
+            sys.stdout.reconfigure(line_buffering=True)
+            sys.stderr.reconfigure(line_buffering=True)
         self._address = address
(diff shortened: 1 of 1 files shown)


stuqdog (approved):

src/viam/module/module.py
```diff
@@ -64,6 +64,16 @@ def from_args(cls) -> Self:
...
 
     def __init__(self, address: str, *, log_level: int = logging.INFO) -> None:
+        if 'reconfigure' in dir(sys.stdout):
+            """Modules are launched from the viam-server, and a module's stdout is not connected to
+            a tty. Python reacts to that environment by disabling line buffering. Which means
+            `print` statements are not immediately flushed to the viam-server. This is confusing for
+            customers and can interfere with debugging. We reconfigure stdout and stderr here to
+            better align python with other languages/viam SDKs and user expectations.
+
+            Note: `reconfigure` only exists in python3.7+. """
```
njooma, line 74: Could we make this docstring slightly less internal sounding and more for devs/users who might go poking around?

```
When a module is launched by viam-server, its stdout is not connected to a tty. 
In response, python desiables line buffering, which prevents `print` statements 
from being immediately flushed to viam-server. This behavior can be confusing, 
interfere with debugging, and is non-standard when compared to other languages. 
Here, stdout and stderr are reconfigured to immediately flush.
```
I think we can drop the note since we only support python3.8+ anyway

njooma (approved):
Just a nonblocking note about the comment, otherwise LGTM.

src/viam/module/module.py
dgottlieb: I wasn't aware of the python minimum. Should I remove the `dir` check then as well? The existence of that note was to justify the conditional `if 'reconfigure' in dir(sys.stdout)` bit.

dgottlieb: Also I'm not sure what this CI error is saying:
image

`stdout` and `stderr` are* of a TextIO[Wrapper] type. And the [API is published](https://docs.python.org/3/library/io.html#io.TextIOWrapper.reconfigure). Any guidance on if there's actually a problem here?

njooma: Yup you can remove the `if reconfig` check.

Also the CI error is about this:

```
# TextIO is used instead of more specific types for the standard streams,
# since they are often monkeypatched at runtime. At startup, the objects
# are initialized to instances of TextIOWrapper.
#
# To use methods from TextIOWrapper, use an isinstance check to ensure that
# the streams have not been overridden:
#
# if isinstance(sys.stdout, io.TextIOWrapper):
#    sys.stdout.reconfigure(...)
```

Easily fixable

PR #577, opened 2024-04-04
RSDK-7192 - Provisioning wrappers

stuqdog:


src/viam/app/provisioning_client.py
```diff
@@ -0,0 +1,95 @@
...
+            dial_options = DialOptions.with_api_key("<API-KEY>", "<API-KEY-ID>")
+            return await ViamClient.create_from_dial_options(dial_options)
+
+
+        async def main():
+
+            # Make a ViamClient
+            viam_client = await connect()
+            # Instantiate a ProvisioningClient to run provisioning client API methods on
+            provisioning_client = viam_client.provisioning_client
```
njooma, line 46: Don't you have to actually add the `provisioning_client` as a property on `viam_client` for this to work?

njooma (changes requested):

src/viam/app/provisioning_client.py
stuqdog: 🤦 yep you're totally right! Thanks for catching that.

njooma (approved):

PR #579, opened 2024-04-09
RSDK-7195: fix no event loop logging error

purplenicole730:
Tested with @thegreatco and confirmed that this fix works.


njooma (approved):
Nice! Can you just add comments in the code describing why we have to do this please

benjirewis (approved):
LGTM! 🧑‍🔧 .

PR #585, opened 2024-04-12
RSDK-7229 add machine_id to cloud metadata

maximpertsov:
~Breaking change!~ Not anymore!

Add `machine_id` to the `get_cloud_metadata` robot client method. ~Also change the return type and rename `robot_part_id` to `machine_part_id` for clarity.~ [We are just deprecating the duplicate field instead](https://github.com/viamrobotics/viam-python-sdk/pull/585#discussion_r1571261813).

Depends on:
* [x] https://github.com/viamrobotics/api/pull/487

### DIFF
--- a/src/viam/robot/client.py
+++ b/src/viam/robot/client.py
@@ -53,6 +53,16 @@ from viam.utils import datetime_to_timestamp, dict_to_struct
 LOGGER = logging.getLogger(__name__)
 
 
+@dataclass
+class CloudMetadata:
+    """App-related information about the robot"""
+
+    primary_org_id: str
+    location_id: str
+    machine_id: str
+    machine_part_id: str
+
+
 class RobotClient:
     """gRPC client for a Robot. This class should be used for all interactions with a robot.
 
@@ -779,7 +789,7 @@ class RobotClient:
     # Get Cloud Metadata #
     ######################
 
-    async def get_cloud_metadata(self) -> GetCloudMetadataResponse:
+    async def get_cloud_metadata(self) -> CloudMetadata:
         """
         Get app-related information about the robot.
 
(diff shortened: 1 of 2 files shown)


src/viam/robot/client.py
```diff
@@ -53,6 +53,16 @@
...
 
 
+@dataclass
+class CloudMetadata:
+    """App-related information about the robot"""
+
+    primary_org_id: str
+    location_id: str
+    machine_id: str
+    machine_part_id: str
```
njooma, line 63: I would rather use a `TypeAlias` here rather than creating a new type, especially since everything simply mirrors the `GetCloudMetadataResponse` anyway

njooma (changes requested):

stuqdog (approved):
lgtm (modulo Naveed's comment)

src/viam/robot/client.py
maximpertsov: would making a `TypeAlias` expose the `robot_part_id` field on `GetCloudMetadataResponse` too? this is a duplicate of `machine_part_id` - we left it to prevent a breaking proto change

njooma: Yes it would. So if you're trying to hide that field, then sure you can create a dataclass. But this is a brewaking change in all the SDKs anyway, so maybe might as well make it a breaking proto change?

maximpertsov: that's fair - I guess there was a feeling that breaking protos is a bit more significant than breaking the client method

maximpertsov: discussed IRL - we're gonna explicitly deprecate the proto field but leave it in: https://github.com/viamrobotics/api/pull/487

njooma (approved):

PR #590, opened 2024-04-18
RSDK-6907 StreamTicks API

oliviamiller:
Generalizes the `MediaStream` and moved to `streams.py` to use for digital interrupt streaming.


src/viam/components/board/board.py
```diff
@@ -390,3 +394,25 @@ async def write_analog(self, pin: str, value: int, *, timeout: Optional[float] =
             value (int): value to write.
         """
         ...
+
+    @abc.abstractmethod
+    async def stream_ticks(
+        self, interrupts: List[DigitalInterrupt], *, timeout: Optional[float] = None, **kwargs
+    ) -> Stream[Tick]:
```
stuqdog, line 401: (minor) since we defined `TickStream` above, should we not use it here in the return?

src/viam/components/board/board.py
```diff
@@ -390,3 +394,25 @@ async def write_analog(self, pin: str, value: int, *, timeout: Optional[float] =
...
+
+            my_board = Board.from_robot(robot=robot, name="my_board")
+            di8 = await my_board.digital_interrupt_by_name(name="8"))
+            di11 = await my_board.digital_interrupt_by_name(name="11"))
+
+            Stream ticks from pins 8 and 11.
+            ticks = my_board.stream_ticks([di8, di11])
+
+        Args:
+            interrupts (List[DigitalInterrupt]) : list of digital interrupts to recieve ticks from.
```
stuqdog, line 416: I think we should add a `Returns:` section to this docstring.

src/viam/components/board/client.py
```diff
@@ -243,3 +249,29 @@ async def write_analog(
...
         request = WriteAnalogRequest(name=self.name, pin=pin, value=value, extra=dict_to_struct(extra))
         await self.client.WriteAnalog(request, timeout=timeout)
+
+    async def stream_ticks(
+        self,
+        interrupts: List[Board.DigitalInterrupt],
+        *,
+        extra: Optional[Dict[str, Any]] = None,
+        **__,
+    ) -> Stream[Tick]:
```
stuqdog, line 259: (minor) similar to above, should this be `TickStream`? If not, should we remove the `TickStream` typedef?

src/viam/components/input/client.py
```diff
@@ -7,6 +7,7 @@
 from grpclib import GRPCError, Status
 from grpclib.client import Channel
 
+
```
stuqdog, line 10: (minor) should we have this newline? I thought only one linebreak separating imports was standard (disregard if I'm mistaken!).

stuqdog (approved):
A couple small things that shouldn't necessitate an extra round of review, otherwise LGTM!

njooma (approved):

PR #591, opened 2024-04-22
DOCS-2199: Viam Python SDK correctness pass

andf-viam:
First: This PR changes only code comments (that appear in [Python SDK docs](https://python.viam.dev/)). No code was harmed in the making of this PR!

Fix straightforward correctness issues that are interfering with automated scraping, such as missing params or returns, formatting (mostly indents), etc. Common issues encountered; fixed:
- Params with multiline descriptions must indent wrapped lines, fixed.
- Returns with multiline descriptions must not indent wrapped lines, fixed.
- Converted multi-item returns to single-item returns: they are all really single-item anyways -- 1 tuple -- and the multi-approach resulted in inconsistencies between usage syntax (at function `def`) and parameters (in dedicated Parameters (`Args:`) section).
	- Other instances of tuple returns already do this ([simple](https://python.viam.dev/autoapi/viam/components/movement_sensor/client/index.html#viam.components.movement_sensor.client.MovementSensorClient.get_position), [explicit (but also has an unrelated indentation issue)](https://python.viam.dev/autoapi/viam/app/app_client/index.html#viam.app.app_client.AppClient.list_organization_members)), so this just brings non-conformers inline. Consistency required for automation.
- Added some missing returns (but please double check my work!)
- Corrected some incorrect return types (but please double check my work!)
- Added missing data type (`str`) to [`delete_fragment`](https://python.viam.dev/autoapi/viam/app/app_client/index.html#viam.app.app_client.AppClient.delete_fragment).
- Some quick typo fixes.
- Moved code examples to be above params for few cases where observed.

NOTES / FURTHER ACTION:
Observed the following further inconsistencies that I cannot fix:
- Motion service is missing a `service.html` page entirely (has: [arm](https://python.viam.dev/_modules/viam/components/arm/service.html), [mlmodel](https://python.viam.dev/_modules/viam/services/mlmodel/service.html); 404: [motion](https://python.viam.dev/_modules/viam/services/motion/service.html)). Motion appears to be the only resource Viam-wide without one.
	- Perhaps relatedly, motion functions are all present in `client.py` file, also at odds with all other service resources, which use `{servicename}.py` for these method declarations.
- Almost all resources omit explicitly including `timeout` or `extra` in Params list (`Args:`), but Motion includes them explicitly. Should Motion remove these explicit (inherited?) entries?

Thanks Team! Happy to chat or answer any questions; hit me up. Apologies if I guessed wrong on any data types! As always, I greatly appreciate your time!

### DIFF
--- a/src/viam/app/app_client.py
+++ b/src/viam/app/app_client.py
@@ -573,7 +573,7 @@ class AppClient:
 
         Returns:
             Tuple[List[viam.proto.app.OrganizationMember], List[viam.proto.app.OrganizationInvite]]: A tuple containing two lists; the first
-                [0] of organization members, and the second [1] of organization invites.
+            [0] of organization members, and the second [1] of organization invites.
         """
         organization_id = await self._get_organization_id()
         request = ListOrganizationMembersRequest(organization_id=organization_id)
@@ -595,9 +595,9 @@ class AppClient:
                 authorizations to include in the invite. If not provided, full owner permissions will
                 be granted.
             send_email_invite (Optional[bool]): Whether or not an email should be sent to the recipient of an invite.
-            The user must accept the email to be added to the associated authorizations.
-            When set to false, the user automatically receives the associated authorization
-            on the next login of the user with the associated email address.
+                The user must accept the email to be added to the associated authorizations.
+                When set to false, the user automatically receives the associated authorization
+                on the next login of the user with the associated email address.
 
         Raises:
             GRPCError: if an invalid email is provided, or if the user is already a member of the org.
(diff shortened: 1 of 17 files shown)


src/viam/app/app_client.py
```diff
@@ -1370,7 +1370,7 @@ async def get_fragment(self, fragment_id: str) -> Fragment:
             # Get a fragment and print its name and when it was created.
             the_fragment = await cloud.get_fragment(
                 fragment_id="12a12ab1-1234-5678-abcd-abcd01234567")
-            print("Name: ", the_fragment.name, "\nCreated on: ", the_fragment.created_on)
+            print("Name: ", the_fragment.name, "\\nCreated on: ", the_fragment.created_on)
```
andf-viam, line 1373: Not sure if this works, is this how to escape newlines in code examples? Alternatively, likely safe to remove the newline. As is it [breaks formatting](https://python.viam.dev/autoapi/viam/app/app_client/index.html#viam.app.app_client.AppClient.get_fragment).

stuqdog:
> Almost all resources omit explicitly including timeout or extra in Params list (Args:), but Motion includes them explicitly. Should Motion remove these explicit (inherited?) entries?

Yes, I suspect that `Motion` should stop including `extra` and `timeout` in the docstring for consistency with the rest of the SDK. I'm not sure why it differs in this way; @njooma do you know?

Probably also we should have a `motion.py` that defines an abstract base class that the client inherits from, but that definitely seems outside of scope here.

stuqdog (approved):
lgtm!

andf-viam:
@stuqdog thank so much! Mergin' in!

PR #593, opened 2024-04-24
Unimplemented plugin

njooma:
grpclib in python is not forwards compatible (see [issue](https://github.com/vmagamedov/grpclib/issues/168)). 

This was getting really annoying for people who didn't make gRPC changes, but had to deal with other people's changes because tests would fail. 

So I made a protoc plugin that's the [grpclib plugin](https://github.com/vmagamedov/grpclib/blob/master/grpclib/plugin/main.py) with an added bit that will automatically generate an `UnimplementedServiceBase` similar to how it happens in [golang](https://github.com/viamrobotics/rdk/blob/main/robot/server/server.go#L44).

This generated a fair few additional lines of code, but it's all generated boilerplate.

You still have to opt-in to using the unimplemented version by subclassing from it. Otherwise everything continues to work as normal. I'm considering making a PR upstream for this


plugin/main.py
```diff
@@ -0,0 +1,279 @@
...
+                buf.add(
+                    "async def {}(self, stream: '{}.{}[{}, {}]') -> None:",
+                    name,
+                    server.__name__,
+                    server.Stream.__name__,
+                    request_type,
+                    reply_type,
+                )
+                with buf.indent():
+                    buf.add("raise grpclib.exceptions.GRPCError(grpclib.const.Status.UNIMPLEMENTED)")
```
njooma, line 141: These are the lines I added compared to the [grpclib](https://github.com/vmagamedov/grpclib/blob/master/grpclib/plugin/main.py) version

src/viam/robot/service.py
njooma, line 1: I tested the new plugin by removing all the unimplemented methods from the `RobotService`. Everything still works as expected, and everything typechecks.

purplenicole730 (approved):
This is amazing; I didn't know this could be done. LGTM!

stuqdog (approved):
LGTM! Just for my own understanding, are there any concerns around important upstream changes being lost such that pushing this upstream becomes higher priority?

njooma:
> LGTM! Just for my own understanding, are there any concerns around important upstream changes being lost such that pushing this upstream becomes higher priority?

Not really. The last time this file was updated was 10 months ago, and then 2 years ago, and then 4 years ago. So I don't think we lose much by bringing this in house.

PR #598, opened 2024-04-29
[RSDK-7443] paginate data

njooma:


agreenb (changes requested):
Thanks for the quick updates, and for updating BinaryDataByFilter!

Mostly just manually testing questions: Were you able to test these different permutations against the app server? Thinking of looping over 'last' and confirming that it retrieves paginated data (especially with a time interval such that you can test running out of data), requesting with count_only, using the unspecified vs a different limit, using the unspecified vs a DESCENDING sort order?

stuqdog (approved):

src/viam/app/data_client.py
```diff
@@ -184,47 +188,61 @@ def __init__(self, channel: Channel, metadata: Mapping[str, str]):
...
+                tabular_data, last = await data_client.tabular_data_by_filter(my_filter, last)
+                if not tabular_data:
+                    break
+                my_data.extend(tabular_data)
+
 
         Args:
             filter (viam.proto.app.data.Filter): Optional `Filter` specifying tabular data to retrieve. No `Filter` implies all tabular
                 data.
+            limit (int): The maximum number of entries to include in a page. Defaults to 50.
```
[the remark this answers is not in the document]
njooma: Yup! Didn't realize the server set the limit, thought we had to do it in the client.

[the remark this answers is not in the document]
agreenb: Great, could possibly update the comment to 'default to 50 if unspecified' to make it clear?

src/viam/app/data_client.py
```diff
@@ -233,59 +255,72 @@ async def tabular_data_by_filter(
...
+            filter (viam.proto.app.data.Filter): Optional `Filter` specifying tabular data to retrieve. No `Filter` implies all binary
+                data.
+            limit (int): The maximum number of entries to include in a page. Defaults to 50.
+            sort_order (viam.proto.app.data.Order): The desired sort order of the data.
+            last (str): Optional string indicating the ID of the last-returned data.
+                        If provided, the server will return the next data entries after the `last` ID.
+            include_binary_data (bool): Boolean specifying whether to actually include the binary file data with each retrieved file.
+                                        Defaults to true (i.e., both the files' data and metadata are returned).
+            count_only (bool): Whether to return only the total count of entries.
+            include_internal_data (bool): Whether to return the internal data. Internal data is used for Viam-specific data ingestion,
```
agreenb, line 299: Since we're exposing `include_internal_data` exposed in the SDK, would great to add the corresponding [param](https://github.com/viamrobotics/api/blob/3e1d32e77cdb555adc41efc72a92259818b5cecb/proto/viam/app/data/v1/data.proto#L169) for TabularDataByFilter for consistency

src/viam/app/data_client.py
```diff
@@ -184,47 +188,65 @@ def __init__(self, channel: Channel, metadata: Mapping[str, str]):
...
-                for struct in response.data
-            ]
-            last = response.last
+
+        data_request = DataRequest(filter=filter)
+        if limit:
+            data_request.limit = limit
+        if sort_order:
+            data_request.sort_order = sort_order
+        if last:
```
agreenb, line 237: Awesome, thanks for updating here and in BinaryDataByFilter! Similar to my PR-level question about testing, could we confirm that keeping these unspecified gives back what we expect in an SDK call?

src/viam/app/data_client.py
njooma: Should it be included?? I can remove it. I saw it in the proto so I added it

src/viam/app/data_client.py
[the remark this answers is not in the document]
njooma: Done

src/viam/app/data_client.py
njooma: I haven't manually tested this code -- I don't have a robot with a ton of data. But also, this is just client code. I added tests to make sure that the client is forwarding values properly to the server, but beyond that the client is simply going to trust that the server returns back the appropriate data for the request

agreenb: Got it, so confirming that if these are unspecified, the proto field is unspecified. If so, that matches what we'd want.

agreenb (approved):

src/viam/app/data_client.py
agreenb: LGTMed, but I just added you into the Data/ML Dev Org so that you can test against this, just so we can be extra sure that this is following the behavior we expect before giving customer updates.

PR #601, opened 2024-04-29
Build process

njooma:
Update the build process so that we aren't duplicating a lot of code. This also enables us to build wheels outside of the process of a release, which is useful for the instance where we're manually testing releases (like for `v0.21.0`)

### DIFF
--- a/.github/workflows/update_protos.yml
+++ b/.github/workflows/update_protos.yml
@@ -30,9 +30,6 @@ jobs:
         env:
           BUF_TOKEN: ${{ secrets.BUF_TOKEN }}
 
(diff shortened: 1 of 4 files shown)


.github/workflows/update_protos.yml
```diff
@@ -30,9 +30,6 @@ jobs:
         env:
           BUF_TOKEN: ${{ secrets.BUF_TOKEN }}
 
-      - name: Generate better imports
-        run: make better_imports
-
```
njooma, line 35: This step is now included in the `make buf` step

PR #603, opened 2024-05-02
RSDK-7189: add data wrappers

purplenicole730:


tests/mocks/services.py
```diff
@@ -916,7 +918,11 @@ async def GetDatabaseConnection(self, stream: Stream[GetDatabaseConnectionReques
         await stream.send_message(GetDatabaseConnectionResponse(hostname=self.hostname_response))
 
     async def ConfigureDatabaseUser(self, stream: Stream[ConfigureDatabaseUserRequest, ConfigureDatabaseUserResponse]) -> None:
-        raise NotImplementedError()
+        request = await stream.recv_message()
+        assert request is not None
+        self.organization_id = request.organization_id
+        self.password = request.password
```
stuqdog, line 924: We assign these values here but never do anything with them (same for `organization_id` and `{s|m}ql_query` below). If we want to keep these around we should probably test that they're properly set in the appropriate tests. That said, typically I think of these kinds of tests (confirming a value has been set on the class object) being important for methods that don't otherwise return anything, which isn't the case here. So probably it's easiest to just get rid of these entirely.

stuqdog (approved):
Nice! One minor comment about the tests (no need to re-request review), otherwise LGTM!

njooma (approved):

PR #606, opened 2024-05-08
[RSDK-7267] Board cleanup part 1: analogs can be written to

penguinland:
- `AnalogReader` is now just `Analog`. Similarly, `analog_reader_by_name` is just `analog_by_name`, etc.
- The `Analog` abstract class can now `.write()` in addition to `.read()`.
- (The underlying RPCs are unchanged: it's still a `WriteAnalog` RPC on the board. but SDK users don't interact directly with the RPCs.)

In an upcoming PR, I'll have more board cleanup to stream digital interrupts. but that's separate enough to warrant a separate PR.

### DIFF
--- a/src/viam/components/board/board.py
+++ b/src/viam/components/board/board.py
@@ -71,7 +71,7 @@ class Board(ComponentBase):
             ::
                 my_board = Board.from_robot(robot=robot, name="my_board")
 
-                # Get the AnalogWriter "my_example_analog_reader".
+                # Get the AnalogWriter "my_example_analog_writer".
                 writer = await my_board.analog_by_name(
                     name="my_example_analog_writer")
 
--- a/src/viam/components/board/board.py
+++ b/src/viam/components/board/board.py
@@ -44,7 +44,7 @@ class Board(ComponentBase):
         @abc.abstractmethod
         async def read(self, *, extra: Optional[Dict[str, Any]] = None, timeout: Optional[float] = None, **kwargs) -> int:
             """
-            Read the current value from a reader.
+            Read the current value from the reader.
 
             ::
 
--- a/src/viam/components/board/board.py
+++ b/src/viam/components/board/board.py
@@ -15,7 +15,7 @@ TickStream = Stream[Tick]
 class Board(ComponentBase):
     """
     Board represents a physical general purpose compute board that contains various
-    components such as analog readers, and digital interrupts.
+    components such as analog readers/writers, and digital interrupts.
 
     This acts as an abstract base class for any drivers representing specific
     board implementations. This cannot be used on its own. If the ``__init__()`` function is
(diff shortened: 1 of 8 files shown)


penguinland:
Don't worry, github-actions bot. I didn't touch `board.gpio_pin_by_name`.

src/viam/components/board/board.py
```diff
@@ -30,45 +30,55 @@ class Board(ComponentBase):
...
 
             ::
 
                 my_board = Board.from_robot(robot=robot, name="my_board")
 
-                # Get the GPIOPin with pin number 15.
-                pin = await my_board.gpio_pin_by_name(name="15")
-
-                # Get if it is true or false that the pin is set to high.
-                duty_cycle = await pin.get_pwm()
```
penguinland, line 57: This paragraph is unrelated to my main change, but also has nothing to do with the function it's documenting. Reading an analog pin doesn't involve PWM, and I suspect this "documentation" was added via a copy-and-paste mistake.

src/viam/components/board/board.py
```diff
@@ -30,45 +30,55 @@ class Board(ComponentBase):
...
+        """The name of the analog pin"""
 
         def __init__(self, name: str):
             self.name = name
 
         @abc.abstractmethod
         async def read(self, *, extra: Optional[Dict[str, Any]] = None, timeout: Optional[float] = None, **kwargs) -> int:
             """
-            Read the current value.
+            Read the current value from a reader.
```
njooma, line 47: [nit] `from THE reader` rather than `from A reader`, since this method is on a specific instance

src/viam/components/board/service.py
```diff
@@ -176,8 +176,12 @@ async def WriteAnalog(self, stream: Stream[WriteAnalogRequest, WriteAnalogRespon
         assert request is not None
         name = request.name
         board = self.get_resource(name)
+        try:
+            analog_writer = await board.analog_by_name(request.pin)
+        except ResourceNotFoundError as e:
+            raise e.grpc_error
         timeout = stream.deadline.time_remaining() if stream.deadline else None
-        await board.write_analog(pin=request.pin, value=request.value, timeout=timeout, metadata=stream.metadata)
+        await analog_writer.write(value=request.value, timeout=timeout, metadata=stream.metadata, extra=request.extra)
```
njooma, line 184: You have to convert `request.extra` from a `Struct` to a `dict`: `struct_to_dict(request.extra)`

This will also get rid of the type checking error that's currently causing the build to fail.

njooma (changes requested):

penguinland:
I've needed to rerun the tests several times because they time out after 1 minute, and only 95ish% of the tests finish in a minute. Is this expected? I'm unsure if I've somehow made the tests much slower than normal, or the timeout threshold ought to be raised. With enough reruns, the tests _can_ pass within the time limit...

stuqdog (approved):
One minor thing (no need to re-request review) otherwise lgtm!

penguinland:
> In an upcoming PR, I'll have more board cleanup to stream digital interrupts. 

Surprise, no I won't! Olivia already did that 2 weeks ago.

PR #614, opened 2024-05-14
RSDK-7188: add app wrappers

purplenicole730:


src/viam/app/app_client.py
```diff
@@ -1529,6 +1712,53 @@ async def remove_role(
...
+                new_identity_id="abc01234-0123-4567-ab12-a11a00a2aa22",
+                new_role="owner",s
+                new_resource_type="organization",
+                new_resource_id="abc12345")
+
+        Args:
+            old_identity_id (str): ID of the entity the role belongs to (e.g., a user ID).
+            old_role (Union[Literal["owner"], Literal["operator"]]): The role to be changed.
+            old_resource_type (Union[Literal["organization"], Literal["location"], Literal["robot"]]): Type of the resource the role is
+                added to. Must match `resource_id`.
```
stuqdog, line 1744: (nit) `s/resource_id/old_resource_id`

src/viam/app/app_client.py
```diff
@@ -1529,6 +1712,53 @@ async def remove_role(
...
+            old_resource_id (str): ID of the resource the role applies to (i.e., either an organization, location, or robot ID).
+
+            new_identity_id (str): New ID of the entity the role blongs to (e.g., a user ID).
+            new_role (Union[Literal["owner"], Literal["operator"]]): The new role.
+            new_resource_type (Union[Literal["organization"], Literal["location"], Literal["robot"]]): Type of the resource to add role to.
+                Must match `resource_id`.
```
stuqdog, line 1750: (nit) `s/resource_id/new_resource_id`

stuqdog (changes requested):
Nice, this is looking good overall! Just a couple nits and one slight "gotcha" of a refactor request

src/viam/app/app_client.py
```diff
@@ -1511,9 +1694,9 @@ async def remove_role(
 
         Args:
             identity_id (str): ID of the entity the role belongs to (e.g., a user ID).
-            role (Union[Literal["owner"], Literal["operator"]]): The role to add.
-            resource_type (Union[Literal["organization"], Literal["location"], Literal["robot"]]): Type of the resource to add role to.
-                Must match `resource_id`.
+            role (Union[Literal["owner"], Literal["operator"]]): The role to remove.
+            resource_type (Union[Literal["organization"], Literal["location"], Literal["robot"]]): Type of the resource the role is added
+                to. Must match `resource_id`.
```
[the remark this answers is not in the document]
purplenicole730: Yeah, that makes sense. I was just thinking about the fact that it was added to the resource_type previously, but we are removing it so "being removed from" makes a lot of sense!

src/viam/app/app_client.py
```diff
@@ -453,9 +481,39 @@ async def _create_authorization_for_new_api_key(self, org_id: str, auth: APIKeyA
             resource_id=auth._resource_id,
         )
 
-    # TODO(RSDK-5569): implement
+    async def get_user_id_by_email(self, email: str) -> str:
+        """Get the ID of a user by email.
+
+        ::
+
+            id = await cloud.get_user_id_by_email("youremail@email.com")`
```
stuqdog, line 489: (nit) we still have this trailing ` here, which we should probably get rid of.

stuqdog (approved):
Nice, one nit otherwise looks great! Thanks for also going back and updating docstring comments from previously implemented methods, much appreciated :)

njooma (approved):

PR #616, opened 2024-05-15
RSDK-7462 - Update Motion service in Python SDK

nfranczak:
We are updating the python SDK to match the following changes made in API (https://github.com/viamrobotics/api/commit/503a9a41e4ef9c8e5c8399a8890c6b2ae923d96b)

We would like to also explicitly call out the changes reviewers should be aware of:
1. `GeoObstacle` has been renamed into `GeoGeometry` - we note that they store the same information and this is just a re-name.
This means that the return type of `GetObstacles` in the navigation service has changes from `GeoObstacle` into `GeoGeometry`
2. We have added an argument to the motion service method MoveOnGlobe. The argument is called `bounding_regions` and is a list of `GeoGeometry`s
3. The motion service method MoveOnGlobe also takes in a list of obstacles of type `GeoObstacle`. This has been updated to `GeoGeometry`


src/viam/services/motion/client.py
```diff
@@ -122,10 +122,11 @@ async def move_on_globe(
         component_name: ResourceName,
         destination: GeoPoint,
         movement_sensor_name: ResourceName,
-        obstacles: Optional[Sequence[GeoObstacle]] = None,
+        obstacles: Optional[Sequence[GeoGeometry]] = None,
         heading: Optional[float] = None,
         configuration: Optional[MotionConfiguration] = None,
         *,
+        boudning_regions: Optional[Sequence[GeoGeometry]] = None,
```
stuqdog, line 129: (nit) `s/boudning/bounding`

src/viam/services/motion/client.py
```diff
@@ -175,6 +176,7 @@ async def move_on_globe(
                 - plan_deviation_m (float): The distance in meters that the machine can deviate from the motion plan.
                 - linear_m_per_sec (float): Linear velocity this machine should target when moving.
                 - angular_degs_per_sec (float): Angular velocity this machine should target when turning.
+            boudning_regions (Optional[Sequence[GeoGeometry]]): Set of obstacles which the robot must remain within while navigating
```
stuqdog, line 179: (nit) `s/boudning/bounding`

stuqdog (commented):
These changes all look good to me. However, I don't know that the way we're handling protos here (updating in this PR rather than the `update-protos` automated job) is the best approach. Historically we've had PRs like this branch off of the `update-protos` PR, then merge into that and then merge it all into `main`. I _think_ what we're doing here is probably okay but would want to get buy-in on that from others before being too confident.

njooma (changes requested):
Please base this branch off of the branch `workflow/update-protos`, as that already has the protos compiled and will make this PR a lot smaller and easier to follow

PR #617, opened 2024-05-16
RSDK-7565, RSDK-7471: add new methods get_properties and capture_all_from_camera

bhaney:
Adds two new methods, and tests for them

- GetProperties
- CaptureAllFromCamera

Adding it to the rc-0.21.0 branch

PR #619, opened 2024-05-16
[RSDK-7154] update analog reader interface to include range/accuracy

penguinland:
Surprisingly, the compiled protobufs in this repo already have the new data, and all I needed to do was start using it!

When reading an analog pin on a board, you now get the raw value, the minimum/maximum possible value the reader can output, and the minimum change the reader can notice.

### DIFF
--- a/src/viam/components/board/board.py
+++ b/src/viam/components/board/board.py
@@ -1,13 +1,18 @@
 import abc
-from dataclasses import dataclass
 from datetime import timedelta
+import sys
 from typing import Any, Dict, Final, List, Optional
 
-from viam.proto.component.board import PowerMode, StreamTicksResponse
+from viam.proto.component.board import PowerMode, StreamTicksResponse, ReadAnalogReaderResponse
 from viam.resource.types import RESOURCE_NAMESPACE_RDK, RESOURCE_TYPE_COMPONENT, Subtype
 from viam.streams import Stream
 
 from ..component_base import ComponentBase
+
+if sys.version_info >= (3, 10):
+    from typing import TypeAlias
+else:
+    from typing_extensions import TypeAlias
 
 Tick = StreamTicksResponse
 TickStream = Stream[Tick]
@@ -39,22 +44,12 @@ class Board(ComponentBase):
         name: str
         """The name of the analog pin"""
 
-        @dataclass
-        class Value:
-            """
-            Data obtained from reading an analog pin.
-            """
-            value: int
-            """The raw bits from the analog reader"""
-
-            min_range: float
-            """The minimum value the analog reader can output"""
-
-            max_range: float
-            """The maximum value the analog reader can output"""
-
-            step_size: float
-            """The precision per bit of the reading"""
+        Value: "TypeAlias" = ReadAnalogReaderResponse
+        """
+        Value contains the result of reading an analog reader. It contains the raw data read,
+        the reader's minimum and maximum possible values, and its step size (the minimum possible
+        change between values it can read).
+        """
 
         def __init__(self, name: str):
             self.name = name
--- a/src/viam/components/board/board.py
+++ b/src/viam/components/board/board.py
@@ -1,4 +1,5 @@
 import abc
+from dataclasses import dataclass
 from datetime import timedelta
 from typing import Any, Dict, Final, List, Optional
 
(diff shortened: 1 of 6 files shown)


src/viam/components/board/board.py
```diff
@@ -38,11 +38,23 @@ class Analog:
         name: str
         """The name of the analog pin"""
 
+        class Value:
```
njooma, line 41: You might want to make this a `@dataclass` to automatically generate an initializer

src/viam/components/board/board.py
```diff
@@ -38,11 +38,23 @@ class Analog:
...
+            """
+            Data obtained from reading an analog pin. The value is the raw bits obtained, while the
+            range indicates indicates the what the extreme values indicate. The step_size is the
+            precision per bit of the reading. The value read, when interpreted in
+            volts/degrees/etc., is (min_range + value * step_size)
+            """
+            value: int
+            min_range: float
+            max_range: float
+            step_size: float
```
njooma, line 51: Please put the individual attribute docs as docstrings. Then, the class docstring can be how to use this class

```python3
value: int
"""The raw bits obtained"""

min_range: float
"""The minimum of the range"""

max_range: float
"""..."""

step_size: float
"""Precision per bit of the reading"""
```

src/viam/components/board/client.py
```diff
@@ -48,12 +48,17 @@ async def read(
...
         if extra is None:
             extra = {}
         request = ReadAnalogReaderRequest(board_name=self.board.name, analog_reader_name=self.name, extra=dict_to_struct(extra))
         response: ReadAnalogReaderResponse = await self.board.client.ReadAnalogReader(request, timeout=timeout)
-        return response.value
+        result = self.Value()
+        result.value = response.value
+        result.min_range = response.min_range
+        result.max_range = response.max_range
+        result.step_size = response.step_size
```
njooma, line 60: Once you make `Value` a `@dataclass`, you'll be able to move this into the constructor

njooma (changes requested):

src/viam/components/board/board.py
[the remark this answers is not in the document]
njooma: I think this is fine. Value is a bit overloaded (particularly with protobuf.Value), but it being namespaced inside of Analog is perfect

src/viam/components/board/board.py
```diff
@@ -38,11 +39,28 @@ class Analog:
         name: str
         """The name of the analog pin"""
 
+        @dataclass
+        class Value:
```
njooma, line 43: You could probably make this a `TypeAlias` if all the attributes are the same

```
if sys.version_info >= (3, 10):
    from typing import TypeAlias
else:
    from typing_extensions import TypeAlias

class Board:
    class Analog:
        Value: "TypeAlias" = ReadAnalogReaderResponse
```

Similar to how it's done in [Vision](https://github.com/viamrobotics/viam-python-sdk/pull/617/files#diff-a7311669680bf0703ece9bd2c4887bb9bf823cd40ffd8c4154d56c4f38733e15R58)

njooma (commented):
Comment, open for discussion

src/viam/components/board/board.py
njooma: This way, we don't have to keep `Value` in sync with the response type should any fields be added/removed

stuqdog (approved):
lgtm, modulo @njooma's `TypeAlias` suggestion (which I agree with).

src/viam/components/board/board.py
penguinland: Oh dang, that's _super_ clean! I had no idea that was a thing. Thanks!

src/viam/components/board/client.py
```diff
@@ -48,12 +48,14 @@ async def read(
...
-    ) -> int:
+    ) -> Board.Analog.Value:
         if extra is None:
             extra = {}
         request = ReadAnalogReaderRequest(board_name=self.board.name, analog_reader_name=self.name, extra=dict_to_struct(extra))
         response: ReadAnalogReaderResponse = await self.board.client.ReadAnalogReader(request, timeout=timeout)
-        return response.value
+        result = self.Value(value=response.value, min_range=response.min_range,
+                            max_range=response.max_range, step_size=response.step_size)
+        return result
```
njooma, line 58: you should just be able to do `return response` now

njooma (changes requested):

src/viam/components/board/client.py
penguinland: Good idea! Take one more look.

src/viam/components/board/service.py
```diff
@@ -139,7 +139,8 @@ async def ReadAnalogReader(self, stream: Stream[ReadAnalogReaderRequest, ReadAna
             raise e.grpc_error
         timeout = stream.deadline.time_remaining() if stream.deadline else None
         value = await analog_reader.read(extra=struct_to_dict(request.extra), timeout=timeout, metadata=stream.metadata)
-        response = ReadAnalogReaderResponse(value=value)
+        response = ReadAnalogReaderResponse(value=value.value, min_range=value.min_range,
+                                            max_range=value.max_range, step_size=value.step_size)
         await stream.send_message(response)
```
njooma, line 144: You should be able to just do `stream.send_message(value)` now that `value` is of type `ReadAnalogReaderResponse`

njooma (changes requested):

src/viam/components/board/service.py
penguinland: Sorry to make you catch all these things that I should have caught myself. I appreciate it; I think this version of the code looks way better than the original!

njooma (approved):
nice

PR #622, opened 2024-05-21
Force kwargs in DialOptions

maximpertsov:
Removes a small footgun - it's easy to write code like `dial_options=DialOptions({"insecure": True})` and wonder why SSL is still required.


purplenicole730 (approved):
Hahaha, thanks for this!

njooma (approved):

njooma:
This is technically a breaking change tho

PR #623, opened 2024-05-21
Simplify the vision service code sample

bhaney:
Changes the code sample for the front-end to use the new "get_properties"


bhaney:
@mestcihazal on May 28th the code samples should be simplified for the vision service

purplenicole730 (approved):

PR #624, opened 2024-05-21
[RSDK-4561] Remove unused check

njooma:
When `MovementSensor` was still a subclass of `Sensor`, we had to exclude adding registering a `Sensor` subtype with the robot for every `MovementSensor` so as to not duplicate the available resources. 

Now that `MovementSensor` is distinct from `Sensor`, we can remove that check


purplenicole730 (approved):
Seems like the only real change was the removal in `client.py`, and the other changes were formatting? If so, LGTM!

stuqdog (approved):

PR #629, opened 2024-05-30
RSDK-7776 - Add mime_types to GetPropertiesResponse

seanavery:
Simple PR to add `mime_types` to `GetProperties` response. This field was added to the camera `GetPropertiesResponse ` proto a few months back - see PR [here](https://github.com/viamrobotics/api/pull/408).

### DIFF
--- a/tests/mocks/components.py
+++ b/tests/mocks/components.py
@@ -371,6 +371,7 @@ class MockCamera(Camera):
             supports_pcd=False,
             intrinsic_parameters=IntrinsicParameters(width_px=1, height_px=2, focal_x_px=3, focal_y_px=4, center_x_px=5, center_y_px=6),
             distortion_parameters=DistortionParameters(model="no_distortion"),
+            mime_types=[CameraMimeType.PNG, CameraMimeType.JPEG],
         )
         self.timeout: Optional[float] = None
         ts = Timestamp()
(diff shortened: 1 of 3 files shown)


tests/mocks/components.py
```diff
@@ -371,6 +371,7 @@ def __init__(self, name: str):
             supports_pcd=False,
             intrinsic_parameters=IntrinsicParameters(width_px=1, height_px=2, focal_x_px=3, focal_y_px=4, center_x_px=5, center_y_px=6),
             distortion_parameters=DistortionParameters(model="no_distortion"),
+            mime_types=[CameraMimeType.PNG, CameraMimeType.JPEG],
```
randhid, line 374: I know we have a bunch more, is there any advantages to adding H264 or PointCloud at this stage? Okay to keep it scoped down for the sdks that do not return properties as an object, please coordinate with @hexbabe to make sure C++ is in line with python.

seanavery: Python SDK shows:
```
    VIAM_RGBA = "image/vnd.viam.rgba"
    VIAM_RAW_DEPTH = "image/vnd.viam.dep"
    JPEG = "image/jpeg"
    PNG = "image/png"
    PCD = "pointcloud/pcd"
```
Just added a couple here in the mock camera for tests.

AFAIK h264 "mime_type" has not been fully flushed out yet. We do have a case for encoded images in gostream [here](https://github.com/viamrobotics/rdk/blob/main/gostream/stream.go#L271) and a h264 mime type in RDK. However, do not think it was ever working fully.

njooma (approved):

PR #632, opened 2024-06-07
RSDK-7854 - fix local dialing, server_hostname setting

stuqdog:
Creates an unverified context to allow for proper connection when dialing to localhost. Additionally, prevents us from improperly setting `server_hostname` to our `api_key_id` when authing with an api-key.

This fix allows us to connect properly to `local.viam.cloud` addresses as well as to `localhost`.

### DIFF
--- a/src/viam/rpc/dial.py
+++ b/src/viam/rpc/dial.py
@@ -313,12 +313,16 @@ async def _dial_direct(address: str, options: Optional[DialOptions] = None) -> C
     if insecure:
         ctx = None
     else:
-        ctx = ssl.create_default_context(purpose=ssl.Purpose.SERVER_AUTH)
+        is_local_host = host.startswith("localhost") or host.startswith("0.0.0.0") or host.startswith("127.")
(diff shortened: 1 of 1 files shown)


src/viam/rpc/dial.py
```diff
@@ -313,12 +313,22 @@ async def _dial_direct(address: str, options: Optional[DialOptions] = None) -> C
     if insecure:
         ctx = None
     else:
-        ctx = ssl.create_default_context(purpose=ssl.Purpose.SERVER_AUTH)
+        is_local_host = host is not None and (host.startswith("localhost") or host.startswith("0.0.0.0") or host.startswith("127."))
```
stuqdog, line 316: Not positive about how prioritization of bool operators works in python, added the parens around the `or` statements for explicitness but I'm not sure if it's actually necessary?

njooma: I never am either. I like this because it's explicit

njooma (approved):

PR #633, opened 2024-06-07
RSDK-7698: Add Shutdown command to Python SDK

danielbotros:
### Description
[**RSDK-7698**](https://viam.atlassian.net/browse/RSDK-7698): Add Shutdown command to Python SDK
* Implements `RobotClient` `shutdown()` with the same error handling logic as the Go SDK seen in [this PR](https://github.com/viamrobotics/rdk/pull/4042)
### Testing
* Adds a robot client test by creating a mock robot shutdown service which sends a `ShutdownResponse()` and a mock robot client shutdown which makes a `ShutdownRequest()` and returns the `ShutdownResponse()`

### DIFF
--- a/src/viam/robot/client.py
+++ b/src/viam/robot/client.py
@@ -796,16 +796,24 @@ class RobotClient:
 
     async def shutdown(self):
         """
-        Shutdown shuts down the robot. update comment here
+        Shutdown shuts down the robot. May return DeadlineExceeded error if shutdown request times out,
+        or if robot server shuts down before having a chance to send a response. May return Unavailable error
+        if server is unavailable, or if robot server is in the process of shutting down when response is ready.
 
         """
         request = ShutdownRequest()
         try:
             await self._client.Shutdown(request)
+            LOGGER.info("robot shutdown successful")
         except GRPCError as e:
             if e.status == Status.INTERNAL or e.status == Status.UNKNOWN:
+                LOGGER.info("robot shutdown successful")
                 pass
-            elif e.status == Status.DEADLINE_EXCEEDED:
+            elif e.status == Status.UNAVAILABLE:
+                LOGGER.warn("server unavailable, likely due to successful robot shutdown")
                 raise e
-            # unfinishd logic
-            raise e
+            elif e.status == Status.DEADLINE_EXCEEDED:
+                LOGGER.warn("request timeout, robot shutdown may still be successful")
+                raise e
+            else:
+                raise e
--- a/src/viam/robot/client.py
+++ b/src/viam/robot/client.py
@@ -796,10 +796,13 @@ class RobotClient:
 
     async def shutdown(self):
         """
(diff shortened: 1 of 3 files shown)


tests/test_robot.py
```diff
@@ -599,3 +609,21 @@ async def get_geometries(
...
+    @pytest.mark.asyncio
+    async def test_shutdown(self, service: RobotService):
+        async with ChannelFor([service]) as channel:
+
+            async def shutdown_client_mock(self):
+                return await self._client.Shutdown(ShutdownRequest())
+
+            client = await RobotClient.with_channel(channel, RobotClient.Options())
+
+            with mock.patch("viam.robot.client.RobotClient.shutdown") as shutdown_mock:
```
benjirewis, line 622: Why do we need `mock.patch` here? I don't see it elsewhere in this file.

benjirewis (approved):
Nice! Just a question.

tests/test_robot.py
danielbotros: `shutdown()` doesn't return anything (may only potentially raise an error), so there is no way to check any properties of it actually being called i.e. writing a non-trivial test. By `mock.patch`ing it we can use `assert_called_once()` to check that it is actually being called (and also mock a return value that it receives from the mock shutdown service!)

src/viam/robot/client.py
```diff
@@ -787,3 +789,30 @@ async def get_cloud_metadata(self) -> GetCloudMetadataResponse:
...
+
+    ############
+    # Shutdown #
+    ############
+
+    async def shutdown(self):
+        """
+        Shutdown shuts down the robot. May return DeadlineExceeded error if shutdown request times out,
+        or if robot server shuts down before having a chance to send a response. May return Unavailable error
+        if server is unavailable, or if robot server is in the process of shutting down when response is ready.
```
stuqdog, line 801: (minor) This isn't formatted consistently with the SDK. Probably better to do something like
```
"""
Shuts down the robot.

Raises:
  <ErrorTypeFoo>: If {foo}
  <ErrorTypeBar>: If {bar}
"""
```

You should be able to see examples of this elsewhere in this file (see, e.g., the `get_component` method).

src/viam/robot/client.py
```diff
@@ -787,3 +789,30 @@ async def get_cloud_metadata(self) -> GetCloudMetadataResponse:
...
+
```
stuqdog, line 802: (minor) delete newline here.

stuqdog (commented):
Some very minor formatting changes requested for the `shutdown` docstring, otherwise looks good to me!

tests/test_robot.py
njooma: I know I'm the one who helped with this idea, but I'm realizing now that you could've just changed the service implementation to raise an error with a message, and then in this test check to see if that specific error and specific message was raised. 

This also works.

njooma (approved):
Looks good % @stuqdog's changes

stuqdog (approved):

PR #635, opened 2024-06-10
Add new fields to apis.json

njooma:
In support of making it easier for fleet to have accurate code samples

### DIFF
--- a/examples/apis.json
+++ b/examples/apis.json
@@ -14,10 +14,25 @@
   "gantry": { "func": "get_lengths" },
   "gripper": { "func": "is_moving" },
   "movement_sensor": { "func": "get_linear_acceleration" },
-  "input_controller": { "func": "get_controls" },
+  "input_controller": {
+    "func": "get_controls" ,
+    "fullImportText": "from viam.components.input import Controller",
+    "packageName": "viam.components.input",
+    "importName": "Controller"
+  },
   "audio": { "func": "get_properties" },
   "pose_tracker": { "func": "get_poses" },
   "power_sensor": { "func": "get_power" },
   "motion": { "func": "get_pose" },
-  "vision": { "func": "get_properties" }
+  "vision": { "func": "get_properties" },
+  "mlmodel": {
+    "fullImportText": "from viam.services.mlmodel import MLModelClient",
+    "packageName": "viam.services.mlmodel",
+    "importName": "MLModelClient"
+  },
+  "slam": {
+    "fullImportText": "from viam.services.slam import SLAMClient",
+    "packageName": "viam.services.slam",
+    "importName": "SLAMClient"
+  }
 }
--- a/examples/apis.json
+++ b/examples/apis.json
@@ -1,11 +1,13 @@
 {
   "base": {
-    "func": "is_moving"
+    "func": "is_moving",
+    "packagePath": "viam.components"
   },
   "board": {
     "func": "gpio_pin_by_name",
     "comment": "Note that the pin supplied is a placeholder. Please change this to a valid pin you are using.",
-    "args": ["\"16\""]
+    "args": ["\"16\""],
+    "packagePath": "viam.components"
   },
   "camera": {
     "func": "get_image",
@@ -76,7 +78,7 @@
     "importName": "MLModelClient"
   },
   "slam": {
-    "fullImportText": "viam.services",
+    "packagePath": "viam.services",
     "importName": "SLAMClient"
   }
 }
--- a/examples/apis.json
+++ b/examples/apis.json
@@ -1,38 +1,82 @@
 {
-  "base": { "func": "is_moving" },
+  "base": {
+    "func": "is_moving"
+  },
   "board": {
     "func": "gpio_pin_by_name",
     "comment": "Note that the pin supplied is a placeholder. Please change this to a valid pin you are using.",
     "args": ["\"16\""]
   },
-  "camera": { "func": "get_image" },
-  "encoder": { "func": "get_position" },
-  "motor": { "func": "is_moving" },
-  "sensor": { "func": "get_readings" },
-  "servo": { "func": "get_position" },
-  "arm": { "func": "get_end_position" },
-  "gantry": { "func": "get_lengths" },
-  "gripper": { "func": "is_moving" },
-  "movement_sensor": { "func": "get_linear_acceleration" },
+  "camera": {
+    "func": "get_image",
+    "packagePath": "viam.components"
+  },
+  "encoder": {
+    "func": "get_position",
+    "packagePath": "viam.components"
+  },
+  "motor": {
+    "func": "is_moving",
+    "packagePath": "viam.components"
+  },
+  "sensor": {
+    "func": "get_readings",
+    "packagePath": "viam.components"
+  },
+  "servo": {
+    "func": "get_position",
+    "packagePath": "viam.components"
+  },
+  "arm": {
+    "func": "get_end_position",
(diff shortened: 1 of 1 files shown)


stuqdog (approved):

examples/apis.json
```diff
@@ -14,10 +14,25 @@
   "gantry": { "func": "get_lengths" },
   "gripper": { "func": "is_moving" },
   "movement_sensor": { "func": "get_linear_acceleration" },
-  "input_controller": { "func": "get_controls" },
+  "input_controller": {
+    "func": "get_controls" ,
+    "fullImportText": "from viam.components.input import Controller",
+    "packageName": "viam.components.input",
+    "importName": "Controller"
```
DTCurrie, line 21: Ok so looking at this and looking at the code, I am going to make suggest something different. Let's start with the code that generates a python code sample:

```ts
const resourceToImportName = (resource: ReadonlyCommonResource): string => {
  if (isComponent(resource)) {
    if (resource.api.subtype === 'input_controller') {
      return 'Controller';
    }
    return pascalCase(resource.api.subtype);
  }

  let serviceImportName;
  switch (resource.api.subtype) {
    case 'mlmodel': {
      serviceImportName = 'MLModel';
      break;
    }
    case 'slam': {
      serviceImportName = 'SLAM';
      break;
    }
    default: {
      serviceImportName = pascalCase(resource.api.subtype);
    }
  }
  return `${serviceImportName}Client`;
};

// Python uses unique names for some subtypes, catch those here
const subtypeToImportName = (subtype: string): string => {
  if (subtype === 'input_controller') {
    return 'input';
  }

  return subtype;
};

const generateImports = (resources: ReadonlyCommonResource[]): string => {
  const resourceImports = [
    ...new Set(
      resources.map((resource) =>
        isComponent(resource)
          ? `from viam.components.${subtypeToImportName(
              resource.api.subtype
            )} import ${resourceToImportName(resource)}`
          : `from viam.services.${subtypeToImportName(
              resource.api.subtype
            )} import ${resourceToImportName(resource)}`
      )
    ),
  ];
  return resourceImports.join(`\n`);
};

const generatePythonFunction = (
  resource: ReadonlyCommonResource,
  supportedResource: SupportedResource
) => {
  const { name } = resource;
  const resourceName = snakeCase(name);
  const retVal = `${snakeCase(name)}_return_value`;
  const { func, args, comment } = supportedResource;
  const importValue = `${resourceName} = ${resourceToImportName(resource)}`;
  const commentVal = `\n    # ${comment}`;
  return `${comment ? commentVal : ''}
    # ${name}
    ${importValue}.from_robot(machine, "${name}")
    ${retVal} = await ${resourceName}.${func}(${args ? args.join(', ') : ''})
    print(f"${name} ${func} return value: {${retVal}}")`;
};
```

Looking at this, I think we tweak this slightly to have `packagePath`, `packageName`, and `importName`.

```suggestion
    "packagePath": "viam.components",
    "packageName": "input",
    "importName": "Controller"
```

Then we can rewrite the above code to something a little simpler:

```ts
const generateImports = (
  resources: ReadonlyCommonResource[],
  supportedResources: SupportedResourceBySubtype
): string => {
  const resourceImports = new Set<string>();
  for (const resource of resources) {
    const supportedResource = supportedResources[resource.api.subtype];

    if (supportedResource !== undefined) {
      const { packagePath } = supportedResource;
      // use provided package/import name or fallback to subtype
      const packageName = supportedResource.packageName ?? resource.api.subtype;
      const importName = supportedResource.importName ?? resource.api.subtype;

      resourceImports.add(
        `from ${packagePath}.${packageName} import ${importName}`
      );
    }
  }

  return [...resourceImports].join(`\n`);
};

const generatePythonFunction = (
  resource: ReadonlyCommonResource,
  supportedResource: SupportedResource
) => {
  const { name, api } = resource;
  const resourceName = snakeCase(name);
  const retVal = `${snakeCase(name)}_return_value`;
  const { func, args, comment, importName } = supportedResource;

  // use provided import name or fallback to just pascal case of subtype
  const importValue = `${resourceName} = ${
    importName ?? pascalCase(api.subtype)
  }`;

  const commentVal = `\n    # ${comment}`;

  return `${comment ? commentVal : ''}
    # ${name}
    ${importValue}.from_robot(machine, "${name}")
    ${retVal} = await ${resourceName}.${func}(${args ? args.join(', ') : ''})
    print(f"${name} ${func} return value: {${retVal}}")`;
};
```

The important thing to note here is this makes `packagePath` a required field for every entry similar to `func`, but that doesn't feel too bad. Makes it obvious what is a component vs a service.

njooma: Would this be extensible for `go` or other languages as well? Or would each language's `apis.json` file need a different set of fields?

DTCurrie: Good question, I will take a look at the other generation code and see how a similar refactor would work.

njooma: Looks like `go` doesn't need the `packagePath` or the `packageName`, but does need the `importName` for anything that's not default. If we wanted to make `packagePath` required across all `apis.json`, we could and change the function `.../code-sample/golang/golang.ts:generateImports` to:

```ts
const generateImports = (resources: ReadonlyCommonResource[]): string => {
  const resourceImports = [
    ...new Set(
      resources.map((resource) => {
        return `${apis.importPath}/${apis.importName}`
      })
    ),
  ];
  return resourceImports.join(`
    `);
};
```

where `apis.importPath` is obtained from the `apis.json` file (I don't know how that's ingested but assume it's available?)

DTCurrie: Yeah that looks about right, here is what I did with a similar refactor for go:

```ts
const generateImports = (
  resources: ReadonlyCommonResource[],
  supportedResources: SupportedResourceBySubtype
): string => {
  const resourceImports = new Set<string>();
  for (const resource of resources) {
    const supportedResource = supportedResources[resource.api.subtype];

    if (supportedResource !== undefined) {
      const { packagePath } = supportedResource; // assuming this is go.viam.com/rdk/components or go.viam.com/rdk/services
      // use provided package/import name or fallback to subtype
      const packageName = supportedResource.packageName ?? resource.api.subtype; // assuming this is input or movementsensor etc... when applicable
      // const importName = supportedResource.importName ?? resource.api.subtype; // it looks like we shouldn't need this for go

      resourceImports.add(`"${packagePath}/${packageName}"`);
    }
  }

  return [...resourceImports].join(`\n`);
};
```

njooma: Perf ok so I'll make that change for go and update the ticket

purplenicole730 (approved):

examples/apis.json
```diff
@@ -1,23 +1,82 @@
 {
-  "base": { "func": "is_moving" },
+  "base": {
+    "func": "is_moving"
```
DTCurrie, line 3: I think we landed on making `packagePath` required, is it missing here?

examples/apis.json
```diff
@@ -1,23 +1,82 @@
...
+    "func": "get_properties",
+    "packagePath": "viam.services",
+    "importName": "VisionClient"
+  },
+  "mlmodel": {
+    "packagePath": "viam.services",
+    "importName": "MLModelClient"
+  },
+  "slam": {
+    "fullImportText": "viam.services",
```
DTCurrie, line 79:
```suggestion
    "packagePath": "viam.services",
```

DTCurrie (approved):
✅ 

Just to be safe, let's merge this after the 11am deploy in app. I will work on getting the reactors we talked about into a PR today so we should be able to test this in staging and see it go live tomorrow.

PR #638, opened 2024-06-11
DOCS-2313: Make fixups to app client, data client, and motion service package docs

sguequierre:
* Fix code samples, parameter and return descriptions and formatting
* Update descriptions for more info on grabbing organization and dataset ids 
* Add periods and remove "e.g". bc it causes linting error in docs.viam (we aren't supposed to use it)

Note: Motion move_on_globe and move_on_map configuration object formatting: sub-bulleted list -- still looks slightly odd with indenting but couldn't figure out how to completely fix

### DIFF
--- a/src/viam/app/data_client.py
+++ b/src/viam/app/data_client.py
@@ -200,8 +200,9 @@ class DataClient:
             dest (str): Optional filepath for writing retrieved data.
 
         Returns:
-            List[TabularData]: The tabular data.
-            int: The count (number of entries)
+            Tuple[List[TabularData], int, str]: A tuple containing the following:
+            List[TabularData]: The tabular data,
+            int: The count (number of entries),
             str: The last-returned page ID.
         """
         filter = filter if filter else Filter()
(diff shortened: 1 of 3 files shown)


src/viam/app/data_client.py
```diff
@@ -773,8 +780,16 @@ async def configure_database_user(self, organization_id: str, password: str) ->
         """Configure a database user for the Viam organization's MongoDB Atlas Data Federation instance. It can also be used to reset the
         password of the existing database user.
 
+        ::
```
npentrel, line 783: Why is this `::` here? copy paste error?

npentrel (approved):
LGTM % one small issue

src/viam/app/data_client.py
sguequierre: @npentrel To start the code sample!

njooma (approved):
Out of curiosity, why does `e.g.` break docs.viam?

sguequierre:
> Out of curiosity, why does `e.g.` break docs.viam?

It doesn't really break it, it was decided that it wasn't allowed in the docs re. style conventions and the Vale linter raises an error for it that stops merging

npentrel:
@njooma it doesn't break. We've got a linter set up to remind us not to use it, per our style guide. Within the docs community it's a generally accepted convention to move away from latin abbreviations for accessibility reasons, so it's a good change to make across the board

stuqdog (approved):

PR #641, opened 2024-06-11
DOCS-2313: Fix docstrings in arm and encoder where link not parsing

sguequierre:
Some final fixups from this ticket

* Fix encoder PositionType.ValueType by changing to PositionType (like in other parts of docstring) + fix code sample (value type seems to be off)
* Fix arm KinematicsFileFormat.ValueType by giving the two value types in docstring (link to KinematicsFileFormat does not explain much)


src/viam/components/encoder/encoder.py
```diff
@@ -64,23 +64,24 @@ async def get_position(
...
-        ``PositionType.TICKS`` or absolute units (degrees along a circle) for ``PositionType.DEGREES``.
+        ``PositionType.POSITION_TYPE_TICKS_COUNT`` or absolute units (degrees along a circle)
+        for ``PositionType.POSITION_TYPE_ANGLE_DEGREES``.
 
         ::
 
             my_encoder = Encoder.from_robot(robot=robot, name='my_encoder')
 
             # Get the position of the encoder in ticks
-            position = await my_encoder.get_position(encoder.PositionTypeTicks)
```
sguequierre, line 74: LMK if this is wrong as far as I can tell new way is way to do it?

src/viam/components/encoder/encoder.py
```diff
@@ -64,23 +64,24 @@ async def get_position(
...
+            position = await my_encoder.get_position(PositionType.POSITION_TYPE_TICKS_COUNT)
             print("The encoder position is currently ", position[0], position[1])
 
         Args:
-            position_type (PositionType.ValueType): The desired output type of the position.
```
njooma, line 78: This technically needs to be `PositionType.ValueType` because of the way protobuf implements these types of enumerations. If this links, then it should link to the `.ValueType`

njooma (approved):
One small thing that's a technicality and honestly I'd be ok not implementing it because it's just confusing overall and I think most developers will understand

src/viam/components/encoder/encoder.py
sguequierre: Ahh ok-- the link wasn't being generated by the parser. I'll try adding it manually?

src/viam/components/encoder/encoder.py
```diff
@@ -64,23 +64,27 @@ async def get_position(
...
+            position_type (PositionType): The desired output type of the position.
 
         Returns:
-            Tuple[float, PositionType]: A tuple containing two values; the first [0] the Position of the encoder
-            which can either be ticks since last zeroing for a relative encoder or degrees for an absolute encoder,
-            and the second [1] the type of position the encoder returns (ticks or degrees).
+            Tuple[float,
+            `PositionType.ValueType <https://python.viam.dev/autoapi/viam/gen/component/encoder/v1/\
+            encoder_pb2/index.html#viam.gen.component.encoder.v1.encoder_pb2.PositionType>`__]:
```
npentrel, line 84:
```suggestion
            `PositionType.ValueType <https://python.viam.dev/autoapi/viam/gen/component/encoder/v1/
            encoder_pb2/index.html#viam.gen.component.encoder.v1.encoder_pb2.PositionType>`__]:
```

src/viam/components/encoder/encoder.py
njooma: Ohh interesting. Ok for linking purposes I think we can keep this as just `PositionType` then. The actual underlying type is a property of `PositionType`, so it should be straightforward. And since the actual method signature has the correct type, typechecking should work just fine. 

Up to y'all as this only affects docs.

purplenicole730 (approved):

src/viam/components/encoder/encoder.py
sguequierre: That's good with me, I figured it was straightforward as well

npentrel:
@sguequierre lgtm if you've confirmed this fixes the issue and renders correctly

sguequierre:
> @sguequierre lgtm if you've confirmed this fixes the issue and renders correctly

yup, renders correctly changed to PositionType as Naveed suggested!

PR #643, opened 2024-06-13
RSDK-7732 EasyResource mixin

abe-winter:
## What changed
- add `EasyResource` mixin class for shorter model classes
- `run_from_registry` and `run_with_models` entrypoints in `Module` class
- docstrings, tests, example
## Why
Reduce friction to create new models + get started with our platform.
## Checklist
- [x] resolve + remove `todo(review)` comments


src/viam/resource/easy_resource.py
```diff
@@ -0,0 +1,84 @@
...
+from .base import ResourceBase
+from .types import Model, ModelFamily, Subtype
+from .registry import Registry, ResourceCreatorRegistration
+
+modelRegex = re.compile(r"^([^:]+):([^:]+):([^:]+)$")
+
+logger = logging.getLogger(__name__)
+
+
+def parse_model(orig: Union[str, Model]) -> Model:
```
njooma, line 17: [nit] can we make this "private" (start with `_` underscore)

src/viam/resource/easy_resource.py
```diff
@@ -0,0 +1,84 @@
...
+        return self
+
+    @classmethod
+    def register(cls):
+        """
+        This adds the model to the global registry. It is called by __init_subclass__ and you typically
+        won't call it directly.
+        """
+        logger.debug('registering %s %s', cls.SUBTYPE, cls.MODEL)
+        # note: this would pass if EasyResource inherited ResourceBase, but that crashes in the mro() walk in ResourceManager.register.
```
njooma, line 79: What does this comment mean? What would "pass"?

src/viam/module/module.py
```diff
@@ -35,6 +37,16 @@
 LOGGER = logging.getLogger(__name__)
 
 
+def parse_module_args() -> argparse.Namespace:
```
njooma, line 40: Perhaps make this private as well

src/viam/module/module.py
```diff
@@ -57,13 +69,40 @@ def from_args(cls) -> Self:
...
+    async def run_with_models(cls, *models: ResourceBase):
+        """
+        Module entrypoint that takes a list of ResourceBase subclasses.
+        In most cases you'll want to use run_from_registry instead (see below).
+        """
+        args = parse_module_args()
+        module = cls(args.socket_path, log_level=args.log_level)
+        for model in models:
+            # todo(review): the pyright ignore below is for ResourceBase.MODEL. Is this not a required field? Should I not rely on it?
+            module.add_model_from_registry(model.SUBTYPE, model.MODEL)  # pyright: ignore [reportAttributeAccessIssue]
```
njooma, line 90: `ResourceBase` does not have a `MODEL` field at all, since `MODEL`s are for implementation of a `ResourceBase`, and `ResourceBase` is for creating the resource definition/interface. This would fail in the possible event that someone passes `Arm` as the param instead of `MyModuleArm`

njooma: So would be useful to maybe do a `hasattr` here and fail if someone doesn't pass the correct type. And also add that disclaimer to the docstring

src/viam/module/module.py
```diff
@@ -57,13 +69,40 @@ def from_args(cls) -> Self:
...
+        await module.start()
+
+    @classmethod
+    async def run_from_registry(cls):
+        """
+        Module entrypoint that automatically includes all the resources you've created in your program.
+        For usage, see the bottom of examples/easy_resource/main.py.
```
njooma, line 97: Would be good to have usage instructions here, rather than forcing someone to go to the example

src/viam/module/module.py
```diff
@@ -57,13 +69,40 @@ def from_args(cls) -> Self:
...
+        """
+        args = parse_module_args()
+        module = cls(args.socket_path, log_level=args.log_level)
+        for key in Registry.REGISTERED_RESOURCE_CREATORS().keys():
+            # todo(review): this would be cleaner if resource creator key became a tuple
```
njooma, line 102: I agree, seems like it'd be easy but it is technically a breaking change, but I don't know anyone who would be using the Registry externally like that

src/viam/module/module.py
```diff
@@ -57,13 +69,40 @@ def from_args(cls) -> Self:
...
+            module.add_model_from_registry(*key.split('/'))  # pyright: ignore [reportArgumentType]
```
njooma, line 103: Dislike the disable pyright here but also seems dumb to create a Subtype just to stringify it again in the next method call. No action needed, just writing it down

njooma (changes requested):
This is awesome

src/viam/module/module.py
abe-winter: non-breaking middle ground: `REGISTERED_RESOURCE_CREATORS()` transforms the tuple key to a string key, have another getter that provides access to the tuple?

src/viam/module/module.py
abe-winter: add hasattr w/ helpful error

I didn't expand docstring because it seems obvious that the provided resources have to be complete + runnable? but can add that if you think it's helpful

src/viam/resource/easy_resource.py
njooma: Ah gotcha 👍

njooma (approved):

PR #648, opened 2024-06-14
use poetry cache in CI

abe-winter:
## What changed
- add `cache: poetry` param to setup-python action
## Why
Faster builds
## Comparison
I did a run [here](https://github.com/viamrobotics/viam-python-sdk/actions/runs/9518864325) and then reran again after it succeeded (to get cache).

Speedup is about 40 seconds for the slowest job (2:30 -> 1:50).

Before and after:
before-after

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,6 +32,7 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
+          cache: poetry
 
       - name: Install Poetry
         uses: snok/install-poetry@v1
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,14 +28,13 @@ jobs:
         uses: actions/checkout@v3
         with:
           ref: ${{ github.event.pull_request.head.sha }}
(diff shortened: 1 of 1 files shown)


njooma (approved):
Surprised that "Install Poetry" works before "Setup python"

I wanted to test if the min/max requirements worked with the cache and I guess the answer is yes! Thanks for doing the test for me

abe-winter:
I'm not actually sure the min version step is getting cached

times are:

run | install | min version
initial | 45s | 26
cached | 2s | 26

not sure, but guessing it's not using the requirements file as part of the cache key? (but not sure that should matter; would have thought each matrix job has its own cache)

abe-winter:
switching to `pipx install poetry` per [their example](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#caching-packages) seems to gain a few seconds (slowest cached job is now 1:20-1:30 vs 1:52 before)

also the min requirements step is installing pillow and it is not available as a wheel for new python I think -- if not required anymore, that seems to be the limiting step now

njooma:
We still need Pillow in our tests because we have PIL-specific utilities

PR #651, opened 2024-06-17
Remove RC step and move all to Release workflow

njooma:
With the move to a weekly release cadence, remove the need to do RC releases. Instead, we now cut a release directly from `main`, bump the version, open a PR back into main with the new version number, and create a draft release off the new version branch. 

If any bugs come up between creating the release and actual release, simply delete the release, fix the bug directly into main, and then re-run the release workflow.


purplenicole730 (approved):
Woohoo! 🎉

stuqdog (approved):

PR #652, opened 2024-06-18
Remove latin abbreviations

npentrel:
I think that's the last of the latin abbreviations - thanks for letting us change these.


lia-viam:
The current changes look fine, but if we're removing latin abbreviations entirely maybe we want to get rid of "e.g." as well? I'm still seeing some hits for that one

npentrel:
good point, I will do that in this PR too

npentrel:
Do you think we could merge this today @lia-viam ? We want to do a demo in on Thursday with these docs so it would be nice to have it then :)

stuqdog (approved):

PR #661, opened 2024-06-21
RSDK-7732 add stub_model class decorator

abe-winter:
## What changed
- add `stub_model` class decorator that will implement the abstract methods of an ABC
- the stubbed methods are appropriately sync or async based on the method they're overriding. they log the name of the called method + raise a `MethodNotImplementedError`
- added examples/stubbed_model
- `stub_model` docstring + readme in new example both warn people not to use this in prod
- incidental: change type of `EasyResource.MODEL` class attr from `Model` to `Union[Model, str]`. this is so that type checkers don't reject classes which provide this as string. A potential other way to do this [recommended by mypy docs](https://mypy.readthedocs.io/en/latest/more_types.html#mixin-classes): use a protocol to type the mixin's inputs (cls + self), but maintain the Model type on the result type so it's clear what you're getting
## Why
When dashing off new models, it's useful to begin testing before your class is fully implemented, especially for classes with many methods like Motor or MovementSensor.
## Notes
- I gave this its own example rather than putting it in examples/easy_resource because the easy_resource example is supposed to be an ultra simple, compact starting point; didn't want to clutter it
- tried doing this in the existing mixin but the ABC `__abstractmethods__` collection wasn't available yet; decorator worked better


src/viam/resource/easy_resource.py
```diff
@@ -25,6 +28,50 @@ def _parse_model(orig: Union[str, Model]) -> Model:
...
 
+def _create_stub_fn(name: str, is_async: bool) -> Callable:
+    """
+    This creates a sync or async stub function which returns a MethodNotImplementedError.
+    The stub_model decorator uses these to stub out an abstract base class.
+    """
+    if is_async:
+        # note: this is a pyright bug https://github.com/microsoft/pyright/issues/2136
+        async def stub_fn(*args, **kwargs):  # pyright: ignore [reportRedeclaration]
+            logger.info(f'{name} not implemented')
```
njooma, line 39: Do we need the log statements? Seems redundant with the error type

src/viam/resource/easy_resource.py
```diff
@@ -25,6 +28,50 @@ def _parse_model(orig: Union[str, Model]) -> Model:
...
+    Normally this class would fail to instantiate. With the decorator, it will succeed but the unimplemented
+    methods will throw errors at runtime.
+    """
+    for attr in cls.__abstractmethods__:
+        val = getattr(cls, attr)
+        is_async = inspect.iscoroutinefunction(val)
+        stub_fn = _create_stub_fn(attr, is_async)
+        setattr(cls, attr, stub_fn)
+        logger.debug('patched %s.%s with %s', cls, attr, stub_fn)
+        cls.__abstractmethods__ -= {attr}
```
njooma, line 71: Updating the object while you're iterating over it... dangerous :P

src/viam/resource/easy_resource.py
```diff
@@ -41,7 +88,7 @@ async def get_readings(self, **kwargs):
     See examples/easy_resource/main.py for extended usage.
     """
     SUBTYPE: Subtype
-    MODEL: Model
+    MODEL: Union[Model, str]
```
njooma, line 91: I'm not sure how I feel about this -- I think this leaves a lot more chance for a user to provide a malformed/nonsense string

src/viam/resource/easy_resource.py
abe-winter: there are some edge cases in the Motor control card that don't get toast popups in app

(one method that is required to even display the motor RC card, another that is being polled in the background all the time; I think one is get properties and the other is is_moving)

src/viam/resource/easy_resource.py
abe-winter: reverted in 26c2b7e8

if this ends up causing type problems we can solve at leisure as separate project

src/viam/resource/easy_resource.py
abe-winter: :grimacing: eek solved in 6b3a1a50

njooma (approved):

PR #668, opened 2024-06-28
RSDK-7933 fix logging bugs

purplenicole730:
[Jira Ticket](https://viam.atlassian.net/browse/RSDK-7933)

fixed a bug where formatting logs with arguments did not work.

FLYBY: renamed the `log` method parameter `log` to `message` to keep it consistent with the RDK and the `ModuleHandler`.

### DIFF
--- a/src/viam/robot/client.py
+++ b/src/viam/robot/client.py
@@ -793,7 +793,7 @@ class RobotClient:
     # LOG #
     #######
 
-    async def log(self, name: str, level: str, time: datetime, log: str, stack: str):
+    async def log(self, name: str, level: str, time: datetime, message: str, stack: str):
         """Send log from Python module over gRPC.
 
         Create a LogEntry object from the log to send to RDK.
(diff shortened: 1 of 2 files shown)


benjirewis (approved):
Nice!

src/viam/robot/client.py
```diff
@@ -793,7 +793,7 @@ async def stop_all(self, extra: Dict[ResourceName, Dict[str, Any]] = {}):
     # LOG #
     #######
 
-    async def log(self, name: str, level: str, time: datetime, log: str, stack: str):
```
abe-winter, line 796: param rename is technically a breaking change if someone is passing named arguments

you're making it match the official API, so I'm guessing this is the right call, but flagging so we can debug quickly if it causes problems down the road

abe-winter (approved):

src/viam/robot/client.py
purplenicole730: Hmm, that is a good point! Thank you for bringing it up. I'll see what the others think.

njooma: Good call out. Let's make this change but call it out appropriately in the breaking changes channel

njooma (approved):

PR #677, opened 2024-07-08
[RSDK-8126]: Update fragment methods to use fragment visibility

agavram:
Context: https://viaminc.slack.com/archives/C02KFSY4H89/p1720458747424109

This addresses the above issue a customer is facing. (In the meantime, Shawn has figured out a workaround) 

With unlisted fragment changes already present in app, we should be making use of `FragmentVisibility` rather than `Public`/`ShowPublic`, and in a future PR, any references to `Public`/`ShowPublic` will be removed entirely.

### DIFF
--- a/src/viam/app/app_client.py
+++ b/src/viam/app/app_client.py
@@ -280,57 +280,55 @@ class LogEntry:
         )
 
 
-class FragmentVisibility(str, Enum):
-    """
-    FragmentVisibility specifies who is permitted to view the fragment.
-    """
-
-    PRIVATE = "private"
-    """
-    Only visible to members in the fragment's organization.
-    """
-
-    PUBLIC = "public"
-    """
-    Visible to anyone and appears on the fragments page.
-    """
-
-    PUBLIC_UNLISTED = "public_unlisted"
-    """
-    Visible to anyone but does not appear on the fragments page.
-    """
-
-    UNSPECIFIED = "unspecified"
-    """
-    Uninitialized visibility.
-    """
-
-
 class Fragment:
     """A class that mirrors the `Fragment` proto message.
 
     Use this class to make the attributes of a `viam.proto.app.RobotPart` more accessible and easier to read/interpret.
     """
 
-    @classmethod
-    def visibility_from_proto(cls, visibility: FragmentVisibilityPB.ValueType) -> FragmentVisibility:
-        if visibility == FragmentVisibilityPB.FRAGMENT_VISIBILITY_PRIVATE:
-            return FragmentVisibility.PRIVATE
-        if visibility == FragmentVisibilityPB.FRAGMENT_VISIBILITY_PUBLIC:
-            return FragmentVisibility.PUBLIC
-        if visibility == FragmentVisibilityPB.FRAGMENT_VISIBILITY_PUBLIC_UNLISTED:
-            return FragmentVisibility.PUBLIC_UNLISTED
-        return FragmentVisibility.UNSPECIFIED
+    class Visibility(str, Enum):
+        """
+        FragmentVisibility specifies who is permitted to view the fragment.
+        """
 
-    @classmethod
-    def visibility_to_proto(cls, visibility: FragmentVisibility) -> FragmentVisibilityPB.ValueType:
-        if visibility == FragmentVisibility.PRIVATE:
-            return FragmentVisibilityPB.FRAGMENT_VISIBILITY_PRIVATE
-        if visibility == FragmentVisibility.PUBLIC:
-            return FragmentVisibilityPB.FRAGMENT_VISIBILITY_PUBLIC
-        if visibility == FragmentVisibility.PUBLIC_UNLISTED:
-            return FragmentVisibilityPB.FRAGMENT_VISIBILITY_PUBLIC_UNLISTED
-        return FragmentVisibilityPB.FRAGMENT_VISIBILITY_UNSPECIFIED
+        PRIVATE = "private"
+        """
+        Only visible to members in the fragment's organization.
+        """
+
+        PUBLIC = "public"
+        """
+        Visible to anyone and appears on the fragments page.
+        """
+
+        PUBLIC_UNLISTED = "public_unlisted"
+        """
+        Visible to anyone but does not appear on the fragments page.
+        """
+
+        UNSPECIFIED = "unspecified"
+        """
+        Uninitialized visibility.
+        """
+
+        @classmethod
+        def from_proto(cls, visibility: FragmentVisibilityPB.ValueType) -> "Fragment.Visibility":
+            if visibility == FragmentVisibilityPB.FRAGMENT_VISIBILITY_PRIVATE:
+                return Fragment.Visibility.PRIVATE
+            if visibility == FragmentVisibilityPB.FRAGMENT_VISIBILITY_PUBLIC:
+                return Fragment.Visibility.PUBLIC
+            if visibility == FragmentVisibilityPB.FRAGMENT_VISIBILITY_PUBLIC_UNLISTED:
+                return Fragment.Visibility.PUBLIC_UNLISTED
+            return Fragment.Visibility.UNSPECIFIED
+
+        def to_proto(self) -> FragmentVisibilityPB.ValueType:
+            if self == self.PRIVATE:
+                return FragmentVisibilityPB.FRAGMENT_VISIBILITY_PRIVATE
+            if self == self.PUBLIC:
+                return FragmentVisibilityPB.FRAGMENT_VISIBILITY_PUBLIC
+            if self == self.PUBLIC_UNLISTED:
+                return FragmentVisibilityPB.FRAGMENT_VISIBILITY_PUBLIC_UNLISTED
+            return FragmentVisibilityPB.FRAGMENT_VISIBILITY_UNSPECIFIED
 
     @classmethod
     def from_proto(cls, fragment: FragmentPB) -> Self:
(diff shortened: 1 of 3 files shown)


piokasar:
re-tested locally with local viam python sdk pointing to local app and works as expected 👍

piokasar (approved):

src/viam/app/app_client.py
```diff
@@ -278,12 +280,58 @@ def proto(self) -> LogEntryPB:
...
+
+    PUBLIC_UNLISTED = "public_unlisted"
+    """
+    Visible to anyone but does not appear on the fragments page.
+    """
+
+    UNSPECIFIED = "unspecified"
+    """
+    Uninitialized visibility.
+    """
```
njooma, line 306: I would move this class to be a nested class within `Fragment` and rename it to just `Visibility`

src/viam/app/app_client.py
```diff
@@ -278,12 +280,58 @@ def proto(self) -> LogEntryPB:
...
+
+    @classmethod
+    def visibility_to_proto(cls, visibility: FragmentVisibility) -> FragmentVisibilityPB.ValueType:
+        if visibility == FragmentVisibility.PRIVATE:
+            return FragmentVisibilityPB.FRAGMENT_VISIBILITY_PRIVATE
+        if visibility == FragmentVisibility.PUBLIC:
+            return FragmentVisibilityPB.FRAGMENT_VISIBILITY_PUBLIC
+        if visibility == FragmentVisibility.PUBLIC_UNLISTED:
+            return FragmentVisibilityPB.FRAGMENT_VISIBILITY_PUBLIC_UNLISTED
+        return FragmentVisibilityPB.FRAGMENT_VISIBILITY_UNSPECIFIED
```
njooma, line 333: And move these methods to be inside of the enum class. Rename to simply `from_proto` and `to_proto`

njooma: Also, `to_proto` should not be a class method in the new version

src/viam/app/app_client.py
```diff
@@ -305,6 +353,7 @@ def from_proto(cls, fragment: FragmentPB) -> Self:
         self.robot_part_count = fragment.robot_part_count
         self.organization_count = fragment.organization_count
         self.only_used_by_owner = fragment.only_used_by_owner
+        self.visibility = self.visibility_from_proto(fragment.visibility)
```
njooma, line 356: And then call the new method as simply `fragment.visibility.to_proto()`

src/viam/app/app_client.py
```diff
@@ -278,12 +280,58 @@ def proto(self) -> LogEntryPB:
         )
 
 
+class FragmentVisibility(str, Enum):
```
njooma, line 283: In the [proto](https://github.com/viamrobotics/api/blob/c1ef199d69647d3ec7b78c05a7fc44641b644b0f/proto/viam/app/v1/app.proto#L730), this is technically an `int` rather than a string. I don't think it matters, but the string names don't provide any additional data so might be worthwhile changing it to an int to let it match the proto

src/viam/app/app_client.py
```diff
@@ -1581,25 +1632,25 @@ async def delete_robot(self, robot_id: str) -> None:
         request = DeleteRobotRequest(id=robot_id)
         await self._app_client.DeleteRobot(request, metadata=self._metadata)
 
-    async def list_fragments(self, org_id: str, show_public: bool = True) -> List[Fragment]:
+    async def list_fragments(self, org_id: str, show_public: bool = True, visibilities: List[FragmentVisibility] = []) -> List[Fragment]:
```
njooma, line 1635: This is a python gotcha where defaulting to an empty list can actually cause bugs (see the *Default parameter values are evaluated from left to right when the function definition is executed.* section of the [python docs](https://docs.python.org/3/reference/compound_stmts.html#function-definitions)). 

So if the actual parameter type should be `visibilities: Optional[List[FragmentVisibility]] = None` and the body of the method should check to see if `visibilities` is `None` and if so, set an empty list.

src/viam/app/app_client.py
```diff
@@ -1581,25 +1632,25 @@ async def delete_robot(self, robot_id: str) -> None:
...
         ::
 
-            fragments_list = await cloud.list_fragments(org_id="org-id", show_public=False)
+            fragments_list = await cloud.list_fragments(org_id="org-id", visibilities=[])
 
         Args:
             org_id (str): The ID of the organization to list fragments for.
                 You can obtain your organization ID from the Viam app's organization settings page.
-            show_public: Optional boolean specifying whether or not to only show public fragments. If True, only public fragments will
-                return. If False, only private fragments will return. Defaults to True.
```
njooma, line 1595: I think you should leave this in here for now (since the method signature still has it) and mark as deprecated

src/viam/app/app_client.py
```diff
@@ -1581,25 +1632,25 @@ async def delete_robot(self, robot_id: str) -> None:
...
+            visibilities: List of FragmentVisibilities specifying which types of fragments to include in the results.
+                If empty, by default only public fragments will be returned.
 
         Returns:
             List[viam.app.app_client.Fragment]: The list of fragments.
 
         For more information, see `Fleet Management API <https://docs.viam.com/appendix/apis/fleet/>`_.
         """
-        request = ListFragmentsRequest(organization_id=org_id, show_public=show_public)
+        request = ListFragmentsRequest(organization_id=org_id, fragment_visibility=map(Fragment.visibility_to_proto, visibilities))
```
njooma, line 1653: Similarly, I think you should continue to include `show_public` since we haven't pulled it out yet, giving people time to migrate to the new visibilities workflow

src/viam/app/app_client.py
```diff
@@ -1670,8 +1726,9 @@ async def update_fragment(
             name (str): New name to associate with the fragment.
             config (Optional[Mapping[str, Any]]): Optional Dictionary representation of new config to assign to specified fragment. Not
                 passing this parameter will leave the fragment's config unchanged.
-            public (bool): Boolean specifying whether the fragment is public. Not passing this parameter will leave the fragment's
-                visibility unchanged. A fragment is private by default when created.
```
njooma, line 1674: Same as above -- leave this in but mark as deprecated

njooma (changes requested):

src/viam/app/app_client.py
agavram: On the app side, we have completely removed any references to show_public, so passing it here wouldn't do anything. I can add a comment stating it is deprecated though

src/viam/app/app_client.py
agavram: Interesting, thanks for this call out!

src/viam/app/app_client.py
agavram: True, but I'm imagining the case where the user is looking at `Fragment.visibility` where it would be more useful to see `"public"` rather than an integer value representing the same thing.

src/viam/app/app_client.py
agavram: Nevermind! Talked with Katherine and we will be adding the `show_public` functionality on the app side

piokasar: Yeah , spoke with Michael this morning and he suggested we just accommodate both fields again for a while

agavram:
Thanks for the quick review @njooma, just pushed the latest set of changes to address the feedback

src/viam/app/app_client.py
```diff
@@ -1581,25 +1630,33 @@ async def delete_robot(self, robot_id: str) -> None:
...
+            fragments_list = await cloud.list_fragments(org_id="org-id", visibilities=[])
 
         Args:
             org_id (str): The ID of the organization to list fragments for.
                 You can obtain your organization ID from the Viam app's organization settings page.
-            show_public: Optional boolean specifying whether or not to only show public fragments. If True, only public fragments will
-                return. If False, only private fragments will return. Defaults to True.
+            visibilities: List of FragmentVisibilities specifying which types of fragments to include in the results.
+                If empty, by default only public fragments will be returned.
+            show_public: Deprecated: use visibilities instead. Optional boolean specifying whether or not to only show public
```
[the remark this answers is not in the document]
njooma: Also, add the following below the param doc:

```python3
show_public (bool): **Deprecated**: Use ``visibilities`` instead
    Optional boolean specifying...

    .. deprecated:: 0.25.0
        Use ``visibilities`` instead.
```

Be sure to include the new line and tabs. This will add a deprecated line to the rendered documentation
image

ALSO: please add type hints to these doc strings

src/viam/app/app_client.py
```diff
@@ -1670,8 +1732,11 @@ async def update_fragment(
             name (str): New name to associate with the fragment.
             config (Optional[Mapping[str, Any]]): Optional Dictionary representation of new config to assign to specified fragment. Not
                 passing this parameter will leave the fragment's config unchanged.
-            public (bool): Boolean specifying whether the fragment is public. Not passing this parameter will leave the fragment's
-                visibility unchanged. A fragment is private by default when created.
+            visibility (Optional[FragmentVisibility]): Optional FragmentVisibility list specifying who should be allowed
+                to view the fragment. Not passing this parameter will leave the fragment's visibility unchanged.
+                A fragment is private by default when created.
+            public (bool): Deprecated: use visibility instead. Boolean specifying whether the fragment is public. Not passing this parameter
```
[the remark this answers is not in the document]
njooma: Same comment as above re: deprecation tag

src/viam/app/app_client.py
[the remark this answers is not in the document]
njooma: You can check the rendering by running `make documentation` from the project root directory, and then opening the generated HTML and navigating to the function: `open docs/_build/html/autoapi/viam/app/app_client/index.html`

njooma (changes requested):

src/viam/app/app_client.py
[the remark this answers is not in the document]
piokasar: this is the current output I'm getting with the changes!!
image

njooma (approved):

PR #685, opened 2024-07-16
[RSDK-8045] connect to machine from viam client

njooma:
Adds the ability to connect to a machine with an already-connected Viam client

### DIFF
--- a/src/viam/app/viam_client.py
+++ b/src/viam/app/viam_client.py
@@ -219,7 +219,7 @@ class ViamClient:
 
         if id is not None and address is None:
             parts = await self.app_client.get_robot_parts(id)
-            main_part = [p for p in parts if p.main_part][0]
+            main_part = next(p for p in parts if p.main_part)
             address = main_part.fqdn
 
         opts = RobotClient.Options(dial_options=self._dial_options)
--- a/src/viam/app/viam_client.py
+++ b/src/viam/app/viam_client.py
@@ -9,7 +9,8 @@ from viam.app.billing_client import BillingClient
 from viam.app.data_client import DataClient
 from viam.app.ml_training_client import MLTrainingClient
 from viam.app.provisioning_client import ProvisioningClient
-from viam.rpc.dial import DialOptions, _dial_app, _get_access_token
+from viam.robot.client import RobotClient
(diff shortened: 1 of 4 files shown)


src/viam/services/service_base.py
```diff
@@ -29,8 +29,9 @@ def from_robot(cls, robot: "RobotClient", name: str) -> Self:
 
             async def connect() -> ViamClient:
                 # Replace "<API-KEY>" (including brackets) with your API key and "<API-KEY-ID>" with your API key ID
-                dial_options = DialOptions.with_api_key("<API-KEY>", "<API-KEY-ID>")
-                return await ViamClient.create_from_dial_options(dial_options)
+                options = RobotClient.Options.with_api_key("<API-KEY>", "<API-KEY-ID>")
+                # Replace "<MACHINE-URL>" (included brackets) with your machine's connection URL or FQDN
+                return await RobotClient.at_address("<MACHINE-URL>", options)
```
purplenicole730, line 34: `connect()` would now return a `RobotClient` instead of a `ViamClient`

src/viam/app/viam_client.py
```diff
@@ -179,3 +182,47 @@ def close(self):
...
+
+        Returns:
+            RobotClient: The active connection to the machine.
+        """
+        if address is None and id is None:
+            raise ValueError("Either a machine address or ID must be provided")
+
+        if id is not None and address is None:
+            parts = await self.app_client.get_robot_parts(id)
+            main_part = [p for p in parts if p.main_part][0]
```
purplenicole730, line 222: I don't know how main parts work; are there multiple main parts with the same address?

purplenicole730 (approved):
One question about main parts, and a quick comment fix. Otherwise, looks good to me! Thank you for adding a `viam_client` test!

src/viam/app/viam_client.py
njooma: There should be one main part per machine, and this will find it.

stuqdog (approved):

src/viam/app/viam_client.py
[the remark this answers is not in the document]
njooma: Tested it, and it's about 25% faster to do the iterator method so I updated

PR #686, opened 2024-07-17
RSDK-7970: finish adding submitcustomtrainingjob

purplenicole730:


tahiyasalam (approved):

src/viam/app/ml_training_client.py
```diff
@@ -119,7 +119,7 @@ async def submit_training_job(
         return response.id
 
     async def submit_custom_training_job(
-        self, org_id: str, dataset_id: str, registry_item_id: str, model_name: str, model_version: str
+        self, org_id: str, dataset_id: str, registry_item_id: str, registry_item_version: str, model_name: str, model_version: str
```
njooma, line 122: Adding an arg in the middle here means this becomes a breaking change. Which is fine, but everyone will need to be notified

njooma (approved):

PR #689, opened 2024-07-23
RSDK-8303: get and send versions

purplenicole730:
Changes:
- Get the `api_version` from the `update_protos.yml` workflow
- Get the new `sdk_version` from `release` workflow
- Store the `api_version` and `sdk_version` in a new `version.py`
- When sending a request, add `viam-client=python;{sdk_version};{api_version}`

Testing:
- Tested storing the new API version when running `update-protos.yml` by locally using `act` and putting in a custom `client_payload.tag` using `--eventpath` argument.
- Tested `send_request(event: SendRequest)` by including `raise Exception(event.metadata)` in the function. Result of running unit tests `FAILED tests/test_sessions_client.py::test_sessions_heartbeat_thread_blocked - Exception: <MultiDict('viam-client': 'python;0.25.2;0.1.327')>`. This confirms that `send_request` is being called, and the `event.metadata` is being updated.
- Tested against the RDK by printing the `motor.server.go.Stop.ctx`: `map[content-type:[application/grpc] hey:[nicole] te:[trailers] user-agent:[grpc-python-grpclib/0.4.7 (darwin; cpython/3.12.3)] viam-client:[python;0.25.2;0.1.327] viam-sid:[2aeb60f3-f815-46a1-8a4e-db180f333923]]`

RDK when using fmt.Printf("DEBUG INFO----\n\n%+v\n----\n\n", md):
```
DEBUG INFO----

map[content-type:[application/grpc] te:[trailers] user-agent:[grpc-python-grpclib/0.4.7 (darwin; cpython/3.12.3)] viam-client:[python;0.25.1;0.1.327] viam-sid:[f799b9aa-cef0-4ee2-b10a-20d6092aa2ec]]
----
```


pyproject.toml
```diff
@@ -1,3 +1,5 @@
+api_version = "0.1.327"
+
```
purplenicole730, line 2: Wasn't entirely sure where to put this other than where the SDK `version` is already sitting

src/viam/__init__.py
```diff
@@ -1,71 +1,73 @@
-import sys
-from importlib.metadata import PackageNotFoundError, version
-from uuid import uuid4
-
-from viam.gen.common.v1.common_pb2 import ResourceName as _ResourceName
```
purplenicole730, line 5: I don't understand what's going on here. I know that `update_api_version.py` asked for the file to be re-written, but it looks the same. I also checked on diffchecker, and it was confirmed that nothing changed.

src/viam/__init__.py
```diff
@@ -1,71 +1,73 @@
...
+##############
+# VERSIONING #
+##############
+try:
+    __version__ = version("viam-sdk")
+except PackageNotFoundError:
+    pass
+
+api_version = "0.1.327"
+
```
purplenicole730, line 18: Only lines that actually changed. `update_api_version.py` asked for the file to be re-written, but it should be the same. I also checked on diffchecker, and it was confirmed that nothing changed.

src/viam/rpc/dial.py
```diff
@@ -277,15 +278,28 @@ def free_str(self, ptr: ctypes.c_void_p):
 
 
 async def dial(address: str, options: Optional[DialOptions] = None) -> ViamChannel:
+    with open("../__init__.py", "r") as file:
+        match = re.search(r'api_version\s*=\s*"([0-9]+\.[0-9]+\.[0-9]+)"', file.read())
+        api_version = match.group(1) if match else None
+
+    with open("../../../pyproject.toml", "r") as file:
+        match = re.search(r'version\s*=\s*"([0-9]+\.[0-9]+\.[0-9]+)"', file.read())
+        sdk_version = match.group(1) if match else None
```
purplenicole730, line 287: For now, I could only find the version in `pyproject.toml` since `__init__.py` technically doesn't have the version in the file itself. Is there a way we can access this variable, something like `viam.__version__`?

src/viam/rpc/dial.py
```diff
@@ -277,15 +279,28 @@ def free_str(self, ptr: ctypes.c_void_p):
...
+    directory = os.path.dirname(__file__)
+    with open(os.path.join(directory, "../__init__.py"), "r") as file:
+        match = re.search(r'api_version\s*=\s*"([0-9]+\.[0-9]+\.[0-9]+)"', file.read())
+        api_version = match.group(1) if match else None
+    with open(os.path.join(directory, "../../../pyproject.toml"), "r") as file:
+        match = re.search(r'version\s*=\s*"([0-9]+\.[0-9]+\.[0-9]+)"', file.read())
+        sdk_version = match.group(1) if match else None
+
+    async def send_request(event: SendRequest):
+        event.metadata["viam-client"] = f'python;{sdk_version};{api_version}'
```
purplenicole730, line 291: Is there a better way to do this? I wish it could be a bit more elegant, but it does work!
I had to get the `sdk_version` from `pyproject.toml` since `__init__.py` does not have the version printed. Is that okay or is there a better way to get the version number?

src/viam/__init__.py
purplenicole730: Doesn't seem like it does, as my RDK test worked well.

stuqdog: Does this mean that this file will be completely rewritten every time API version updates? That doesn't seem great to me, I don't think we should be getting into the habit of periodically ignoring ~150 lines of diff in a non-computer-generated file. Could we figure out how to more selectively replace the value, or store the version in a smaller file that only contains that information, which is then imported by `__init__.py`?

src/viam/rpc/dial.py
purplenicole730: The `pyproject.toml` file was where I first wanted to put in the version, but Naveed realized that this file wasn't included when building.
I also thought about adding it to their own separate file, but we weren't sure what the best practice was. I personally like this approach, so let's see how it goes!

.github/workflows/release.yml
```diff
@@ -66,6 +66,11 @@ jobs:
         if: |
           steps.release_exists.outputs.id != ''
 
+      - name: Store SDK version
+        run: |
+          python3 etc/update_api_version.py src/viam/versions.py sdk ${{ env.SDK_VERSION }}
+          echo "Updated SDK version to ${{ env.SDK_VERSION }} in versions.py"
+
```
purplenicole730, line 73: Other alternative that was being done before in case we don't like updating the version on release: getting the SDK version from reading `pyproject.toml` directly from `dial()`.

.github/workflows/release.yml
```diff
@@ -66,6 +66,11 @@ jobs:
         if: |
           steps.release_exists.outputs.id != ''
 
+      - name: Store SDK version
```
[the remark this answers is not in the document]
njooma: We definitely shouldn't be doing this. You an easily get the version of the python sdk by doing `import viam; viam.__version__`

njooma (changes requested):

stuqdog (approved):

.github/workflows/update_protos.yml
```diff
@@ -8,7 +8,7 @@ on:
 
 jobs:
   update-protos:
-    if: github.repository_owner == 'viamrobotics'
+    # if: github.repository_owner == 'viamrobotics'
```
njooma, line 11: Let's uncomment this now that testing is done

njooma (approved):

PR #696, opened 2024-07-30
RSDK-8303-3: use api version in makefile

purplenicole730:
Ensures that when a user runs `make buf` that they're generating the right API version.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,8 @@ typecheck:
 _buf: clean
 	rm -rf src/viam/gen
 	chmod +x plugin/main.py
-	buf generate buf.build/viamrobotics/api
+	$(eval API_VERSION := $(shell grep 'API_VERSION' src/viam/version_metadata.py | awk -F '"' '{print $$2}'))
+	buf generate buf.build/viamrobotics/api:v${API_VERSION}
 	buf generate buf.build/viamrobotics/goutils
 	protol -e googl* --in-place -s _grpc.py -s _pb2.py -s _pb2.pyi -o src/viam/gen buf buf.build/viamrobotics/api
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -23,7 +23,8 @@ typecheck:
 _buf: clean
 	rm -rf src/viam/gen
 	chmod +x plugin/main.py
-	buf generate buf.build/viamrobotics/api
+	$(eval API_VERSION := $(shell grep 'API_VERSION' src/viam/version_metadata.py | awk -F '"' '{print $$2}'))
```
stuqdog, line 26: I think there's one too many `$` in here? The command as written fails for me locally, but changing to `print $2` works.

purplenicole730: Oh that's interesting. It ran well for me. I just tried making that change and running it. It fails when I only have 1 `$`.

purplenicole730: From what I understand, we need two `$` since only using one means `make` thinks that it's a makefile variable.  `make` should be passing `$$2` as `$2` to `awk`. If this is true, I wonder why it's not working for you, and how it works with only one `$`. Could it be a difference in versions?

stuqdog: Interesting... Well I've been running just on the command line, if it works for you from the `makefile` then it's probably fine, thanks for following up :)

stuqdog (approved):

Makefile
purplenicole730: Oh, I've been testing by running `make buf` on the command line.
If you copy/paste `grep 'API_VERSION' src/viam/version_metadata.py | awk -F '"' '{print $$2}'` into the terminal, then it will fail. Is that what you did?

PR #699, opened 2024-07-31
Remove duplication

npentrel:
Can leave as is if this was intended but the duplication seems odd to me


stuqdog (commented):
This seems fine to me but I don't know if there's a particular reason if we had it the way we did before, I'd want to get @njooma's perspective on this.

njooma (approved):
Probably just there as a way to make sure people didn't miss it

stuqdog (approved):

PR #701, opened 2024-08-02
[APP-5754] Python GetFragmentHistory SDK changes

piokasar:
going to re-request when checks pass, requests got added immediately


src/viam/app/app_client.py
```diff
@@ -1781,13 +1852,48 @@ async def delete_fragment(self, fragment_id: str) -> None:
...
 
+    async def get_fragment_history(self,
+                                   id: str,
+                                   page_token: Optional[str] = "",
+                                   page_limit: Optional[int] = 10) -> list[FragmentHistoryEntry]:
+        """Get fragment history.
+
+        ::
+
+            await cloud.get_fragment_history(
```
stuqdog, line 1863: we should assign some value to this (`fragment_history = await cloud.get_fragment_history(...)`).

stuqdog (commented):
Generally looks good! A couple small things, and we'll want to make sure tests pass before we try to merge this.

Out of curiosity, why all the unrelated formatting changes? Was this something your editor did automatically?

Edit: oh I see you removed me from review until tests pass! Let me know if/when I should come back to look.

piokasar:
> Generally looks good! A couple small things, and we'll want to make sure tests pass before we try to merge this.
> 
> Out of curiosity, why all the unrelated formatting changes? Was this something your editor did automatically?
> 
> Edit: oh I see you removed me from review until tests pass! Let me know if/when I should come back to look.

thanks for looking anyways! I will re-request shortly

.idea/.gitignore
stuqdog, line 1: I think the `.idea` folder should probably not be checked in. Looks like it's pycharm specific? I'd suggest backing these files out and then adding `.idea/` to the `.gitignore`, or (ideally) to your own personal `.gitignore` (see https://stackoverflow.com/questions/7335420/global-git-ignore)

src/viam/app/app_client.py
```diff
@@ -8,10 +8,9 @@
 
 from viam import logging
 from viam.app._logs import _LogsStream, _LogsStreamWithIterator
+from viam.proto.app import AddRoleRequest, APIKeyWithAuthorizations, AppServiceStub
 from viam.proto.app import (
-    AddRoleRequest,
-    APIKeyWithAuthorizations,
-    AppServiceStub,
+    AuthenticatorInfo,
```
stuqdog, line 13: (minor) pretty weird that these two import lines are separated out! I know the `make format` suggested we do that, but I suspect that's because there was a `import Foo as FooPB` line in there to explain why. I'd suggest making this all one import statement again.

src/viam/app/app_client.py
```diff
@@ -383,6 +385,120 @@ def proto(self) -> FragmentPB:
         )
 
 
+# class AuthenticatorInfo:
```
stuqdog, line 388: Let's delete all of this commented out code.

src/viam/app/app_client.py
```diff
@@ -1781,6 +1897,38 @@ async def delete_fragment(self, fragment_id: str) -> None:
...
+        Raises:
+            GRPCError: if an invalid fragment id, page token or page limit is passed.
+
+        Returns:
+            viam.app.app_client.FragmentHistoryResponse: The fragment history document(s).
+
+        For more information, see `Fleet Management API <https://docs.viam.com/appendix/apis/fleet/>`_.
+        """
+
+        request: GetFragmentHistoryRequest = GetFragmentHistoryRequest(id=id, page_token=page_token, page_limit=page_limit)
```
stuqdog, line 1928: (minor) no need for the type annotation here, the use of the `GetFragmentHistoryRequest` constructor makes it obvious.

tests/package.json
stuqdog, line 1: I'm not sure what this is but I suspect we don't want/need it checked in.

tests/mocks/services.py
```diff
@@ -1490,6 +1493,12 @@ async def GetFragment(self, stream: Stream[GetFragmentRequest, GetFragmentRespon
         self.fragment_id = request.id
         await stream.send_message(GetFragmentResponse(fragment=self.fragment))
 
+    async def GetFragmentHistory(self, stream: Stream[GetFragmentHistoryRequest, GetFragmentHistoryResponse]) -> None:
+        request = await stream.recv_message()
+        assert request is not None
+        self.id = request.id
```
stuqdog, line 1499: Can we also set `self.page_token` and `self.page_id` here so we can confirm that those get passed correctly?

tests/test_app_client.py
```diff
@@ -36,6 +37,8 @@
 IDS = [ID]
 NAME = "name"
 CID = "cid"
+PAGE_TOKEN = ""
```
stuqdog, line 40: Can we put some other value here? The default `string` value is an empty string, so testing against that isn't necessarily going to tell us much.

tests/test_app_client.py
```diff
@@ -36,6 +37,8 @@
...
+PAGE_LIMIT = 10
```
stuqdog, line 41: Can we change this to a non-default value for testing purposes?

tests/test_app_client.py
```diff
@@ -630,6 +637,16 @@ async def test_delete_fragment(self, service: MockApp):
             await client.delete_fragment(fragment_id=ID)
             assert service.id == ID
 
+    @pytest.mark.asyncio
+    async def test_get_fragment_history(self, service: MockApp):
+        async with ChannelFor([service]) as channel:
+            client = AppClient(channel, METADATA, ID)
+            fragment_history = await client.get_fragment_history(id=ID, page_token="", page_limit=20)
```
stuqdog, line 644: We should set `page_token=PAGE_TOKEN` and `page_limit=PAGE_LIMIT`.

tests/test_app_client.py
```diff
@@ -630,6 +637,16 @@ async def test_delete_fragment(self, service: MockApp):
...
+            assert service.fragment.id == ID
```
stuqdog, line 645: Can we also add `assert service.id == ID` and similar for `page_token` and `page_limit`?

stuqdog (changes requested):
Generally looks good! Just some small suggestions and removing unneeded files from the commit.

tests/mocks/services.py
```diff
@@ -1173,6 +1174,9 @@ def __init__(
         available: bool,
         location_auth: LocationAuth,
         robot_part_history: List[RobotPartHistoryEntry],
+        fragment_history: List[FragmentHistoryEntry],
+        page_token: str,
```
piokasar, line 1178: in the proto these are optional - should they be optional here as well ?

stuqdog: I actually think we shouldn't make this a default part of the class at all. Currently we're setting them to our sentinel values when we create the service, and then passing those same sentinel values when we call the `get_fragment_history` method in the test, and then confirming that they're set to the sentinel value. But, because there's two places where the setting happens, we don't have confidence that the test actually did anything. 

What I'd suggest instead is just removing these as arguments or set values in the constructor, such that they don't exist in general. Then, we can leave everything else as it is such that the `GetFragmentHistory` method sets these values, and then we can meaningfully confirm that they're properly set in the `test_get_fragment_history` test.

tests/mocks/services.py
```diff
@@ -1184,6 +1188,7 @@ def __init__(
         items: List[RegistryItem],
         package_type: PackageType.ValueType,
     ):
+        self.page_token = None
```
stuqdog, line 1191: per above comment, this should be deleted

tests/mocks/services.py
```diff
@@ -1195,6 +1200,9 @@ def __init__(
         self.available = available
         self.location_auth = location_auth
         self.robot_part_history = robot_part_history
+        self.fragment_history = fragment_history
+        self.page_token = page_token
+        self.page_limit = page_limit
```
stuqdog, line 1205: per above comment, this should be deleted

stuqdog (commented):
Looking good! One small requested change to the `MockApp` class, otherwise lgtm :)

tests/mocks/services.py
piokasar: I got rid of page_token and page_limit but I kept fragment_history because that was the return value - does that make sense ? Apologies if I'm misunderstanding.

tests/mocks/services.py
piokasar: I'm not sure I'm understanding "such that they don't exist in general", but I ended up keeping them but setting to default values because when GetFragmentHistory tried to access them, the class attributes did not exist which gave me errors. Then in GetFragmentHistory, they were set based on the values in the request - it seems like this addressed your concern of setting twice, but if there's a more appropriate way to do this let me know!

tests/mocks/services.py
```diff
@@ -1195,6 +1197,9 @@ def __init__(
         self.available = available
         self.location_auth = location_auth
         self.robot_part_history = robot_part_history
+        self.fragment_history = fragment_history
+        self.page_token = ""
+        self.page_limit = 0
```
stuqdog, line 1202: (minor, opt) I'm pretty sure these aren't necessary/slightly surprised to hear you say you had errors if they weren't included. You _should_ be able to delete them, see e.g. how `self.public` is set in the `UpdateFragment` method, despite no `public` field ever being defined prior. 

That said, it's pretty harmless to have this here so it's not a big deal if you don't want to bother figuring out how to remove them.

tests/mocks/services.py
stuqdog: Curious what you mean by "tried to access them"? You should be able to just set to them (i.e., `self.some_fake_value = "this is a fake value that wasn't defined prior!"` should work with no issues), but if you tried to access the value with `foo = self.some_other_fake_value_that_has_not_been_defined` then yeah, I'd expect that to fail.

stuqdog (approved):
I still think we can remove the default definitions for `page_token` and `page_limit` but there's no major harm in including them so I totally understand if you'd rather just be done with the ticket. Otherwise, lgtm!

tests/mocks/services.py
piokasar: By access, I guess meant "set". But it must have been unrelated because I was able to get rid of those default values successfully, so thank you!
Once tests pass I will merge- thank you for your help on this!!

PR #704, opened 2024-08-07
Add MachineStatus to Robot API

ale7714:
https://viam.atlassian.net/browse/RSDK-7901

### DIFF
--- a/src/viam/robot/client.py
+++ b/src/viam/robot/client.py
@@ -907,7 +907,7 @@ class RobotClient:
 
         ::
 
-            machine_status = machine.get_machine_status()
+            machine_status = await machine.get_machine_status()
             resource_statuses = machine_status.resources
 
         Returns:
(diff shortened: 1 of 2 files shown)


src/viam/robot/client.py
```diff
@@ -894,3 +896,25 @@ async def get_version(self) -> GetVersionResponse:
...
+    # Get Machine Status #
+    ######################
+
+    async def get_machine_status(self) -> GetMachineStatusResponse:
+        """
+        Get status information about the robot.
+
+        ::
+
+            machine_status = machine.get_machine_status()
```
stuqdog, line 910: think we're missing an `await` here

stuqdog (approved):
One small thing (no need to re-request review from me) otherwise LGTM!

src/viam/robot/client.py
ale7714: good catch ty!

PR #706, opened 2024-08-12
RSDK-8440: fix license finder

purplenicole730:
Based on the available logs, this workflow has been failing for at least 3 months. Before that, the git logs have been removed, so we're not sure when the failure started.

`license_finder` has some issues with Python version 3.12, so now the workflow is using 3.11.5.
In the workflow, the command to run `license_finder` was changed so that errors fail the workflow instead of letting it succeed.

Things I tried that didn't work, wasn't the best solution, or introduced more errors:
- `pip install six` doesn't change anything
- `pip install pip==22.0` introduces another error
- `pip install pip==23.0` introduces another error
- `pip install pip==20.0.2` introduces another error
- `pip install pip==9.0.0` introduces another error
- `curl -sS https://bootstrap.pypa.io/get-pip.py | python3` doesn't change anything
- `python version 3.11` doesn't change anything
- `pip install pyenv` doesn't change anything
-  `/opt/hostedtoolcache/Ruby/*/x64/lib/ruby/gems/*/gems/license_finder-*/bin/license_finder_pip.py requirements.txt` doesn't actually run license_finder
- `poetry run license_finder --python-version=3` fails even after using python v3.11.5
- `poetry run license_finder` fails even after using python v3.11.5

### DIFF
--- a/.github/workflows/license_finder.yml
+++ b/.github/workflows/license_finder.yml
@@ -39,7 +39,5 @@ jobs:
 
       - name: Run license finder and capture output
         run: |
-          rm -rf ~/.local/share/virtualenvs
-          pipx install pipenv
-          pipenv --version
+          pip
           /opt/hostedtoolcache/Ruby/3.3.4/x64/lib/ruby/gems/3.3.0/gems/license_finder-7.2.1/bin/license_finder_pip.py requirements.txt
--- a/.github/workflows/license_finder.yml
+++ b/.github/workflows/license_finder.yml
@@ -40,5 +40,5 @@ jobs:
 
       - name: Run license finder
         run: |
-          license_finder
-          echo "Exit code: $?"
+          /opt/hostedtoolcache/Ruby/*/x64/lib/ruby/gems/*/gems/license_finder-*/bin/license_finder_pip.py requirements.txt
+          poetry run license_finder --python-version=3
--- a/.github/workflows/license_finder.yml
+++ b/.github/workflows/license_finder.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v5
         with:
-          python-version: "3.11.5"
+          python-version: "3.11"
 
       - uses: ruby/setup-ruby@v1
         with:
@@ -40,5 +40,4 @@ jobs:
 
       - name: Run license finder
         run: |
-          /opt/hostedtoolcache/Ruby/*/x64/lib/ruby/gems/*/gems/license_finder-*/bin/license_finder_pip.py requirements.txt
           license_finder
--- a/.github/workflows/license_finder.yml
+++ b/.github/workflows/license_finder.yml
@@ -39,4 +39,4 @@ jobs:
 
       - name: Run license finder and capture output
         run: |
-          ./.gem/ruby/*/gems/license_finder-*/bin/license_finder_pip.py requirements.txt
+          /opt/hostedtoolcache/Ruby/3.3.4/x64/lib/ruby/gems/3.3.0/gems/license_finder-7.2.1/bin/license_finder_pip.py requirements.txt
--- a/.github/workflows/license_finder.yml
+++ b/.github/workflows/license_finder.yml
@@ -41,4 +41,4 @@ jobs:
       - name: Run license finder
         run: |
           /opt/hostedtoolcache/Ruby/*/x64/lib/ruby/gems/*/gems/license_finder-*/bin/license_finder_pip.py requirements.txt
-          poetry run license_finder
+          license_finder
--- a/.github/workflows/license_finder.yml
+++ b/.github/workflows/license_finder.yml
@@ -39,7 +39,4 @@ jobs:
 
       - name: Run license finder and capture output
         run: |
-          poetry run license_finder --python-version=3 > license_finder_output.txt 2>&1 || true
-
-      - name: Display output
-        run: cat license_finder_output.txt
+          ~/.gem/ruby/*/gems/license_finder-*/bin/license_finder_pip.py requirements.txt
--- a/.github/workflows/license_finder.yml
+++ b/.github/workflows/license_finder.yml
@@ -40,5 +40,5 @@ jobs:
 
       - name: Run license finder and capture output
         run: |
-          pip install six
+          pip install --upgrade pip==22.0.2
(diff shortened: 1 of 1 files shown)


.github/workflows/license_finder.yml
```diff
@@ -36,7 +36,8 @@ jobs:
       - name: Generate requirements.txt (exclude dev dependencies)
         run: |
           poetry export -f requirements.txt --without-hashes > requirements.txt
+          pip install -r requirements.txt
 
       - name: Run license finder
         run: |
-          poetry run license_finder --python-version=3
+          /opt/hostedtoolcache/Ruby/*/x64/lib/ruby/gems/*/gems/license_finder-*/bin/license_finder_pip.py requirements.txt
```
stuqdog, line 43: I think we need to invoke this via `python3`. It looks like in the failed jobs it's `python3 /path/to/license_finder_pip.py /path/to/requirements.txt`, and in successful runs in other SDKs the output is different from here (it should say `All dependencies are approved for use`).

stuqdog (changes requested):

.github/workflows/license_finder.yml
purplenicole730: Let me try that!

purplenicole730: You're absolutely right about the expected output! Unfortunately using python3 was not the method. It was the script itself, so I made that fix. Thanks for catching that!

stuqdog (approved):
Nice! Looks good and the license_finder check succeeds 🎉

.github/workflows/license_finder.yml
purplenicole730: I realized that we still need this command to check license finder and fail the workflow if needed. I checked, and for some reason `license_finder` returns a exit status of 0 even though it's failing.

.github/workflows/license_finder.yml
```diff
@@ -36,7 +36,9 @@ jobs:
...
+          poetry run license_finder
```
purplenicole730, line 44: Do we need `poetry run` here? It seems to pass just fine if we simply call `license_finder`. For some reason it fails if we use `--python-version=3`

.github/workflows/license_finder.yml
```diff
@@ -36,7 +36,9 @@ jobs:
       - name: Generate requirements.txt (exclude dev dependencies)
         run: |
           poetry export -f requirements.txt --without-hashes > requirements.txt
+          pip install -r requirements.txt
 
       - name: Run license finder
         run: |
-          poetry run license_finder --python-version=3
+          /opt/hostedtoolcache/Ruby/*/x64/lib/ruby/gems/*/gems/license_finder-*/bin/license_finder_pip.py requirements.txt
```
purplenicole730, line 43: We still need this command to check license finder and fail the workflow if needed. Running license finder itself returns an exit status of 0 even if it's failing.

.github/workflows/license_finder.yml
```diff
@@ -36,7 +36,9 @@ jobs:
...
+          license_finder
```
purplenicole730, line 44: This command exits with an exit code of 0, which is why running `license_finder_pip.py` in the previous step is still necessary.

stuqdog: Huh, that's interesting. How did you test it? I just ran `license_finder` locally and got an exit code of 1 with unknown dependencies.

purplenicole730: I did it before [here](https://github.com/viamrobotics/viam-python-sdk/actions/runs/10405240710/job/28815550358).

In case it was me doing it wrong, I tried it again but it [failed without actually failing the workflow again](https://github.com/viamrobotics/viam-python-sdk/actions/runs/10406128924/job/28818455686?pr=706).

stuqdog: This looks like a slightly different failure, where it doesn't even recognize what the dependencies _are_. I'm guessing the linked failures are not running on `3.11.5`? 
In cases where `license_finder` can successfully see dependencies and run, I'd expect that we should see an appropriate failure. In local testing in the flutter sdk e.g., I get the following output

```
LicenseFinder::Pub: [32mis active for '/Users/ethanrodkin/Documents/code/viam/viam-flutter-sdk'[0m

Dependencies that need approval:
_fe_analyzer_shared, 67.0.0, unknown
analyzer, 6.4.1, unknown
...
```

and running `echo $?` returns `1`.

purplenicole730: I'm not sure what you mean by slightly different failure. `ModuleNotFoundError: No module named 'pip._vendor.six'` is the error we've been getting in the Python SDK.

I forgot to mention, I did print out the exit code [here](https://github.com/viamrobotics/viam-python-sdk/actions/runs/10405355402/job/28815903802), and you can see running returns 0 even though it failed.

stuqdog: > I'm not sure what you mean by slightly different failure. `ModuleNotFoundError: No module named 'pip._vendor.six'` is the error we've been getting in the Python SDK.

So what I mean is, there are broadly three classes of response here:
1. `license_finder` runs successfully, all licenses are a-okay and we return `0` (good!)
2. `license_finder` runs successfully, one or more licenses are unknown or invalid and we return `1` (for the purposes of this test, also good!)
3. we get the `ModuleNotFound` error that you have encountered and return `0` (bad!)

My claim is that we want to get away from case 3. When I say "slightly different failure" what I mean is, you're pointing to the case 3 failure, but I am describing encountering a "slightly different failure" of case 2 locally.

The reason I ask about being on `3.11.5` is because, when we had python at `3.11.5` and just ran `license_finder` without anything special, we found ourselves in case 1. When moving to a different version of python, we get into case 3. This makes me believe that other versions of python are creating inconsistencies that prevent `license_finder` from running properly; I believe this is supported by the fact that the error we are seeing is a `pip3` error, not a `license_finder` one.

Therefore, I believe that being on `3.11.5` and just running `license_finder`, which puts us in case 1, is the ideal state and what we should be merging. I believe if in that state you were to add an unlicensed dependency in `requirements.txt` and run the test, you would find yourself in case 2, _not_ in case 3.

purplenicole730: This makes sense to me. Thanks for taking the time to explain that, Ethan!

njooma (approved):

PR #716, opened 2024-08-21
RSDK-8553-RSDK-8552: create python module template

purplenicole730:
[Jira ticket](https://viam.atlassian.net/browse/RSDK-8553)

This ticket is the start of writing the module templates, with the first being in the Python SDK. Later, this will move on to another repo. A lot of work still has to be done, but those are out of scope for this work.

The `cookiecutter` tool and package is being used to create a Python module after prompting a user to give necessary information.

An example run is given in the `README.md`. The `README.md` is a work in progress, and is specific to how it's being done now. It will change significantly as more work is done on this epic, but it is a start.
_(For reviewers: To see a preview of the README, go to the `Files changed` tab, scroll to the README, click the menu button on the top right, and click `View file`.)_
image


module_generator/README.md
```diff
@@ -0,0 +1,220 @@
+# Module Generator Template
+
+This is the Python module template for [CLI module generator](https://docs.viam.com/cli/#module).
+
+## How to Use
+
+1. Install cookiecutter: `pipx install cookiecutter`. If pipx is not an option, use `python -m pip install --user cookiecutter`.
```
HipsterBrown, line 7: It isn't necessarily required to install cookiecutter with `pipx` for a one-off command by using `pipx run cookiecutter` instead: https://github.com/pypa/pipx?tab=readme-ov-file#walkthrough-running-an-application-in-a-temporary-virtual-environment

> For example, the blog post [How to set up a perfect Python project](https://sourcery.ai/blog/python-best-practices/) uses pipx run to kickstart a new project with [cookiecutter](https://github.com/cookiecutter/cookiecutter), a tool that creates projects from project templates.

This can be accomplished in one command without installing the viam-python-sdk:

```console
pipx run cookiecutter gh:viamrobotics/viam-python-sdk --directory module_generator/cookiecutter-template
```

module_generator/README.md
```diff
@@ -0,0 +1,220 @@
...
+2. Call the template: `python3 module_generator/cookiecutter-template`.
+3. Fill out the prompts given (see below).
+4. The necessary files for the created module will be generated in a new directory `module_generator/<MODULE_NAME>`.
+
+NOTE: The imports in `module_generator/<MODULE_NAME>/src/__main__.py` may have to be adjusted. More details below in the example.
+
+## Prompts
+
+To create a module, the generator needs the user to fill out information about the module. The prompts can be seen in `cookiecutter-template/cookiecutter.json`. The public prompts that a user directly affects are as follows:
```
HipsterBrown, line 16: Can we use the `__prompts__` configuration field to show the descriptive text for each prompt listed here? https://cookiecutter.readthedocs.io/en/stable/advanced/human_readable_prompts.html#human-readable-prompts

module_generator/cookiecutter-template/hooks/post_gen_project.py
```diff
@@ -0,0 +1,12 @@
+import sys
+import subprocess
+
+
+if __name__ == "__main__":
+    subprocess.run(
```
HipsterBrown, line 6: I ran into an issue when trying this out due to the `post_gen_project` command running in the root directory of the generated project, so the `../generate_main.py` couldn't be found when generating a project outside of the `viam-python-sdk` directory: https://cookiecutter.readthedocs.io/en/stable/advanced/hooks.html#hook-execution

Could the `generate_main.py` be included in the `hooks` directory to be called without shelling out to execute?

module_generator/cookiecutter-template/{{cookiecutter.module_name}}/README.md
```diff
@@ -0,0 +1,5 @@
+# {{cookiecutter.module_name}} Module
```
HipsterBrown, line 1: Can we use the recommended README from the docs here? https://docs.viam.com/how-tos/create-module/#optional-create-a-readme

module_generator/generate_main.py
```diff
@@ -0,0 +1,103 @@
...
+            config (ComponentConfig): The new configuration
+            dependencies (Mapping[ResourceName, ResourceBase]): Any dependencies (both implicit and explicit)
+        """
+        return super().reconfigure(config, dependencies)
+
+    {method_stubs}
+
+if __name__ == "__main__":
+    """This function will run your module and make it available to the machine."""
+    asyncio.run(Module.run_from_registry())
```
HipsterBrown, line 91: When does `Registry.register_resource_creator` get run in order for `Module.run_from_registry` to pull this model from the registry?

HipsterBrown (commented):
Really excited to see this becoming built into the SDK (and maybe part of the CLI in the future)! 

I left some feedback after trying this out locally. I think it provides a lot of good defaults overall. 

As a broader question, almost all Python project generators (pdm, hatch, uv, poetry) will create a `pyproject.toml` as the standard file for managing dependencies and other project metadata. Should this generator also opt for that standard file over the bare `requirements.txt` file?

module_generator/README.md
purplenicole730: Very cool know that running it in one line is possible, but I am hitting an error when I try this command.

That being said, I don't think we'll be running the template with cookiecutter in the first place. We'll actually be calling it through the CLI, but this is just a beginning to set up the template.

module_generator/README.md
purplenicole730: Good suggestion! I almost missed that, although I'm not sure how many people will really see this since it's going to be run in the CLI. Doesn't hurt to add though :)

module_generator/cookiecutter-template/hooks/post_gen_project.py
purplenicole730: How did you generate a project outside of the `viam-python-sdk` directory? Right now, the way I set up the project, it should be generated within the `viam-python-sdk/cookiecutter-template` directory. Eventually, the end goal is to initialize a git repository from the template. This is the [epic](https://viam.atlassian.net/browse/RSDK-8550) if you'd like to see it.
When I put `generate_main.py` in the `hooks` directory and try to import the function, the script can't find the `generate_main` file.

purplenicole730: Ah, I just tried it while in the `viam-python-sdk` directory. I hate relative paths for imports 😩

module_generator/README.md
```diff
@@ -0,0 +1,220 @@
...
+`module_generator/rover/.viam-gen-info`
+
+```json
+{
+  "module_name": "rover",
+  "language": "python",
+  "namespace": "test",
+  "resource": "vision service",
+  "model_name": "vizz",
+  "generator_version": "0.14.0",
```
stuqdog, line 67: (q) Just out of curiosity, where this `generator_version` come from? It looks like we try to use the SDK version and fallback on `0.1.0`, how did we end up with `0.14.0`?

module_generator/README.md
HipsterBrown: Who is this README for if this will be called through the CLI? 

>  since this PR isn't merged yet, the template isn't available through github
`cookiecutter` accepts a `checkout` flag to get templates from alternative branches in a git repo. I actually tested this template by running:

```console
pipx run cookiecutter gh:purplenicole730/viam-python-sdk --checkout RSDK-8553-create-python-module-template --directory module_generator/cookiecutter-template
```

module_generator/README.md
HipsterBrown: Is the CLI going to prompt with a different interface and pass those arguments to this template generator?

module_generator/generate_main.py
purplenicole730: Right now I guess it doesn't. I haven't actually tested the module itself. Thanks for catching that!

module_generator/cookiecutter-template/{{cookiecutter.module_name}}/README.md
purplenicole730: Oh I didn't even realize there was a template! Thank you

module_generator/README.md
```diff
@@ -0,0 +1,220 @@
...
+from viam.module.module import Module
+from viam.proto.app.robot import ComponentConfig
+from viam.proto.common import ResourceName
+from viam.resource.base import ResourceBase
+from viam.resource.easy_resource import EasyResource
+from viam.resource.types import Model, ModelFamily
+from viam.services.vision import Vision
+
+
+class Vizz(Vision, EasyResource):
```
stuqdog, line 117: We're not inheriting from `Reconfigurable` here but we expose the `reconfigure` method below. This is kind of a tricky case where I think some manual user input will be hard to avoid, _but_ if users define a `reconfigure` method I think it's pretty important to derive from `Reconfigurable`, otherwise `reconfigure_resource` calls won't behave correctly (`reconfigure_resource` checks if a resource `isinstance` of `Reconfigurable`, and I don't believe the `isinstance` method uses duck typing). 

As an aside, might be worth considering getting rid of the `Reconfigurable` type at all and wrapping the call to `reconfigure` in a `try/except` block. but that's a separate conversation and a separate ticket.

module_generator/README.md
purplenicole730: From what I know so far, we're not sure how we're going to tie things together yet. We'll be working through it and having meetings about it in the coming month.

module_generator/README.md
purplenicole730: Right now, it was more of a way to document everything that was happening for me and the SDK team. Things are still up in the air, but I think there's a big chance we'll be moving this to its own repo since we also want this for other languages. We decided to just include this template to the python sdk for ease of testing for the first run.

module_generator/README.md
stuqdog: some details are TBD, but the expectation is that the CLI will wrap this `cookiecutter` code by creating an execuable via `pyinstaller`. I'd be pretty surprised if we moved away from that at a high level. So, the prompts here will be seen in the CLI.

module_generator/README.md
```diff
@@ -0,0 +1,220 @@
...
+    """This is your custom Vision service.
+
+    All the methods listed are optional.
+    You can implement the methods you want and remove the methods you don't need.
```
stuqdog, line 121: I'm not sure if it's great to recommend that people can just remove methods. Technically the method is _not_ optional; the abstract base class defines that these methods are what makes something a `Vision` service. Given how python works it's okay to remove the method, but if, e.g., a `stop` method was removed and someone tried to call it they would get fallback to the definition in the ABC, which is basically just `pass`. If we leave it as-is here, we'd get a `NotImplementedError`. The latter seems better to me imo, otherwise it would be easy for someone to acting in reliance on a method that doesn't actually do anything.

module_generator/cookiecutter-template/cookiecutter.json
```diff
@@ -0,0 +1,76 @@
+{
+  "module_name": "MyModule",
+  "language": ["python"],
+  "namespace": "ns",
+  "resource": [
```
stuqdog, line 5: maybe more for a v2, but I think when we're looking at all these resource types we end up with a potentially overwhelming list in the user interface. I wonder if some sort of fuzzy finding support is possible?

module_generator/cookiecutter-template/{{cookiecutter.module_name}}/requirements.txt
```diff
@@ -0,0 +1 @@
+viam-sdk==<CURRENT VERSION>
```
stuqdog, line 1: should we add here also python version limitations? Since we don't support before 3.8.

stuqdog (changes requested):
Generally looking good, haven't tested yet but here are some first pass thoughts!

module_generator/README.md
stuqdog: @HipsterBrown We're actually actively working on CLI integration currently so I think it's best to hold off on encouraging use until that's merged. My concern is that the location of this code will likely change (at a minimum the `cookiecutter` code probably will end up in RDK, and the templates could possibly move as well) and encouraging users to interact with it before then could lead to unpleasant user experiences when workflows are suddenly changed from under them. Also, the amount of time where 1) all of this is merged into the Python SDK, but 2) we don't yet have CLI integration, will be on the order of weeks, not months, so I don't think we're holding people back for a significant amount of time.

Let me know if this doesn't sound reasonable to you; happy to chat offline if you want!

HipsterBrown: Thanks for breaking that down. I think having the disclaimer [you suggested](https://github.com/viamrobotics/viam-python-sdk/pull/716#discussion_r1727198707) will be helpful for anyone who happens upon until it is fully ready. 

Given the eventual inclusion in the RDK codebase, I'm curious if a go-based scaffolding library was explored before settling on cookiecutter. But that can be discussed outside of this PR.

module_generator/README.md
```diff
@@ -0,0 +1,220 @@
...
+This is the Python module template for [CLI module generator](https://docs.viam.com/cli/#module).
+
+## How to Use
+
+1. Install cookiecutter: `pipx install cookiecutter`. If pipx is not an option, use `python -m pip install --user cookiecutter`.
+2. Call the template: `python3 module_generator/cookiecutter-template`.
+3. Fill out the prompts given (see below).
+4. The necessary files for the created module will be generated in a new directory `module_generator/<MODULE_NAME>`.
+
+NOTE: The imports in `module_generator/<MODULE_NAME>/src/__main__.py` may have to be adjusted. More details below in the example.
```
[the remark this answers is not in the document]
purplenicole730: I would also really enjoy this. I'm sure there is a way to get this done, but I'll have to look at how

module_generator/README.md
purplenicole730: Should I be removing the `reconfigure` method for now?

module_generator/README.md
purplenicole730: This makes sense. This was copied and pasted from another template, so I'll just erase the `remove the methods you don't need` part.

module_generator/cookiecutter-template/cookiecutter.json
purplenicole730: I personally had no issue with this, as it gives a list of the components/services that are available. Although you're right that this can get messy easily if the list gets bigger than what it is now, or maybe it already is too big. Right now, it looks like this:
image

(Had to look up what fuzzy search was) If we want to implement a fuzzy search, I do agree that should be a v2.

module_generator/README.md
stuqdog: > Given the eventual inclusion in the RDK codebase, I'm curious if a go-based scaffolding library was explored before settling on cookiecutter.

It was! We looked at a go-based scaffolding library but the difference was `cookiecutter` has tens of thousands of stars and is still maintained, whereas the go library had dozens of stars and (iirc) hadn't been committed to in quite a while. So I think we chose `cookiecutter` because the ecosystem seems more robust and we're less likely to have to maintain it ourselves.

module_generator/cookiecutter-template/{{cookiecutter.module_name}}/requirements.txt
purplenicole730: `requirements.txt` is more for package dependencies, so from what I know, you can't specify a python version here. I think we'd need a `pyproject.toml` file for that.

module_generator/README.md
purplenicole730: I looked into `EasyResource`, and this class has a `reconfigure` method. Does that mean that every `EasyResource` potentially doesn't work if a descending class calls this function?

module_generator/README.md
```diff
@@ -0,0 +1,263 @@
...
+    {
+      "api": "rdk:component:vision",
+      "model": "test:rover:vizz"
+    }
+  ],
+  "entrypoint": "run.sh"
+}
+```
+
+`module_generator/rover/README.md`
```
npentrel, line 92: Thank you for adding this here 💖

njooma:
Closing this PR since we moved this functionality to the RDK

PR #717, opened 2024-08-22
remove requirements-test.txt

stuqdog:
This file was added almost a year ago and hasn't been touched, doesn't look like it's used anywhere in the SDK. pretty sure it was added by mistake so let's get rid of it!

Happy to close the PR if it turns out there's a reason for the file to be checked in.


stuqdog:
haha just kidding! turns out it is necessary, forgot to pass the `--hidden` flag to `rg` 🙃

PR #740, opened 2024-09-19
RSDK-8813: Add frame_rate to python sdk 

jckras:
Blocked by the API changes being merged (API PR #553)
- Added in support for `frame_rate` to the GetProperties() in service.py 
- Added in a fake `frame_rate `in the MockCamera for testing and added a check for `frame_rate` in test_camera.py


src/viam/version_metadata.py
jckras, line 1: When trying to compile locally I had to remove the "v" from "v0.1.340" because there was another "v" being added in the makefile - not sure if this will cause a problem with github

jckras:
Q: I am having trouble getting the tests to pass since the API changes do not seem to be showing up/the new proto files are not being used. How can I fix this?

src/viam/components/camera/service.py
```diff
@@ -89,6 +89,7 @@ async def GetProperties(self, stream: Stream[GetPropertiesRequest, GetProperties
             intrinsic_parameters=properties.intrinsic_parameters,
             distortion_parameters=properties.distortion_parameters,
             mime_types=properties.mime_types,
+            frame_rate=properties.frame_rate,
```
njooma, line 92: You can actually delete this entire assignment. `properties` is actually a `GetPropertiesResponse`, so you should be able to remove all of this entirely

njooma (changes requested):

src/viam/components/camera/service.py
jckras: got it. in this case should the other property assignments be removed to?

njooma: Yup! You can actually get rid of the entire `response = GetPropertiesResponse...` bit and simply do `await stream.send_message(properties)`

njooma (approved):

PR #745, opened 2024-09-25
Update data_client.py

npentrel:
it seems like some formatting issues snuck in
Screenshot 2024-09-25 at 15 11 57

cc: @sguequierre

### DIFF
--- a/src/viam/app/data_client.py
+++ b/src/viam/app/data_client.py
@@ -176,6 +176,7 @@ class DataClient:
         If the file is not empty, it will be overwritten.
 
         ::
+            from viam.utils import create_filter
 
             my_data = []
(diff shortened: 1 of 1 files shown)


src/viam/app/data_client.py
```diff
@@ -176,6 +176,7 @@ async def tabular_data_by_filter(
         If the file is not empty, it will be overwritten.
 
         ::
+            from viam.utils import create_filter
```
npentrel, line 179:
```suggestion

            from viam.utils import create_filter
```

njooma (approved):

sguequierre (approved):
oops! thanks naomi!

PR #747, opened 2024-09-26
RSDK-6169 - gRPC debugging client api

stuqdog:
Adds the gRPC debugging client api.

Note to reviewers: a lot of this is pretty boilerplate. I certainly won't complain if you look at all of it but I think the main places to focus are `rpc_client_base.py` and `utils.py`

Flyby: refactors `dict_to_struct` to make the obj argument type optional, such that we don't need to have `if extra is None: extra = {}` in every method.

Example code:

```python
robot = await connect()
motor = Motor.from_robot(robot, "<my-motor-name>")
# only a single metadata need be created, which can then be passed to any number of methods
# from any number of resources, provided the user is happy with the logs sharing the same key
md = motor.Metadata()
md.enable_debug_logging('my-cool-key') # the key is optional; if not provided, a random key will be generated
await motor.go_for(rpm=100, revolutions=5, metadata=md)
await robot.close()
```

example logs: 

Screenshot 2024-09-26 at 11 43 18

### DIFF
--- a/src/viam/components/arm/client.py
+++ b/src/viam/components/arm/client.py
@@ -39,12 +39,11 @@ class ArmClient(Arm, ReconfigurableResourceRPCClientBase):
         *,
         extra: Optional[Dict[str, Any]] = None,
         timeout: Optional[float] = None,
-        **__,
+        **kwargs,
     ) -> Pose:
-        if extra is None:
-            extra = {}
+        md = kwargs.get('metadata', self.Metadata()).proto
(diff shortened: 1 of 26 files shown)


stuqdog:
@dgottlieb  fyi

dgottlieb:
This makes sense to me, thanks for linking the PR!

purplenicole730 (commented):
Kind of skimmed over the components/services since the changes seemed straightforward. Thank you for the flyby change!

purplenicole730 (approved):

PR #748, opened 2024-09-30
DOCS-2902: Edit machine management api from QA

sguequierre:


src/viam/robot/client.py
```diff
@@ -81,19 +81,20 @@ class RobotClient:
...
+        async def connect(address) -> RobotClient:
+            opts = RobotClient.Options(
+                disable_sessions=True, dial_options=DialOptions(timeout=10)).with_api_key(
+                    # Replace "<API-KEY>" (including brackets) with your API key
+                    api_key='<API-KEY>',
+                    # Replace "<API-KEY-ID>" (including brackets) with your API key
+                    # ID
+                    api_key_id='<API-KEY-ID>'
+                )
+            return await RobotClient.at_address(address=address, options=opts)
```
npentrel, line 93:
```suggestion
            return await RobotClient.at_address(address="<ADDRESS-FROM-THE-VIAM-APP>", options=opts)
```

src/viam/robot/client.py
```diff
@@ -81,19 +81,20 @@ class RobotClient:
...
 
         async def main():
             # Make a RobotClient
-            machine = await connect()
+            machine = await connect('ADDRESS FROM CODE SAMPLE TAB OF VIAM APP')
```
npentrel, line 97: I did this in a code sample I sent you because I needed to connect to many machines and so was passing that in as an argument, that's not needed here
```suggestion
            machine = await connect()
```

src/viam/robot/client.py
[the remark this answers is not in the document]
sguequierre: I should do that and remove the below? Sorry, I thought this way was your recommendation!

src/viam/robot/client.py
```diff
@@ -710,7 +712,24 @@ async def transform_pose(
...
+                o_z=0.0,  # Orientation radian Z
+                theta=0.0 # Orientation angle in radians
+            )
+
+            pose_in_frame = PoseInFrame(
+                reference_frame="world",
+                pose=pose
+            )
+
+            pose = await machine.transform_pose(pose_in_frame, "world")
```
npentrel, line 732: should this overwrite the pose? Or should the variable be called something else?

src/viam/robot/client.py
```diff
@@ -738,25 +757,29 @@ async def discover_components(
...
+        Get a list of discovered potential component configurations.
+        Only implemented for webcam cameras in builtin components. Returns module names for modules.
 
         ::
 
+            from viam.proto.robot import DiscoveryQuery
+
             # Define a new discovery query.
-            q = machine.DiscoveryQuery(subtype=acme.API, model="some model")
+            q = DiscoveryQuery(subtype="camera", model="webcam")
```
npentrel, line 768: good work

src/viam/robot/client.py
```diff
@@ -738,25 +757,29 @@ async def discover_components(
...
 
             # Define a list of discovery queries.
             qs = [q]
 
             # Get component configurations with these queries.
             component_configs = await machine.discover_components(qs)
+            print(component_configs)
```
npentrel, line 775: I think we don't generally add print statements to the docs
```suggestion
```

src/viam/robot/client.py
```diff
@@ -738,25 +757,29 @@ async def discover_components(
         queries: List[DiscoveryQuery],
     ) -> List[Discovery]:
         """
-        Get the list of discovered component configurations.
+        Get a list of discovered potential component configurations.
```
npentrel, line 760: I can't say I understand what potential means in this context. Do you know? Can you explain?

src/viam/robot/client.py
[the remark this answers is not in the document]
npentrel: yep - accept both suggestions

src/viam/robot/client.py
```diff
@@ -710,7 +714,24 @@ async def transform_pose(
...
+                o_z=0.0,  # Orientation radian Z
+                theta=0.0 # Orientation angle in radians
+            )
+
+            pose_in_frame = PoseInFrame(
+                reference_frame="world",
+                pose=pose
+            )
+
+            pose = await machine.transform_pose(pose_in_frame, "world")
```
sguequierre, line 734:
```suggestion
            transformed_pose = await machine.transform_pose(pose_in_frame, "world")
```

src/viam/robot/client.py
```diff
@@ -81,14 +81,17 @@ class RobotClient:
...
 
-        async def connect():
-            opts = RobotClient.Options.with_api_key(
-                # Replace "<API-KEY>" (including brackets) with your machine's API key
-                api_key='<API-KEY>',
-                # Replace "<API-KEY-ID>" (including brackets) with your machine's API key ID
-                api_key_id='<API-KEY-ID>'
-            )
-            return await RobotClient.at_address('<ADDRESS-FROM-THE-VIAM-APP>', opts)
+        async def connect(address) -> RobotClient:
```
sguequierre, line 84:
```suggestion
        async def connect() -> RobotClient:
```

src/viam/robot/client.py
sguequierre: rand explained it to me, its basically a long list of all the discovered possible configurations for the camera. like height this many pixels width this many pixels this webcam path or this webcam path etc. can confirm thats what the printout looks like and she approved of this terminology

src/viam/robot/client.py
```diff
@@ -738,12 +759,15 @@ async def discover_components(
         queries: List[DiscoveryQuery],
     ) -> List[Discovery]:
         """
-        Get the list of discovered component configurations.
+        Get a list of discovered potential component configurations.
+        Only implemented for webcam cameras in builtin components. Returns module names for modules.
```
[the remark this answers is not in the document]
sguequierre: only the webcam out of the builtin components, nothing else has it. Your suggestion seems accurate though

src/viam/robot/client.py
```diff
@@ -81,14 +81,17 @@ class RobotClient:
...
-            opts = RobotClient.Options.with_api_key(
-                # Replace "<API-KEY>" (including brackets) with your machine's API key
-                api_key='<API-KEY>',
-                # Replace "<API-KEY-ID>" (including brackets) with your machine's API key ID
-                api_key_id='<API-KEY-ID>'
-            )
-            return await RobotClient.at_address('<ADDRESS-FROM-THE-VIAM-APP>', opts)
+        async def connect() -> RobotClient:
+            opts = RobotClient.Options(
+                disable_sessions=True,
```
njooma, line 86: Why are we suggesting disabling sessions?

src/viam/robot/client.py
```diff
@@ -738,12 +759,15 @@ async def discover_components(
         queries: List[DiscoveryQuery],
     ) -> List[Discovery]:
         """
-        Get the list of discovered component configurations.
+        Get a list of discovered potential component configurations for webcam cameras, for example listing different supported resolutions.
```
njooma, line 762: Does discovery only work for webcam cameras? If no, this seems to suggest that

src/viam/robot/client.py
njooma: For `subtype`, you could do `Camera.SUBTYPE.resource_subtype` which is a bit more verbose but fully typechecked.

This way also works, but just in case people don't know how to get the subtype? Up to y'all, just showing an additional way of doing it

src/viam/robot/client.py
sguequierre: Yeah it only works for webcam camera re. rand

src/viam/robot/client.py
sguequierre: I don't know, I assume it was a copy paste error but just to double check @npentrel do you know if there was a reason for this? If not I will make a pr to edit on main docs page too.

src/viam/robot/client.py
sguequierre: This way seems more straightforward but thank you

src/viam/robot/client.py
npentrel: Where did you get this code from? Was that from [here](https://docs.viam.com/appendix/apis/robot/#establish-a-connection)?

npentrel (changes requested):

src/viam/robot/client.py
```diff
@@ -101,6 +101,7 @@ async def main():
         if __name__ == '__main__':
             asyncio.run(main())
 
+
```
sguequierre, line 104:
```suggestion
```

src/viam/robot/client.py
sguequierre: From the code sample you sent

npentrel (approved):

src/viam/robot/client.py
njooma: Is there a chance it will work for other components in the future? I don't want to have to be overly specific in the event that we just haven't gotten to implementing it elsewhere just yet but plan to

src/viam/robot/client.py
```diff
@@ -741,12 +759,15 @@ async def discover_components(
         queries: List[DiscoveryQuery],
     ) -> List[Discovery]:
         """
-        Get the list of discovered component configurations.
+        Get a list of discovered potential component configurations for webcam cameras, for example listing different supported resolutions.
```
npentrel, line 762: Let's go with something like this. As anything does get implemented we'll have to update it. But I can't think of usecases for it. For webcams it is because it suggests relevant resolutions. I can't think of a similar thing for other components
```suggestion
        Get a list of discovered potential component configurations, for example listing different supported resolutions. Currently only works for some cameras.
```

njooma (approved):

PR #751, opened 2024-10-01
Fix arm, base, camera snippets

JessamyT:
DOCS-2779

### DIFF
--- a/src/viam/components/camera/camera.py
+++ b/src/viam/components/camera/camera.py
@@ -48,7 +48,10 @@ class Camera(ComponentBase):
 
         ::
 
-            my_camera = Camera.from_robot(robot=robot, name="my_camera")
+           # from viam.media.video import CameraMimeType    
+           # Add the above to imports at top of file!
+        
+            my_camera = Camera.from_robot(robot=machine, name="my_camera")
 
             # Assume "frame" has a mime_type of "image/vnd.viam.dep"
             frame = await my_camera.get_image(mime_type = CameraMimeType.VIAM_RAW_DEPTH)
@@ -74,10 +77,10 @@ class Camera(ComponentBase):
 
         ::
 
-            my_camera = Camera.from_robot(robot=robot, name="my_camera")
+            my_camera = Camera.from_robot(robot=machine, name="my_camera")
 
             images, metadata = await my_camera.get_images()
-            img0 = images[0].image
+            image_list = images[0]
             timestamp = metadata.captured_at
 
         Returns:
@@ -105,7 +108,7 @@ class Camera(ComponentBase):
             import numpy as np
             import open3d as o3d
 
-            data, _ = await camera.get_point_cloud()
+            data, _ = await my_camera.get_point_cloud()
 
             # write the point cloud into a temporary file
             with open("/tmp/pointcloud_data.pcd", "wb") as f:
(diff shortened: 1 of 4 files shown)


src/viam/components/camera/camera.py
```diff
@@ -48,7 +48,10 @@ async def get_image(
 
         ::
 
-            my_camera = Camera.from_robot(robot=robot, name="my_camera")
+           # from viam.media.video import CameraMimeType    
+           # Add the above to imports at top of file!
```
npentrel, line 52: They can add it anywhere above where it's used. If they use python, they don't need to be told the rest
```suggestion
           from viam.media.video import CameraMimeType
```

npentrel (approved):
LGTM % nits

src/viam/components/camera/camera.py
```diff
@@ -48,7 +48,10 @@ async def get_image(
...
+
```
npentrel, line 53: Make sure to set your editor to trim whitespace at the end of lines
```suggestion

```

src/viam/components/camera/camera.py
```diff
@@ -48,7 +48,10 @@ async def get_image(
 
         ::
 
-            my_camera = Camera.from_robot(robot=robot, name="my_camera")
+           # from viam.media.video import CameraMimeType    
+           # Add the above to imports at top of file!
+        
+            my_camera = Camera.from_robot(robot=machine, name="my_camera")
```
npentrel, line 54: If you haven't, make sure to set your editor to trim whitespace at the end of lines

```suggestion

            my_camera = Camera.from_robot(robot=machine, name="my_camera")
```

src/viam/components/camera/camera.py
```diff
@@ -74,10 +77,10 @@ async def get_images(self, *, timeout: Optional[float] = None, **kwargs) -> Tupl
 
         ::
 
-            my_camera = Camera.from_robot(robot=robot, name="my_camera")
+            my_camera = Camera.from_robot(robot=machine, name="my_camera")
 
             images, metadata = await my_camera.get_images()
-            img0 = images[0].image
+            image_list = images[0]
```
purplenicole730, line 83: Technically this should be `image_list = images`. The `[0]` means it's only the first image.

purplenicole730 (approved):
One comment and after Naomi's comments get resolved, LGTM

stuqdog (commented):
lgtm modulo other folks' comments re: `image_list` and whitespace fixes.

src/viam/components/camera/camera.py
JessamyT: Okay I agree, I was just doing this because that's how we do it in another docs-side example. I'll change that one too

src/viam/components/camera/camera.py
JessamyT: I think 0 is the list of images and 1 is the metadata, according to the docs?

JessamyT: My impression was that the first image would be images[0][0]?

purplenicole730: If you look above, it's already split out. `images, metadata = await get_images...`, so it's already done the first [0].

JessamyT:
```suggestion
            first_image = images[0]
```

JessamyT: oOoh ok ty

PR #754, opened 2024-10-07
Migrate to uv

njooma:
[uv](https://docs.astral.sh/uv/) is a great new tool. Works incredibly fast.

The main change is that you should either prepend all `make` commands that require the environment with `uv run` e.g. `uv run make buf` OR activate the virtual env at the start of your session `source .venv/bin/activate; make buf`.

There's no more poetry magic of automatically managing environments. Which, while magical, was a little confusing for when you're running things inside and outside of the env. Now it's explicit.


njooma:
Closing in favor of #755

PR #755, opened 2024-10-07
Migrate to uv

njooma:
Mirror pr of #754 but on upstream


njooma:
Test are failing because they use the old workflow, but here is it passing when I test manually: https://github.com/viamrobotics/viam-python-sdk/actions/runs/11225353798

njooma:
CC @abe-winter 👀

src/viam/resource/base.py
```diff
@@ -76,7 +76,7 @@ async def do_command(
         ::
 
             command = {"cmd": "test", "data1": 500}
-            result = await component.do_command(command)
+            result = component.do(command)
```
npentrel, line 79: are do commands synchronous? @njooma

PR #774, opened 2024-10-25
RSDK-9146: Change Python TabularDataBySQL/MQL return type to raw BSON

jckras:
For context: The return type of the `TabularDataBySQL/MQL` proto has changed to return both a list of structs, the existing return type, and also a list of bytearrays that represent BSON.

- Added a required dependency to `pymongo` to use the `BSON` library
- `BSON` provides a `decode()` function for converting lists of byte sequences into Python dictionaries. 

**Testing**:
- I ensured that date fields are returned as native Python `datetime` objects instead of strings by running tabular_data_by_mql against real tabular data on viam-dev and printed the decoded data and their types: 
        Screenshot 2024-10-29 at 11 56 09 AM
- Updated the `TABULAR_QUERY_RESPONSE` testing data to include datetime objects, as the Python SDK will receive date fields in this format.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,6 +15,7 @@ dependencies = [
     "grpclib>=0.4.7",
     "protobuf==5.28.2",
     "typing-extensions>=4.12.2",
+	"pymongo>=4.10.1"
 ]
 
 [project.urls]
@@ -57,7 +58,6 @@ dev-dependencies = [
 	"sphinx-autoapi>=3.0.0; python_version>='3.9'",
 	"sphinx-rtd-theme>=2.0.0",
 	"types-pillow>=10.2.0.20240822",
-	"pymongo>=4.10.1"
 ]
 
 [tool.pytest.ini_options]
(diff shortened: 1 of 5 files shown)


jckras:
- bson.decode() returns datetime objects which I confirmed when running `tabular_data_by_mql` against real tabular data on viam-dev. 
Screenshot 2024-10-29 at 11 56 09 AM

pyproject.toml
```diff
@@ -57,6 +57,7 @@ dev-dependencies = [
 	"sphinx-autoapi>=3.0.0; python_version>='3.9'",
 	"sphinx-rtd-theme>=2.0.0",
 	"types-pillow>=10.2.0.20240822",
+	"pymongo>=4.10.1"
```
njooma, line 60: This isn't a `dev-dependency`, but rather an ACTUAL `dependency` and should be moved up into that section

njooma (changes requested):
nearly there! only one change

stuqdog (approved):

njooma (approved):

PR #776, opened 2024-10-30
DOCS-2954: Update vision.py

npentrel:


src/viam/services/vision/vision.py
```diff
@@ -93,17 +93,16 @@ async def capture_all_from_camera(
...
         ::
 
-            camera_name = "cam1"
+            my_detector = VisionClient.from_robot(machine, "my_detector")
 
-            # Grab the detector you configured on your machine
-            my_detector = VisionClient.from_robot(robot, "my_detector")
-
-            # capture all from the next image from the camera
+            # // Get the captured data for a camera
```
jckras, line 98: do we want the # and // ?

jckras (approved):
left one comment about a possible typo, LGTM!

src/viam/services/vision/vision.py
npentrel: good catch - that's a typo

npentrel, line 98:
```suggestion
            # Get the captured data for a camera
```

PR #799, opened 2024-12-10
DATA-3565 Add example parameters to GetLatestTabularData documentation

kaywux:
We have received feedback that users are unsure of what values to pass in for the GetLatestTabularData parameters, and that having an example in the documentation would help.

The plan is to add similar examples to other SDKs once this PR is approved!

### DIFF
--- a/src/viam/app/data_client.py
+++ b/src/viam/app/data_client.py
@@ -317,9 +317,9 @@ class DataClient:
 
         Args:
             part_id (str): The ID of the part that owns the data.
-            resource_name (str): The name of the requested resource that captured the data.
-            resource_subtype (str): The subtype of the requested resource that captured the data.
-            method_name (str): The data capture method name.
+            resource_name (str): The name of the requested resource that captured the data. Ex: "my-sensor".
+            resource_subtype (str): The subtype of the requested resource that captured the data. Ex: "rdk:component:sensor".
+            method_name (str): The data capture method name. Ex: "Readings".
 
         Returns:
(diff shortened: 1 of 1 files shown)


njooma (approved):
Approving based on the assumption that these values have been tested

src/viam/app/data_client.py
```diff
@@ -317,9 +317,9 @@ async def get_latest_tabular_data(
 
         Args:
             part_id (str): The ID of the part that owns the data.
-            resource_name (str): The name of the requested resource that captured the data.
-            resource_subtype (str): The subtype of the requested resource that captured the data.
-            method_name (str): The data capture method name.
+            resource_name (str): The name of the requested resource that captured the data. Ex: "my-sensor".
+            resource_subtype (str): The subtype of the requested resource that captured the data. Ex: "rdk:component:sensor".
+            method_name (str): The data capture method name. Ex: "Readings".
```
purplenicole730, line 322: Is there a data capture method called `Readings`? I'm not familiar with this method.

purplenicole730 (approved):
As long as there actually is a method called `Readings` for sensors. (The method doesn't have to be for sensors, but thought it would be better since the other examples specify a sensor.)

src/viam/app/data_client.py
kaywux: Yes there is! I chose to include this as the example since this is the <component type, capture method> pair that has the most use for this endpoint in prod

Screenshot 2024-12-17 at 3 02 14 PM
Screenshot 2024-12-17 at 3 03 28 PM

PR #800, opened 2024-12-11
DATA-3443: Add export_tabular_data to data client

katiepeters:
This PR adds the new Data endpoint [ExportTabularData](https://github.com/viamrobotics/app/pull/6969/files) to the python SDK.


src/viam/app/data_client.py
```diff
@@ -338,6 +339,44 @@ async def get_latest_tabular_data(
...
+            List[ExportTabularDataResponse]: The unified tabular data and metadata.
+
+        For more information, see `Data Client API <https://docs.viam.com/appendix/apis/data-client/>`_.
+        """
+
+        interval=CaptureInterval(start=datetime_to_timestamp(start_time), end=datetime_to_timestamp(end_time))
+        request = ExportTabularDataRequest(
+            part_id=part_id, resource_name=resource_name, resource_subtype=resource_subtype, method_name=method_name, interval=interval
+        )
+        return await self._data_client.ExportTabularData(request, metadata=self._metadata)
```
njooma, line 378: `ExportTabularDataResponse` has a lot of data in it that is maybe not needed? Which fields on the response would the user actually care about? Is it only the `payload`? Or also `tags`? 

`payload` is a proto.Struct, which has really bad UI. We should avoid using that where we can, so we should determine what's actually important to return

katiepeters: There's a note in the scope to optionally only return the payload in the future. As for our comment regarding proto.Struct, probably best for us to talk in person, so I'll wanter over to your area again at some point!

src/viam/app/data_client.py
```diff
@@ -145,6 +148,69 @@ def __eq__(self, other: object) -> bool:
...
+        part_name: str
+        """The robot part name"""
+
+        method_parameters: Mapping[str, Any]
+        """Additional parameters associated with the data capture method"""
+
+        tags: List[str]
+        """A list of tags associated with the data point"""
+
+        payload: Mapping[str, Any]
```
njooma, line 191: Because this payload is coming from a `proto.Struct`, you might want to make this `Mapping[str, ValueTypes]` (`from viam.utils import ValueTypes`)

njooma (approved):
This is great, thanks!

vijayvuyyuru (approved):
These changes look good to me, but will defer to anything the SDK team says on this. Thanks Katie!

purplenicole730 (approved):

PR #803, opened 2024-12-17
RSDK-9503 accept bson queries in MQL function

purplenicole730:


src/viam/app/data_client.py
```diff
@@ -269,7 +269,11 @@ async def tabular_data_by_sql(self, organization_id: str, sql_query: str) -> Lis
         response: TabularDataBySQLResponse = await self._data_client.TabularDataBySQL(request, metadata=self._metadata)
         return [bson.decode(bson_bytes) for bson_bytes in response.raw_data]
 
-    async def tabular_data_by_mql(self, organization_id: str, mql_binary: List[bytes]) -> List[Dict[str, Union[ValueTypes, datetime]]]:
+    async def tabular_data_by_mql(
+        self,
+        organization_id: str,
+        mql_query: Union[List[bytes], Dict[str, Union[ValueTypes, datetime]]]
```
purplenicole730, line 275: The parameter name change from `mql_binary` to `mql_query` is questionable. Even though we only want to accept queries, we wanted to make a union so that it wasn't a breaking change yet. Changing the name, however, is a breaking change to any user using named parameters, so at this point, one could argue we might as well change the input type.
On the other hand, leaving it as `mql_binary` is now incorrect, and might push users to keep using bytes when we don't want that. Fine with either way, as long as it's in the direction we want the SDK to go.

src/viam/app/data_client.py
```diff
@@ -269,17 +269,18 @@ async def tabular_data_by_sql(self, organization_id: str, sql_query: str) -> Lis
         response: TabularDataBySQLResponse = await self._data_client.TabularDataBySQL(request, metadata=self._metadata)
         return [bson.decode(bson_bytes) for bson_bytes in response.raw_data]
 
-    async def tabular_data_by_mql(self, organization_id: str, mql_binary: List[bytes]) -> List[Dict[str, Union[ValueTypes, datetime]]]:
+    async def tabular_data_by_mql(
+        self, organization_id: str, mql_queries: Union[List[bytes], List[Dict[str, Union[ValueTypes, datetime]]]]
```
purplenicole730, line 273: The parameter name change from mql_binary to mql_query is questionable. Even though we only want to accept queries, we wanted to make a union so that it wasn't a breaking change yet. Changing the name, however, is a breaking change to any user using named parameters, so one could argue we might as well change the input type completely now.
On the other hand, leaving it as mql_binary is now incorrect, and might push users to keep using bytes when we don't want that.

Fine with keeping or changing the parameter name, as long as it's in the direction we want the SDK to go.

stuqdog: Hmm... I think we definitely want this to not be breaking. This functionality has been around in python long enough - and python is well used enough - that it seems highly likely that we would in fact break someone's workflow.

One option is, we could make a decorator that allows for argument aliasing. Then we can change the name, but set `mql_binary` as an accepted alias for backwards compatibility. Something like the following (cribbed from https://stackoverflow.com/questions/41784308/keyword-arguments-aliases-in-python)

```python
from typing import Callable
import functools

...

def _alias_param(param_name: str, param_alias: str) -> Callable:
    """
    Decorator for aliasing a param in a function

    Args:
        param_name: name of param in function to alias
        param_alias: alias that can be used for this param
    Returns:
    """
    def decorator(func: Callable):
        @functools.wraps(func)
        def wrapper(*args, **kwargs):
            alias_param_value = kwargs.get(param_alias)
            if alias_param_value:
                kwargs[param_name] = alias_param_value
                del kwargs[param_alias]
            result = func(*args, **kwargs)
            return result
        return wrapper
    return decorator
```

Then, we'd define the function like so:
```python
@_alias_param("mql_queries", param_alias="mql_binary")
async def tabular_data_by_mql(
        self, organization_id: str, mql_queries: Union[List[bytes], List[Dict[str, Union[ValueTypes, datetime]]]]
    ) -> List[Dict[str, Union[ValueTypes, datetime]]]:
...
```

After that, we should be providing support for the query as an unnamed param, a `mql_queries` named param, and a `mql_binary` named param.

This has a nice property of being applicable in the future if we need to change param names. I do think we should underscore-prefix the decorator name to try and keep it private and write some documentation that provides clarification to future developers about the intended use case.

src/viam/app/data_client.py
```diff
@@ -288,14 +289,18 @@ async def tabular_data_by_mql(self, organization_id: str, mql_binary: List[bytes
...
 
         Returns:
             List[Dict[str, Union[ValueTypes, datetime]]]: An array of decoded BSON data objects.
 
         For more information, see `Data Client API <https://docs.viam.com/appendix/apis/data-client/>`_.
         """
+        if isinstance(mql_queries[0], dict):
+            mql_binary = [bson.encode(query) for query in mql_queries if isinstance(query, dict)]
+        else:
+            mql_binary = [query for query in mql_queries if isinstance(query, bytes)]
```
stuqdog, line 303: I wonder if it would make sense to raise in a case where the user does a mix and match? Right now, if e.g. someone had a list with 1 dict followed by 100 bytes (or vice versa), we'd only capture the first item and drop everything else. But, it seems to me that we would consider the user input to be malformed in such a case. Raising or returning an error to clarify that we need _all_ bytes or _all_ dicts seems better than silently ignoring certain inputs, imo.

stuqdog (commented):
One minor thing and I think we should revisit the name changing question, but otherwise this looks good!

src/viam/app/data_client.py
purplenicole730: You're completely right. I was trying to appease the linter but in the process ignored errors completely, which we definitely DON'T want. I think this is an appropriate place to ignore the linter as an error will pop up here if it's not of the correct type.

src/viam/app/data_client.py
```diff
@@ -86,6 +87,31 @@
...
+            alias_param_value = kwargs.get(param_alias)
+            if alias_param_value:
+                # Only use alias value if param is not given.
+                if not kwargs.get(param_name):
+                    kwargs[param_name] = alias_param_value
+                del kwargs[param_alias]
+            result = func(*args, **kwargs)
+            return result
+        return wrapper
+    return decorator
```
stuqdog, line 112: I suspect we want this to live in a more generic location, as this could easily end up having applications outside of a data client. Probably `utils.py`.

stuqdog (approved):
Nice! One small code move request, otherwise lgtm!

PR #805, opened 2024-12-19
RSDK-9590 - upgrade numpy

stuqdog:


Makefile
```diff
@@ -16,7 +16,7 @@ buf: clean
 	rm -rf src/viam/gen
 	chmod +x plugin/main.py
 	uv pip install protoletariat
-	uv pip install protobuf --upgrade
+	uv pip install protobuf==5.29.1
```
stuqdog, line 19: @njooma this was the one part I wasn't sure on from your commit, I went with the hard locking to `5.29.1` but will change back if that's not right!

penguinland: :+1: I like pinning the version used! That way, when a new employee joins the team in 6 months and gets protobuf 5.30.0, you don't need to track down bugs that happen on your machine but not on theirs, or vice versa.

Slight preference towards pinning the proletariat version (and all other dependencies), too, but that's perhaps off-topic for the current PR.

src/viam/services/mlmodel/utils.py
```diff
@@ -37,7 +39,14 @@ def make_ndarray(flat_data, dtype, shape):
         """Takes flat data (protobuf RepeatedScalarFieldContainer | bytes) to output an ndarray
         of appropriate dtype and shape"""
         make_array = np.frombuffer if dtype == np.int8 or dtype == np.uint8 else np.array
-        return make_array(flat_data, dtype).reshape(shape)
+        # as per proto, int16 and uint16 are stored as uint32. As of numpy v2, this creates
+        # some strange interactions with negative values for int16 specifically. creating
```
penguinland, line 43: Mild preference to capitalize the first word of each sentence (both lines 42 and 43)

src/viam/services/mlmodel/utils.py
```diff
@@ -37,7 +39,14 @@ def make_ndarray(flat_data, dtype, shape):
...
+        # our array as a uint32 array initially and then casting to int16 solves this.
```
penguinland, line 44: Mild preference to clarify what strange interactions you saw. but I guess it's not so important if they're now fixed. Leave as-is if it's too hard to describe them.

Is it worth making a unit test for this? "No" is a fine answer, but I figured I'd ask.

penguinland (approved):
LGTM!

src/viam/services/mlmodel/utils.py
stuqdog: We already have unit tests for all types of np array conversions, including i16 arrays with negative values. I'm open to the idea of adding new ones but I think all our cases are covered, unless I'm missing something!

src/viam/services/mlmodel/utils.py
stuqdog: Done!

src/viam/services/mlmodel/utils.py
stuqdog: clarification on behavior added!

stuqdog:
> Slight preference towards pinning the proletariat version (and all other dependencies), too, but that's perhaps off-topic for the current PR.

Yeah I'm inclined to leave that for a future PR, though I do like the suggestion!

src/viam/services/mlmodel/utils.py
penguinland: ...if all the cases are covered, did you find this problem because one of the unit tests failed? I might have misunderstood.

stuqdog: Ahh no, the problem is we don't currently support/test python 3.13, and continue to use numpy1.26 in 3.8 through 3.12. 3.13 requires numpy 2.0 or higher, which is where the issue arises.

If we were to have added 3.13 support without making this change, at that point tests would start failing.

penguinland: That makes sense! I'm sure you'll add tests for 3.13 eventually, but it might be too early for that now.

pyproject.toml
```diff
@@ -47,7 +47,8 @@ dev-dependencies = [
 	"myst-nb>=1.0.0; python_version>='3.9'",
 	"nbmake>=1.5.4",
 	"numpy<1.25.0; python_version<'3.9'",
-	"numpy>=1.26.2,<2; python_version>='3.9'",
+	"numpy>=1.26.2,<2; python_version>='3.9' and python_version <'3.13'",
+	"numpy>=2; python_version>='3.13'",
```
njooma, line 51: I don't think we need all of this anymore. I think we just need `"numpy<1.25.0; python_version<'3.9'",`, and then `"numpy>=1.26.2; python_version>='3.9'",`

njooma (approved):
minor thing that i'm not sure is needed so i'm approving anyway

pyproject.toml
stuqdog: ooh yeah, good point! will update.

stuqdog: As an aside, that does mean we'll need to update the version check logic in `flat_tensors_to_ndrrays` to check the numpy version rather than the python version.

PR #807, opened 2024-12-20
DOCS-2347: Update example on GetLatestTabularData

sguequierre:
* feedback from bill where he wanted example values in the example 
* tested out example

### DIFF
--- a/src/viam/app/data_client.py
+++ b/src/viam/app/data_client.py
@@ -378,7 +378,7 @@ class DataClient:
                 method_name="GetImage"
             )
 
-            if tabular_data and len(tabular_data) == 3:
+            if tabular_data:
                 time_captured, time_synced, payload = tabular_data
                 print(f"Time Captured: {time_captured}")
                 print(f"Time Synced: {time_synced}")
--- a/src/viam/app/data_client.py
+++ b/src/viam/app/data_client.py
@@ -371,12 +371,20 @@ class DataClient:
 
         ::
 
-            time_captured, time_synced, payload = await data_client.get_latest_tabular_data(
-                part_id="<PART-ID>",
-                resource_name="<RESOURCE-NAME>",
-                resource_subtype="<RESOURCE-SUBTYPE>",
-                method_name="<METHOD-NAME>"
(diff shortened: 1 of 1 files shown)


src/viam/app/data_client.py
```diff
@@ -371,13 +371,21 @@ async def get_latest_tabular_data(
...
-                resource_subtype="<RESOURCE-SUBTYPE>",
-                method_name="<METHOD-NAME>"
+            tabular_data = await data_client.get_latest_tabular_data(
+                part_id="77ae3145-7b91-123a-a234-e567cdca8910",
+                resource_name="camera-1",
+                resource_subtype="rdk:component:camera",
+                method_name="GetImage"
             )
 
+            if tabular_data and len(tabular_data) == 3:
```
stuqdog, line 381:
```suggestion
            if tabular_data:
```
(minor) I think the `len` check here is unnecessary, since the return value will always be either `None` or a tuple of length 3.

stuqdog (approved):

src/viam/app/data_client.py
```diff
@@ -371,13 +371,21 @@ async def get_latest_tabular_data(
 
         ::
 
-            time_captured, time_synced, payload = await data_client.get_latest_tabular_data(
-                part_id="<PART-ID>",
-                resource_name="<RESOURCE-NAME>",
-                resource_subtype="<RESOURCE-SUBTYPE>",
-                method_name="<METHOD-NAME>"
+            tabular_data = await data_client.get_latest_tabular_data(
+                part_id="77ae3145-7b91-123a-a234-e567cdca8910",
```
stuqdog, line 375: (q) is this a real part-id? Is there any reason we might want to redact it?

src/viam/app/data_client.py
sguequierre: Gotcha, if I run without the check though and it returns `None` I get this error `TypeError: cannot unpack non-iterable NoneType object`. Is that expected/ok?

src/viam/app/data_client.py
sguequierre: Not a real part-id!

PR #808, opened 2024-12-26
RSDK-9560: remove `AnalogNames` and `DigitalInterruptNames` from python sdk

martha-johnston:
removed:  
- `analog_names()`
- `digital_interrupt_names()`
- test functions of both


njooma (approved):

martha-johnston:
are there any other tests I should run before merging @njooma?

njooma:
> are there any other tests I should run before merging @njooma?

I don't think so -- should be OK as is

PR #810, opened 2024-12-30
DOCS-3233: Add readings key info to tabular data capture upload

sguequierre:
- found by matt vella

### DIFF
--- a/src/viam/app/data_client.py
+++ b/src/viam/app/data_client.py
@@ -1360,6 +1360,7 @@ class DataClient:
 
         Args:
             tabular_data (List[Mapping[str, Any]]): List of the data to be uploaded, represented tabularly as a collection of dictionaries.
+                Must include the key `readings` for sensors.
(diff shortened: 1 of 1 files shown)


src/viam/app/data_client.py
```diff
@@ -1360,6 +1360,7 @@ async def tabular_data_capture_upload(
 
         Args:
             tabular_data (List[Mapping[str, Any]]): List of the data to be uploaded, represented tabularly as a collection of dictionaries.
+                Must include the key `readings` for sensors.
```
sguequierre, line 1363:
```suggestion
                Must include the key "readings" for sensors.
```

stuqdog (approved):

sguequierre:
@stuqdog Thank you for the review, if you get a chance can you re-review? Naomi pointed out that it would be better to have this example be a sensor so I changed the example to a movement sensor

PR #831, opened 2025-01-28
RSDK-9798 - ensure machine status is ready

stuqdog:
- Ensures machine status is ready when initially creating a robot client.
- Adds a background loop to the robot client that checks machine status, and takes the lock when status is initializing.

### DIFF
--- a/src/viam/robot/client.py
+++ b/src/viam/robot/client.py
@@ -318,7 +318,7 @@ class RobotClient:
         For more information, see `Machine Management API <https://docs.viam.com/appendix/apis/robot/>`_.
         """
         with self._lock:
-            self._refresh_inner()
+            await self._refresh_inner()
 
     async def _refresh_inner(self):
         response: ResourceNamesResponse = await self._client.ResourceNames(ResourceNamesRequest())
@@ -385,7 +385,7 @@ class RobotClient:
                 # and therefore that the robot got an incorrect response from the `ResourceNames` call.
                 # So, we should refresh right now to make sure everything is correct, before we release
                 # the lock.
-                self._refresh_inner()
+                await self._refresh_inner()
 
     async def _refresh_every(self, interval: int):
         while True:
--- a/src/viam/robot/client.py
+++ b/src/viam/robot/client.py
@@ -264,6 +264,10 @@ class RobotClient:
         )
 
         try:
+            # make sure the robot status isn't initializing (i.e., that the robot is ready for queries)
+            # before we try to refresh
+            while await self._check_still_initializing():
+                pass
             await self.refresh()
         except Exception:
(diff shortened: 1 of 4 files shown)


src/viam/robot/service.py
```diff
@@ -33,6 +35,11 @@ def _generate_metadata(self) -> List[ResourceName]:
 
         return list(md)
 
+    async def GetMachineStatus(self, stream: Stream[GetMachineStatusRequest, GetMachineStatusResponse]) -> None:
+        request = await stream.recv_message()
+        assert request is not None
+        await stream.send_message(GetMachineStatusResponse(state=GetMachineStatusResponse.STATE_RUNNING))
```
stuqdog, line 41: Wanted to flag this, for the purposes of the documentation tests we need the service to have a `GetMachineStatus` implementation. But, I'm leaving it fairly barebones (it just always returns that we're running) because the issue we're trying to fix is one that exposes itself when using RDK as a server specifically.

src/viam/robot/client.py
```diff
@@ -281,6 +285,10 @@ async def _with_channel(cls, channel: Union[Channel, ViamChannel], options: Opti
                 name=f"{viam._TASK_PREFIX}-robot_check_connection",
             )
 
+        self._check_status_task = asyncio.create_task(
+            self._check_status(), name=f"{viam._TASK_PREFIX}-robot_check_status"
+        )
+
```
purplenicole730, line 291: What is the purpose of this variable? I don't see it being called anywhere

purplenicole730 (approved):
This looks good to me. Thanks for looking into this!

src/viam/robot/client.py
stuqdog: It's a background loop task to ensure the robot status remains ready, similar to `_check_connection_task` and `_refresh_task`

stuqdog:
Note that this should not be merged until we've had an opportunity to discuss with team netcode and ensure both teams are on the same page as to when and how to use `get_machine_status` to verify machine readiness (namely, should we return a robot immediately and find some way to delay `resource_names` calls if we're not connected? or should we delay returning a robot when initially created until the connection status is ready? or some third thing?)

src/viam/robot/client.py
purplenicole730: Ah, I forgot how `create_task` worked! Thanks for clarifying

stuqdog:
Putting in draft state while conversation around the preferred implementation/approach develops.

stuqdog:
Closing because this fix is a bit of a lie (it ensures _initial_ connectivity but doesn't actually ensure that we're connected), because it was initially in a scope doc and then removed (so it shouldn't be done without greater discussion/intention), and because most users won't hit this (unless they're trying to query a slow-initializing robot immediately after it goes online).

PR #833, opened 2025-01-30
Move link check to be weekly

npentrel:
Getting lots of duplicate tickets that we don't need. Moving to test once on mondays


.github/workflows/linkcheck.yml
```diff
@@ -8,7 +8,7 @@ name: run-htmltest-external
 on:
   schedule:
     # 10am UTC on weekdays
```
npentrel, line 10:
```suggestion
    # 10am UTC on Mondays
```

PR #837, opened 2025-02-05
Update TabularDataByMQL documentation

dmhilly:
Context here: https://viaminc.slack.com/archives/C0307PGHR7E/p1738787849399009

### DIFF
--- a/src/viam/app/data_client.py
+++ b/src/viam/app/data_client.py
@@ -364,7 +364,7 @@ class DataClient:
             organization_id (str): The ID of the organization that owns the data.
                 You can obtain your organization ID from the Viam app's organization settings page.
             query (Union[List[bytes], List[Dict[str, Any]]]): "The MQL query to run, as a list of MongoDB aggregation pipeline stages. 
-                Note: Each stage can be provided as either raw BSON bytes or as a dictionary, but support for bytes will be removed in the future, 
+                Note: Each stage can be provided as either a dictionary or raw BSON bytes, but support for bytes will be removed in the future, 
                 so using a dictionary is preferred.
 
         Returns:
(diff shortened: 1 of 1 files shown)


src/viam/app/data_client.py
```diff
@@ -363,7 +363,8 @@ async def tabular_data_by_mql(
         Args:
             organization_id (str): The ID of the organization that owns the data.
                 You can obtain your organization ID from the Viam app's organization settings page.
-            query (Union[List[bytes], List[Dict[str, Any]]]): The MQL query to run as a list of BSON queries.
+            query (Union[List[bytes], List[Dict[str, Any]]]): "The MQL query to run, as a list of MongoDB aggregation pipeline stages. 
+                The stages can be provided either as raw BSON bytes, or as dictionaries.
```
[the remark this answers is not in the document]
dmhilly: Makes sense! I tweaked the language a bit so this section is now in the note. I can also just remove that part altogether and keep the note as it was before. LMK what you prefer

src/viam/app/data_client.py
```diff
@@ -363,8 +363,9 @@ async def tabular_data_by_mql(
         Args:
             organization_id (str): The ID of the organization that owns the data.
                 You can obtain your organization ID from the Viam app's organization settings page.
-            query (Union[List[bytes], List[Dict[str, Any]]]): The MQL query to run as a list of BSON queries.
-                Note: Support for bytes will be removed in the future, so using a dictionary is preferred.
+            query (Union[List[bytes], List[Dict[str, Any]]]): "The MQL query to run, as a list of MongoDB aggregation pipeline stages. 
+                Note: Each stage can be provided as either raw BSON bytes or as a dictionary, but support for bytes will be removed in the future, 
+                so using a dictionary is preferred.
```
purplenicole730, line 368: I like this change way more!
If we're keeping bytes in the comment, then I'd prefer having dictionary as the first option so that that's what people read first. `as either a dictionary or raw BSON bytes`. Otherwise, LGTM

purplenicole730 (approved):

src/viam/app/data_client.py
dmhilly: Good call! Changed :)

PR #839, opened 2025-02-07
Fix typo in data client

sguequierre:
removing extra "

### DIFF
--- a/src/viam/app/data_client.py
+++ b/src/viam/app/data_client.py
@@ -364,8 +364,8 @@ class DataClient:
         Args:
             organization_id (str): The ID of the organization that owns the data.
                 You can obtain your organization ID from the Viam app's organization settings page.
-            query (Union[List[bytes], List[Dict[str, Any]]]): The MQL query to run, as a list of MongoDB aggregation pipeline stages.
-                Note: Each stage can be provided as either a dictionary or raw BSON bytes, but support for bytes will be removed in the future,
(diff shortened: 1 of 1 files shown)


src/viam/app/data_client.py
```diff
@@ -364,8 +364,8 @@ async def tabular_data_by_mql(
         Args:
             organization_id (str): The ID of the organization that owns the data.
                 You can obtain your organization ID from the Viam app's organization settings page.
-            query (Union[List[bytes], List[Dict[str, Any]]]): "The MQL query to run, as a list of MongoDB aggregation pipeline stages. 
-                Note: Each stage can be provided as either a dictionary or raw BSON bytes, but support for bytes will be removed in the future, 
+            query (Union[List[bytes], List[Dict[str, Any]]]): The MQL query to run, as a list of MongoDB aggregation pipeline stages.
```
sguequierre, line 367:
```suggestion
            query (Union[List[bytes], List[Dict[str, Any]]]): The MQL query to run, as a list of MongoDB aggregation pipeline stages. 
```

src/viam/app/data_client.py
```diff
@@ -364,8 +364,8 @@ async def tabular_data_by_mql(
...
+                Note: Each stage can be provided as either a dictionary or raw BSON bytes, but support for bytes will be removed in the future,
```
sguequierre, line 368:
```suggestion
                Note: Each stage can be provided as either a dictionary or raw BSON bytes, but support for bytes will be removed in the future, 
```

lia-viam (approved):

PR #846, opened 2025-02-11
add py.typed file per PEP561

penguinland:
The (present-but-empty) `py.typed` file indicates to mypy (and presumably other type checkers) that this module supports type annotations. 

The problem: I have [a repo](https://github.com/viam-modules/re-id-object-tracking) which imports this SDK, but running `mypy main.py` on it gives a bunch of errors like this:
```
src/config/attribute.py:3: error: Skipping analyzing "viam.proto.app.robot": module is installed, but missing library stubs or py.typed marker  [import-untyped]
src/config/config.py:1: error: Skipping analyzing "viam.proto.app.robot": module is installed, but missing library stubs or py.typed marker  [import-untyped]
```
These errors occur despite all the type annotations in the SDK.

The solution: in my virtual environment, I ran `touch venv/lib/python3.12/site-packages/viam/py.typed`, and all those errors went away. This works because of [PEP561](https://peps.python.org/pep-0561/), which says "Package maintainers who wish to support type checking of their code MUST add a marker file named `py.typed` to their package"

This PR hopefully does the same thing to the official version. However, I'm unsure how to test this out, and would welcome help on that front!

Disclaimer: I'm still pinned to SDK version 0.29.3 from October. It's possible this has somehow been fixed since then, but if so, I didn't see it in the SDK repo.


njooma (commented):
Interesting... `viam.proto.app.robot` should not be typechecked IMO because it's a generated file. although it doesn't look like you're using those protos... this is interesting. i don't see any issue with adding it

njooma (approved):

penguinland:
Hm... I didn't actually look too closely at what this code has been importing. You're right that we're importing `ServiceConfig`, which appears to be a generated class. However, it's got stubs in src/viam/gen/app/v1/robot_pb2.pyi, which I think is what mypy will use (`src/viam/proto/app/robot.py` is mainly a wrapper around `...gen.app.v1.robot_pb2`). 

I'm gonna think about this overnight, and merge tomorrow if it still seems like a good idea...

penguinland:
...I still think this is a good idea, regardless of whether the classes I'm using are generated.

PR #852, opened 2025-02-20
DATA-3163 - Remove all references to TagsByFilter

tahiyasalam:


stuqdog (commented):
Heya! the change itself looks reasonable, but I wonder if just removing these methods immediately is the right approach. What is the timeline for removing the `TagsByFilter` APIs? I think it might make more sense here to leave these methods as they are but add deprecation warnings (see [here](https://github.com/viamrobotics/viam-python-sdk/blob/49f421daeb499be67e664ebee6e67f6e87933438/src/viam/rpc/dial.py#L372) as an example), that way anyone using these methods will at least have a bit of warning before they go away.

All that said, I have no insight into how much these methods are actually used so I don't feel too strongly! If you prefer to just remove them outright I'm certainly willing to be convinced 🙂

njooma (commented):
How were the GRPC/protobuf changes made? I think they're still included in the actual `proto` changes, so we shouldn't remove them here either. if desired, we should only update the client/tests, but we should leave anything in `gen` or `proto` dirs as is

njooma:
Further to @stuqdog's comment, is this API being deprecated or removed? If just deprecated, perhaps we also deprecate them in the SDKs instead of removing them entirely

PR #874, opened 2025-03-14
RSDK-10208 - create app client from env vars

stuqdog:
Blocked on release of https://github.com/viamrobotics/rdk/pull/4853, though tested against those changes locally and was able to create a `ViamClient` successfully.

### DIFF
--- a/src/viam/app/viam_client.py
+++ b/src/viam/app/viam_client.py
@@ -1,3 +1,4 @@
+import os
 from typing import Mapping, Optional
 
 from grpclib.client import Channel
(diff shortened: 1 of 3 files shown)


src/viam/app/viam_client.py
```diff
@@ -24,14 +25,46 @@ class ViamClient:
         ViamClient.create_from_dial_options(...)
     """
 
+    @classmethod
+    async def create_from_env_vars(cls, dial_options: Optional[DialOptions] = None, app_url: Optional[str] = None) -> Self:
+        """Create `ViamClient` using credentials set by a module.
```
njooma, line 30: I think instead of "set by a module" maybe say "using credentials set in the environment as `VIAM_API_KEY` and `VIAM_API_KEY_ID`"

njooma (changes requested):

njooma (approved):

PR #879, opened 2025-03-20
RSDK-10294 & RSDK-9902 - automatically merge proto PRs when tests pass

stuqdog:
* Creates a parent test to check so we don't have to mark 10 separate tests as required
* Switches to using the `peter-evans` workflow for 
* Provides support for automatic approval and merging of these PRs when all tests pass

Note to reviewers: see inline notes for thoughts/explanations on the code and how it was tested (or why I believe leaving it untested is acceptable)

TODO:
* once this is merged, update the required status checks to only include the umbrella `test`, rather than all the individual ones

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -60,6 +60,6 @@ jobs:
     needs: test
     runs-on: ubuntu-latest
     steps:
-        run: |
+      - run: |
           echo Python tests: ${{ needs.test.result }}
           [ "${{ needs.test.result }}" == "success" ]
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -60,7 +60,6 @@ jobs:
     needs: test
     runs-on: ubuntu-latest
     steps:
-      - name: Check Results
         run: |
           echo Python tests: ${{ needs.test.result }}
           [ "${{ needs.test.result }}" == "success" ]
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ on:
 
 jobs:
   test:
-    #if: github.repository_owner == 'viamrobotics'
+    if: github.repository_owner == 'viamrobotics'
     runs-on: ubuntu-latest
     strategy:
       matrix:
@@ -56,10 +56,11 @@ jobs:
         run: uv run make test_docs
 
   test_passing:
-    #if: github.repository_owner == 'viamrobotics'
+    if: github.repository_owner == 'viamrobotics'
     needs: test
     runs-on: ubuntu-latest
     steps:
-      - run: |
+      - name: Check Results
+        run: |
           echo Python tests: ${{ needs.test.result }}
           [ "${{ needs.test.result }}" == "success" ]
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -57,8 +57,8 @@ jobs:
 
   test_passing:
     name: All Tests Passing
-    needs: [test]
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -54,3 +54,13 @@ jobs:
...
+
+  test_passing:
+    if: github.repository_owner == 'viamrobotics'
+    needs: test
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check Results
+        run: |
+          echo Python tests: ${{ needs.test.result }}
+          [ "${{ needs.test.result }}" == "success" ]
```
stuqdog, line 66: This creates a new single test that ensures that all our tests from the test matrix have passed. This is important because we want to have tests be marked as required so that we don't auto merge a proto PR where tests have failed, but without this we need to mark all our current tests (`3.8 max`, `3.10 min`, etc) as required. Not only is this annoying, but it's error prone because certain cases might get phased out of the matrix or new cases might get added, and we could forget to update our `required` list. By having a single catch-all here and marking _that_ as required, we ensure that we're always expecting the entire matrix to pass.

This has been tested on my local branch to ensure it works.

.github/workflows/update_protos.yml
```diff
@@ -44,15 +44,35 @@ jobs:
...
+        uses: peter-evans/create-pull-request@v7
+        with:
+          commit-message: '[WORKFLOW] Updating protos from ${{ github.event.client_payload.repo_name }}, commit: ${{ github.event.client_payload.sha }}'
+          branch: 'workflow/update-protos'
+          delete-branch: true
+          title: Automated Protos Update
+          body: This is an auto-generated PR to update proto definitions. Check the commits to see which repos and commits are responsible for the changes
+          assignees: njooma
+          reviewers: njooma
+          token: ${{ secrets.GIT_ACCESS_TOKEN }}
```
stuqdog, line 57: Switching to `peter-evans/create-pull-request`. This hasn't been tested in `python` specifically but we _are_ using this same workflow and version in the typescript SDK, which is passing (the historical failures in TS were related to the token we used and have been resolved since we switched to the `REPO_READ_TOKEN`).

.github/workflows/update_protos.yml
```diff
@@ -44,15 +44,35 @@ jobs:
...
+      - name: Auto approve
+        if: steps.cpr.outputs.pull-request-operation == 'created'
+        run: gh pr review --approve "${{ stps.cpr.outputs.pull-request-number }}"
         env:
+          # (TODO ethan) hard to test for sure but it's possible this is insufficient. it's unclear
+          # to me whether an approval from `viambot` is sufficient to satisfy merge requirements.
+          # if not, then this won't work. but that's not particularly bad! it just means we'll still
+          # have to merge by hand, but then we can add `viambot` as a code owner and then automatic
+          # merging will start to work.
           GH_TOKEN: ${{ secrets.GIT_ACCESS_TOKEN }}
```
stuqdog, line 76: It's possible this token is incorrect. This token causes actions to run as `viambot` but it's unclear to me without actually trying to merge something whether an approval from `viambot` satisfies the merge requirements. I think it might not, but `viambot` _does_ have override permission in this repo so perhaps I'm wrong. 

If this _doesn't_ satisfy, then it's easy to add `viambot` to the list of reviewers from whom a review is required. However, I'd prefer to actually see a failure before doing that and avoid making such a change unless necessary.

njooma (approved):

njooma:
Looks good, but lets see how it actually works

PR #893, opened 2025-04-09
DOCS-3753: Update example.ipynb

npentrel:
We want people to go with the app-provided URLs (discussed in product sync)


njooma:
These URLs are used to test our documentation and make sure our examples work, so we can't update these URLs

npentrel:
Interesting. We can't set up a test machine in viam app and use its url? I guess you'd then need to add the credentials somewhere. I will close this and exclude this page from our data sources instead then

PR #900, opened 2025-04-14
RSDK-10366 - support rust utils on windows

stuqdog:


stuqdog (commented):
Changes to support using `rust-utils` for connection on windows. Tested on windows and connected successfully, tested on mac and observed no regression.

src/viam/sessions_client.py
```diff
@@ -47,13 +49,15 @@ class SessionsClient:
...
+    def __init__(self, channel: Channel, direct_dial_address: str, dial_options: Optional[DialOptions], *, disabled: bool = False, robot_addr: Optional[str] = None):
         self.channel = channel
         self.client = RobotServiceStub(channel)
         self._address = direct_dial_address
+        self._robot_address = robot_addr
         self._disabled = disabled
         self._dial_options = deepcopy(dial_options) if dial_options is not None else DialOptions()
-        self._dial_options.disable_webrtc = True
+        if sys.platform != "win32" and sys.platform != "cygwin":
+            self._dial_options.disable_webrtc = True
```
stuqdog, line 60: I was unable to get the `SessionsClient` to reliably connect _and stay connected_ on windows when webRTC was disabled. Leaving it enabled is a little inefficient insofar as it means we're creating a second rust runtime, but 1) the windows workflow is already a little inefficient because we're using TCP instead of UDS for our proxy, and 2) I figure a minor inefficiency here is better than not supporting a `SessionsClient` on windows at all. No regression on linux/mac since we leave it disabled when the platform isn't windows.

cc @cheukt, who set up the disabling of webRTC initially [here](https://github.com/viamrobotics/viam-python-sdk/pull/447). If there's a particular reason to be concerned about re-enabling, please let me know!

src/viam/sessions_client.py
```diff
@@ -37,6 +38,7 @@ class SessionsClient:
     channel: Channel
     client: RobotServiceStub
     _address: str
+    _robot_address: Optional[str]
```
stuqdog, line 41: We could probably get away with not storing this, since only one of `_address` and `_robot_address` is ever necessary, but the logic got slightly more finicky and less readable, and the cost of storing the address here is very small, so it's probably not a big deal and not worth flagging here but it was annoying me so I'm articulating it!

njooma: I would perhaps add a comment here differentiating between these 2. It took me a second to figure it out, so would be good to leave breadcrumbs for our future selves

njooma (approved):

PR #912, opened 2025-04-24
RSDK-10608 add last_known_update to robot part and fragment update in app client

gmulz:


njooma (approved):

njooma (changes requested):
Can you add tests please? Just adding this to the existing test should be enough, not too much effort

gmulz:
> Can you add tests please? Just adding this to the existing test should be enough, not too much effort

Yep, will do. The existing tests didn't seem that intense so was wondering how necessary it was for this change

tests/test_app_client.py
```diff
@@ -474,7 +474,8 @@ async def test_get_robot_part_history(self, service: MockApp):
     async def test_update_robot_part(self, service: MockApp):
         async with ChannelFor([service]) as channel:
             client = AppClient(channel, METADATA, ID)
-            updated_robot_part = await client.update_robot_part(robot_part_id=ID, name=NAME, robot_config=ROBOT_CONFIG)
+            updated_robot_part = await client.update_robot_part(robot_part_id=ID, name=NAME, robot_config=ROBOT_CONFIG,
```
njooma, line 477: In the `MockApp` class that this test uses as a service, you should save the `last_known_update`:

```python3
# ./tests/mocks/services.py
self.last_known_update = request.last_known_update
```

Then, instead of passing `datetime.now()`, you should pass in a set time and make sure that the time saved on the server equals the time that was passed in:

```python3
# In this test file

last_known_update = datetime.now()
updated_robot_part = await cleint.update_robot_part(..., last_known_update = last_known_update)
...
assert service.last_known_update == last_known_update
```

Same below

njooma (changes requested):
Small change to the test

tests/test_app_client.py
gmulz: ah gotcha. was confused about whether that was necessary since we don't expect `last_known_update` to affect anything about the response, but now I see the `service` acts more as a check for the args on the other end. just addressed

src/viam/app/app_client.py
```diff
@@ -193,6 +193,7 @@ def from_proto(cls, robot_part: RobotPartPB) -> Self:
         self.local_fqdn = robot_part.local_fqdn
         self.created_on = robot_part.created_on.ToDatetime() if robot_part.HasField("created_on") else None
         self.secrets = list(robot_part.secrets)
+        self.last_updated = robot_part.last_updated.ToDatetime() if robot_part.HasField("last_updated") else None
```
gmulz, line 196: @njooma I added `last_updated` to the part and fragment classes, seems like they weren't there to begin with. probably relevant as these would be the fields used for `last_known_update` anyways

njooma (approved):

PR #913, opened 2025-04-25
Update data client SDK docs to better reflect Binary ID deprecation, also spring cleaning

nathan-contino:
Made some documentation improvements to better direct users to use binary data IDs now that Binary ID has been deprecated.

Along the way, I fixed a few issues that nagged me when I built my own custom modules recently. Added some links, fixed some typos, used `~` to hide classpaths (I can change that back if we consciously want them there), fixed some formatting, cleaned up some prose to be a little more readable. Nothing semantic should be changed, if I did my job correctly, but let me know if I broke anything and I can fix it.

I also made one change to the SLAM docs; for some reason, a link in there is currently causing me trouble in the docs build pipeline. I'm not really sure what's going on with the docstring formatting over there, so if I messed it up, let me know, and I can fix it too.

thanks!


src/viam/app/data_client.py
```diff
@@ -409,11 +408,11 @@ async def get_latest_tabular_data(
             method_name (str): The data capture method name. Ex: "Readings".
 
         Returns:
-            Optional[Tuple[datetime, datetime, Dict[str, ValueTypes]]]: A return value of None means that data hasn't been synced yet for the data source
-            or the most recently captured data was over a year ago, otherwise the returned tuple contains the following:
-            datetime: The time captured,
-            datetime: The time synced,
-            Dict[str, ValueTypes]: The latest tabular data captured from the specified data source.
+            Optional[Tuple[datetime, datetime, Dict[str, ValueTypes]]]: A return value of None means that data hasn't been synced yet for
+                the data source or the most recently captured data was over a year ago, otherwise the returned tuple contains the following:
```
JessamyT, line 412: It looks like the way this got parsed resulted in "forthe" in the docs PR

src/viam/gen/robot/v1/robot_pb2.pyi
stuqdog, line 1: I think we should revert these changes and leave the auto-gen code untouched. Since these files are protobuf-generated, things can get messy and complicated quickly if people are manually editing them.

src/viam/gen/service/slam/v1/slam_pb2.pyi
stuqdog, line 1: See above comment re: manually editing the protobuf-generated files.

src/viam/app/data_client.py
```diff
@@ -1474,7 +1490,7 @@ async def streaming_data_capture_upload(
         data_request_times: Optional[Tuple[datetime, datetime]] = None,
         tags: Optional[List[str]] = None,
     ) -> str:
-        """Uploads the metadata and contents of streaming binary data.
+        """Uploads the metadata and contents of streaming binary data to app.viam.com.
```
stuqdog, line 1493: I think technically a user could pass a separate app url and so use something other than `app.viam.com` (say, e.g., someone wants to have a client to `app.viam.dev`). So I don't think we should be too specific here.

edit: looks like `app.viam.com` is hardcoded elsewhere here. Unfortunate! Given that, my slight preference is to still not update here (and possibly remove in other locations?). But I suppose it's not a huge deal, given that this has existed for a while already.

src/viam/gen/robot/v1/robot_pb2.pyi
nathan-contino: I very much agree. But I'd also like to fix the errors that this causes in our Hugo build, which does _not_ like raw links without any kind of descriptive text. Is there any way I can PR a change to the source of truth for that text to give a descriptor for the URL?

If not, I can 'fix' our docs infrastructure to handle these edge cases, but I suspect my solution will be a bit brittle and prone to breaking the build again in the future.

Regardless, I'll revert the changes to the `pyi` files.

src/viam/app/data_client.py
nathan-contino: I actually wanted to remove this elsewhere, but decided to leave the mentions in place because I didn't want to change too much at once... glad we're on the same page! I'll remove all of these mentions, which are 99% redundant.

src/viam/app/data_client.py
nathan-contino: Nice catch. I don't see any hanging whitespace or anything like that, so maybe I should try fixing it in the docs infrastructure? Let me know if you think there's somewhere else where it would make sense to make the fix, but I think I'll be diving into the Python on this one.

src/viam/gen/robot/v1/robot_pb2.pyi
stuqdog: Ah hmm... yeah that's trickier sadly. I'm not sure where (if anywhere) we could update that source of truth. @njooma do you have thoughts? My suspicion is we'll probably end up having to go with something on the Hugo side but I don't have high confidence here.

src/viam/app/data_client.py
nathan-contino: Wound up rewriting this section a bit, hopefully this weird issue disappears as a result.

nathan-contino:
@stuqdog thanks for taking a look 😎 ! I think I've addressed all of your comments, so when you get the chance please give this another look to make sure everything is in order. I think we'll handle the PYI URL descriptor issues on our side for now.

Pinging @npentrel for a final sanity check since this is my first time faffing with our Python SDK doc and I want to make sure I'm not missing anything or overlooking any processes.

src/viam/gen/service/slam/v1/slam_pb2.pyi
nathan-contino: Reverted!

njooma (approved):
thanks for all the updates!

src/viam/gen/robot/v1/robot_pb2.pyi
njooma: These comments are generated from the `.proto` files. However, I think updating the comments there to include would also be an anti-pattern. Unfortunately I think this "fix" has to be done on the Hugo side :(

nathan-contino: I have implemented a fix in Hugo, no more need to worry about this! Thanks for the info. Resolving.

nathan-contino:
@njooma This is 100% ready to merge, now! I'm happy to hit the big merge button, but since I'm not an admin of this repo, I'd appreciate if you merged this time. Let me know if you'd like me to merge PRs myself in the future.

stuqdog:
nice, looks good to me!

PR #915, opened 2025-04-30
DOCS-3940: Update app_client.py

npentrel:


src/viam/app/app_client.py
```diff
@@ -1262,6 +1262,8 @@ async def get_robot_part(self, robot_part_id: str, dest: Optional[str] = None, i
             my_robot_part = await cloud.get_robot_part(
                 robot_part_id="abc12345-1a23-1234-ab12-a22a22a2aa22"
             )
+            # Get the part's address
+            address = my_robot_part.fqdn
```
njooma, line 1266: Address isn't used anywhere? Is this just an example that it exists?

stuqdog (approved):
looks reasonable to me though I share Naveed's curiosity about address.

src/viam/app/app_client.py
npentrel: yes, fqdn is a bit obscure possibly. People are missing that that's how they'd get the machine part's address.

PR #918, opened 2025-05-06
Support Optional Dependencies

aldenh-viam:
See main PR: https://github.com/viamrobotics/rdk/pull/4963

Adds initial support for new optional dependencies feature. Users will need to update their `validate_config` function from returning `Sequence[str]` to `tuple[Sequence[str], Sequence[str]]`.

Interested in opinions on how to improve the UX - ~e.g. if user forgets or isn't aware of this change and uses the old signature, which currently may silently proceed, may raise ValueError, TypeError, and possibly others.~ (we've decided to support both for now: https://github.com/viamrobotics/viam-python-sdk/pull/918/commits/47da59b9e2309519a4d4f95a3a17875f3466d51c)

* Pylance & Pyright will warn about a type mismatch in the register line (but not in the function implementation). Mypy, Pylsp, & Ruff with default settings don't appear to detect it.
image
* add an assert or runtime type check in `module.py` between the `validator(config)` and `return ValidateConfigResponse(...)` calls?
* If needed, how to best support older versions of Python 3 without the latest type checking features?

### DIFF
--- a/src/viam/module/module.py
+++ b/src/viam/module/module.py
@@ -277,6 +277,8 @@ class Module:
         validator = Registry.lookup_validator(api, model)
         try:
             # backwards compatibility. Support both ([], []) or [] with deprecation warning.
+            # If user's validate returns [str], it will be treated as required dependencies only.
+            # Incorect formats, e.g. int, will raise ValidationError.
             _validator_return_test = validator(config)
             if not (
                 isinstance(_validator_return_test, tuple)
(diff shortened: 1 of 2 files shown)


src/viam/module/module.py
```diff
@@ -275,7 +276,33 @@ async def validate_config(self, request: ValidateConfigRequest) -> ValidateConfi
         model = Model.from_string(config.model)
         validator = Registry.lookup_validator(api, model)
         try:
-            dependencies = validator(config)
-            return ValidateConfigResponse(dependencies=dependencies)
+            # backwards compatibility. Support both ([], []) or [] with deprecation warning.
```
benjirewis, line 279: @njooma for Golang and C++, we'd like to change the `Validate`/`validate_config` signature to force module authors to rewrite their methods with the new signature (and, in doing so, become aware of the new optional dependencies feature). We've opted for a less aggressive change in Python, mostly because Python is interpreted and we have to rely runtime errors: this code will (@aldenh-viam can correct me if I'm wrong) accept both the old and new signatures of `validator` and raise an error if neither is fulfilled.

njooma (approved):

PR #924, opened 2025-05-15
RSDK-10717 - support modules on windows

stuqdog:


src/viam/rpc/server.py
```diff
@@ -105,7 +107,24 @@ async def serve(
             logging.setLevel(log_level)
         listen(self._server, RecvRequest, self._grpc_recvrequest_handler)
 
-        with graceful_exit([self._server]):
+        if ':' in path:
+            host_and_port = path.split(':')
+            host = host_and_port[0]
+            port = host_and_port[1]
```
penguinland, line 113:
```suggestion
            port = int(host_and_port[1])
```

src/viam/rpc/server.py
```diff
@@ -105,7 +107,24 @@ async def serve(
             logging.setLevel(log_level)
         listen(self._server, RecvRequest, self._grpc_recvrequest_handler)
 
-        with graceful_exit([self._server]):
+        if ':' in path:
```
penguinland, line 110:
```suggestion
        if path is not None and ':' in path:
```

src/viam/rpc/server.py
```diff
@@ -105,7 +107,18 @@ async def serve(
...
+            with graceful_exit([self._server]):
+                if path:
+                    await self._server.start(path=path)
+                    LOGGER.info(f"Serving on {path}")
+                else:
+                    await self._server.start(host, port)
+                    LOGGER.info(f"Serving on {host}:{port}")
+                await self._server.wait_closed()
+                await self.close()
+                LOGGER.debug("gRPC server closed")
```
penguinland, line 120: My first instinct is that instead of copying and pasting this paragraph, just modify `graceful_exit` to be a no-op on Windows. However, that function is imported from `grpclib.utils`, which is not code we control.

You might try wrapping that above, and using the wrapped version here:
```python
import contextlib

@contextlib.contextmanager
def graceful_exit(*args, **kws):
    if sys.platform == 'win32':
        yield  # No graceful exits are possible because Windows doesn't support signal handling!
    else:
        with grpclib.utils.graceful_exit(*args, **kws):
            yield
...
# (lines 108-121 get reverted back to the way they originally were)
```
(not sure I've coded that up correctly, but hopefully my intention is clear)

I think that would be DRYer, but I don't feel strongly and am content with things as they currently are.

penguinland (approved):
LGTM! My suggestion is optional; feel free to leave as-is if you want.

Thanks for working on this!

src/viam/rpc/server.py
stuqdog: Ooh yeah, this looks cleaner for sure. I kinda rushed and agree that the duplication is not great.

I don't have a really strong intuition on overriding `graceful_exit` so I'm not super confident on how this would work; instead I opted for the slightly-less-clean-but-still-cleaner approach of just factoring out the duplicate code into a function call. Hopefully this looks decent enough!

penguinland (approved):

njooma (approved):

PR #928, opened 2025-06-04
Added gripper GetKinematics

gabegottlob:
1. Added abstract method in gripper.py
2. Implemented GetKinematics in service.py
3. Implemented get_kinematics in client.py
4. Updated test_gripper with new tests (and added get_kinematics to the mock gripper)


stuqdog (approved):
lgtm!

PR #934, opened 2025-06-11
DATA 4228: Update SDKs and CLI 

Emerald-Z:
Updating python sdk to include dataSourceType in CREATE functionality.

Updated tests to match this change and all tests are passing.

### DIFF
--- a/tests/test_data_pipelines.py
+++ b/tests/test_data_pipelines.py
@@ -6,6 +6,7 @@ from grpclib.testing import ChannelFor
 from viam.app.data_client import DataClient
 from viam.proto.app.datapipelines import DataPipeline, DataPipelineRun, DataPipelineRunStatus
 from viam.utils import datetime_to_timestamp
+from viam.proto.app.data import TabularDataSourceType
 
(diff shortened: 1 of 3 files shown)


tests/test_data_pipelines.py
```diff
@@ -15,6 +16,7 @@
 SCHEDULE = "0 0 * * *"
 UPDATED_SCHEDULE = "0 1 * * *"
 MQL_BINARY = []
+DATA_SOURCE_TYPE = TabularDataSourceType.TABULAR_DATA_SOURCE_TYPE_STANDARD
```
gabegottlob, line 19: (nit) but @stuqdog  told me that since the default data source type is standard we could in theory have a case where this is not transmitting correctly but the test still passes because it still gets standard back. Could you maybe use a non-standard type in the test? no need to re request review

gabegottlob (approved):

tests/test_data_pipelines.py
Emerald-Z: will do!

PR #935, opened 2025-06-13
[RSDK-10924] Add IsHoldingSomething to Gripper API

gvaradarajan:


njooma (changes requested):
Minor change request

src/viam/components/gripper/gripper.py
```diff
@@ -73,6 +78,33 @@ async def grab(
...
+        Get information about whether the gripper is currently holding onto an object.
+
+        ::
+
+            my_gripper = Gripper.from_robot(robot=machine, name="my_gripper")
+
+            # Grab with the gripper.
+            holding_status = await my_gripper.is_holding_something()
+            # get the boolean result
+            is_holding_something = holding_status.is_holding_something
```
stuqdog, line 99: (q) naively, `is_holding_something` is the only thing I care about here and I don't understand what the `meta` field does at all, so this looks to me like we're just asking for an extra step to get the relevant information. Can you clarify what the `meta` field in a `HoldingStatus` actually is?

stuqdog (commented):
lgtm modulo Naveed's comment and a question for my own understanding.

stuqdog (approved):

njooma (approved):

PR #947, opened 2025-07-05
DATA-4295: SDK/CLI changes to support ExportTabularData and GetLatestTabularData

Emerald-Z:
Changes include: update to the python sdk functions for ExportTabularData and GetLatestTabularData

**Testing for Golang SDK Changes**
- updated local tests
- wrote a [example testing script](https://github.com/viamrobotics/viam-python-sdk/blob/751e1c7cba493701e571a9dcef550c4489576d5b/src/viam/app/testscripts/main.py#L1-L220) to test the SDK functions with fake data


src/viam/app/data_client.py
```diff
@@ -562,6 +563,7 @@ async def get_latest_tabular_data(
             resource_name (str): The name of the requested resource that captured the data. For example, "my-sensor".
             resource_api (str): The API of the requested resource that captured the data. For example, "rdk:component:sensor".
             method_name (str): The data capture method name. For exampe, "Readings".
+            additional_params (dict): Optional collection of additional parameters to be passed to the query.
```
vijayvuyyuru, line 566: [nit] I think we could reword this to be something like optional additional parameters of the resource that captured the data. So that its in the same format as above.

src/viam/utils.py
```diff
@@ -151,6 +152,11 @@ def _convert(v: ValueTypes) -> Any:
     struct.update({k: _convert(v) for (k, v) in obj.items()})
     return struct
 
+def dict_to_any(obj: Optional[Mapping[str, ValueTypes]]) -> any_pb2.Any:
```
vijayvuyyuru, line 155: I think this function is unused?

src/viam/app/data_client.py
Emerald-Z: done!

src/viam/utils.py
Emerald-Z: yes... (i forgot to delete it ૮(˶ㅠ︿ㅠ)ა)

gloriacai01 (approved):
Great testing!!!

PR #951, opened 2025-07-16
RSDK-11344 - implement restart_module

stuqdog:
cc @felixreichenbach

### DIFF
--- a/src/viam/robot/client.py
+++ b/src/viam/robot/client.py
@@ -35,6 +35,8 @@ from viam.proto.robot import (
     Operation,
     ResourceNamesRequest,
     ResourceNamesResponse,
+    RestartModuleRequest,
+    RestartModuleResponse,
     RobotServiceStub,
     ShutdownRequest,
     StopAllRequest,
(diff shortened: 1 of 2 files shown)


njooma (approved):

tests/test_board.py
```diff
@@ -361,8 +361,9 @@ async def test_set_power_mode(self, board: MockBoard, service: BoardRPCService):
             client = BoardClient(name=board.name, channel=channel)
             pm_mode = PowerMode.POWER_MODE_OFFLINE_DEEP
             pm_timedelta = timedelta(minutes=1)
-            await client.set_power_mode(mode=pm_mode, duration=pm_timedelta, timeout=1.1)
-            assert board.timeout == loose_approx(1.1)
+            await client.set_power_mode(mode=pm_mode, duration=pm_timedelta, timeout=9.83)
```
stuqdog, line 364: For some reason the timeout was dropping by about 35 milliseconds on this call in CI only, whereas our tolerance is 1% of the timeout, or 11 milliseconds. The precise cause is unclear but it's likely just some network latency; when running locally we get a totally reasonable timeout. Either way, raising this to a larger value of 9.83 seconds (which allows for forgiveness of 98 milliseconds) causes the test to reliably pass.

Incidentally, we may want to consider changing the `loose_approx` logic to use an absolute tolerance rather than a relative tolerance. Regardless of what we set the timeout to, we want rough equivalency, less minor floating point weirdness or network latency. Ticket [here](https://viam.atlassian.net/browse/RSDK-11353)

PR #954, opened 2025-07-16
RSDK-11346 - fix loose_approx

stuqdog:


tests/test_camera.py
```diff
@@ -140,9 +140,9 @@ async def test_get_image(self, camera: MockCamera, service: CameraRPCService, im
 
             # Test known mime type
             request = GetImageRequest(name="camera", mime_type=CameraMimeType.PNG)
-            response: GetImageResponse = await client.GetImage(request, timeout=18.2)
+            response: GetImageResponse = await client.GetImage(request, timeout=18.1)
```
stuqdog, line 143: grpclib [rounds timeouts down to the nearest second](https://github.com/vmagamedov/grpclib/blob/b841b4e861c4ac82c7e1f3bfa0aea03dd0b9ba18/grpclib/metadata.py#L49) when serializing them over the wire. 18.2 isn't loosely equal to 18 within a tolerance of 1%, but 18.1 is! 🙃

tests/__init__.py
```diff
@@ -2,4 +2,4 @@
 
 
 def loose_approx(val):
-    return pytest.approx(val, rel=val * 1e-2)
```
stuqdog, line 5: The way this was written before had the tolerance equal to N percent, where N = val. So, e.g., `1 == loose_approx(99999999999)` evaluated to `True`.

lia-viam: 😬

njooma (approved):

PR #963, opened 2025-08-04
add None return type to close() in viam_client

eclair4151:
in viam_client this adds the return type to close()

    def close(self) -> None:

Otherwise type checkers like mypi throw errors during strict checking because a return type is not explicitly set.

Screenshot 2025-08-04 at 5 33 09 PM

### DIFF
--- a/src/viam/app/viam_client.py
+++ b/src/viam/app/viam_client.py
@@ -206,7 +206,7 @@ class ViamClient:
         """
         return ProvisioningClient(self._channel, self._metadata)
 
-    def close(self):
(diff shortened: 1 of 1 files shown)


eclair4151:
Also wanted to confirm this is already done in most of the rest of your code base, this close call is one of the few void functions that doesnt specify a None return type

Screenshot 2025-08-06 at 9 12 20 AM
Screenshot 2025-08-06 at 9 12 12 AM

stuqdog (approved):

PR #988, opened 2025-08-27
Automated Protos Update

viambot:
This is an auto-generated PR to update proto definitions. Check the commits to see which repos and commits are responsible for the changes


stuqdog:
> It looks like the AI updater is failing. It's been a while since I worked in this repo. Is there anything special I need to do?
> 
> I also implemented this service in #987, so this failing step may not be necessary.

hey sorry for the delay! The AI updater is still having some of the kinks ironed out; it passing is not necessary for a merge.

PR #1001, opened 2025-09-15
[RSDK-11569] fix race when creating session

npmenard:
When calling concurrently an API that requires session monitoring a race happens leading to the creation of two sessions. 

Consider two motor if a user calls `GoTo` for each concurrently each resource will create a session. Ultimately the second session created will abort the heartbeat thread spawned by the first session. Leading in turn to the cancellation of the first session by RDK the first motor will then stop.

Interestingly the second motor will be assigned a nil session by RDK which allow the resource to continue to actuate in the absence of heartbeats or a working sessions.

I attempted to fix it by keeping changes as light as possible and avoid reverting [RSDK-4455](https://viam.atlassian.net/browse/RSDK-4455) not a python guru so not sure these changes are sane.

[RSDK-4455]: https://viam.atlassian.net/browse/RSDK-4455?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Should fix RSDK-11569

### DIFF
--- a/src/viam/sessions_client.py
+++ b/src/viam/sessions_client.py
@@ -7,6 +7,8 @@ from datetime import timedelta
 from enum import IntEnum
 from threading import Lock, Thread
 from typing import MutableMapping, Optional
+from concurrent.futures import ThreadPoolExecutor
+from contextlib import asynccontextmanager
 
 from grpclib import Status
 from grpclib.client import Channel
@@ -46,6 +48,7 @@ class SessionsClient:
     _heartbeat_interval: Optional[timedelta]
     _supported: _SupportedState
     _thread: Optional[Thread]
+    _pool: ThreadPoolExecutor
 
     _HEARTBEAT_MONITORED_METHODS: MutableMapping[str, bool] = {}
 
@@ -71,6 +74,7 @@ class SessionsClient:
         self._heartbeat_interval = None
         self._supported = _SupportedState.UNKNOWN
         self._thread = None
+        self._pool = ThreadPoolExecutor()
 
         listen(self.channel, SendRequest, self._send_request)
         listen(self.channel, RecvTrailingMetadata, self._recv_trailers)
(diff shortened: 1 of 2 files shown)


src/viam/sessions_client.py
```diff
@@ -104,40 +108,45 @@ async def _recv_trailers(self, event: RecvTrailingMetadata):
...
             self._heartbeat_interval = response.heartbeat_window.ToTimedelta()
             self._current_id = response.id
 
-        # tick once to ensure heartbeats are supported
-        await self._heartbeat_tick(self.client)
+            # tick once to ensure heartbeats are supported
+            await self._heartbeat_tick(self.client)
 
-        with self._lock:
             if self._thread is not None:
```
npmenard, line 150: the thread would be killed when the second `StartSession` would complete

src/viam/sessions_client.py
```diff
@@ -104,40 +108,45 @@ async def _recv_trailers(self, event: RecvTrailingMetadata):
...
+            if response is None:
+                raise GRPCError(status=Status.INTERNAL, message="Expected response to start session")
+
+            if response.heartbeat_window is None:
+                raise GRPCError(status=Status.INTERNAL, message="Expected heartbeat window in response to start session")
 
-        with self._lock:
             self._supported = _SupportedState.TRUE
             self._heartbeat_interval = response.heartbeat_window.ToTimedelta()
             self._current_id = response.id
```
npmenard, line 145: note that RDK would send a different session id

cheukt (commented):
changes look sane to me, no need for an approval from me given it's been a long time since I touched/owned this code

njooma:
looks nice! didn't test but it all looks good

PR #1018, opened 2025-10-16
Add parameter to create_invoice_and_charge_immediately example

npentrel:
@njooma please make sure that when you add new methods with changes in the method signature, the examples also get updated.


njooma (approved):
Thanks! Will make sure we keep the docs up to date as well

PR #1027, opened 2025-11-03
add new constructor to ViamClient docstring

abe-winter:
## What changed
It looks like there's a second constructor for `ViamClient` now? Updating docstring for discoverability.


njooma (approved):
Thaaaanks

PR #1079, opened 2026-02-02
RSDK-13254: example tests

allisonschiang:


njooma (changes requested):
In last week's meeting, we decided to not test imports but rather test BUILDING the modules. This can be done by importing the server side code and client side code into a test, then running both to make sure everything is behaving as expected

tests/test_examples.py
```diff
@@ -0,0 +1,143 @@
...
+def discover_examples():
+    """Discover example directories with entry points.
+
+    Searches for common entry point patterns. Examples that don't match are skipped.
+    """
+    examples = []
+    for d in sorted(EXAMPLES_DIR.iterdir()):
+        if not d.is_dir():
+            continue
+        for candidate in [d / "src" / "main.py", d / "main.py", d / "module.py", d / "v1" / "server.py"]:
```
allisonschiang: theres no consistent entrypoint in the examples, so to keep it dynamic (in case we add new examples in the future) I'm just adding in the most popular ones)

PR #1188, opened 2026-04-03
test: fix MockSensor do_command state tracking and add missing coverage

luckys00:
### What this PR does
Fixes an inconsistency in the `MockSensor` class where `do_command` was dropping state parameters (`timeout` and `extra`), and adds the missing test coverage for `do_command` across the Sensor, Service, and Client test suites.

### Why is it needed?
While `get_readings` and `get_geometries` properly save `self.extra` and `self.timeout` to the mock's state, the `do_command` implementation was bypassing this and simply returning the command. This makes it impossible to assert if custom commands with timeouts were actually processed correctly by the mock. 

Additionally, this PR adds proper test coverage for `do_command` in `test_sensor.py` to ensure the timeout parameter successfully travels through the gRPC channel.

### Changes Made:
- Updated `do_command` in `tests/mocks/components.py` to properly track `self.timeout` and `self.extra`.
- Added missing `do_command` tests in `TestSensor`, `TestService`, and `TestClient` (asserting timeout state logic while acknowledging gRPC drops `extra` for this specific request).


njooma (changes requested):
Hi! Thanks for contributing! We are actually in the process of updating all our tests. If you want to take a look at how the `test_camera.py` tests are now implemented, we are moving towards that for all our resources. 

If you like, you can update the sensor tests to use that framework!

luckys00:
hey @njooma is this ok.
review it.

tests/test_sensor.py
```diff
@@ -28,39 +28,16 @@
 def sensor() -> MockSensor:
```
njooma, line 28: We should no longer be using components in the `mocks` file. Instead, we should make a magic mock (see [here](https://github.com/njooma/viam-python-sdk/blob/d00eaaeb8ae7059ca9a04db1af56110c923021b6/tests/test_camera.py#L47))

tests/test_sensor.py
```diff
@@ -28,39 +28,16 @@
...
-    async def test_get_status(self, sensor):
-        status = await sensor.get_status()
-        assert status == {}
-
-    async def test_get_geometries(self, sensor):
-        geometries = await sensor.get_geometries()
-        assert geometries == GEOMETRIES
-
-
 @pytest.fixture(scope="function")
```
njooma, line 31: Remove the `manager` fixture

tests/test_sensor.py
```diff
@@ -76,11 +53,18 @@ async def test_do(self, sensor: MockSensor, service: SensorRPCService):
         async with ChannelFor([service]) as channel:
             client = SensorServiceStub(channel)
             command = {"command": "args"}
+
```
njooma, line 56: Set the return type of the mock's method to return an expected value (see [here](https://github.com/njooma/viam-python-sdk/blob/d00eaaeb8ae7059ca9a04db1af56110c923021b6/tests/test_camera.py#L78))

tests/test_sensor.py
```diff
@@ -76,11 +53,18 @@ async def test_do(self, sensor: MockSensor, service: SensorRPCService):
...
             request = DoCommandRequest(name=sensor.name, command=dict_to_struct(command))
-            response: DoCommandResponse = await client.DoCommand(request)
+            assert sensor.extra is None
+
+            response: DoCommandResponse = await client.DoCommand(request, timeout=2.34)
             result = struct_to_dict(response.result)
+
```
njooma, line 62: Assert the returned response contains the expected value set above (see [here](https://github.com/njooma/viam-python-sdk/blob/d00eaaeb8ae7059ca9a04db1af56110c923021b6/tests/test_camera.py#L88))

tests/test_sensor.py
```diff
@@ -76,11 +53,18 @@ async def test_do(self, sensor: MockSensor, service: SensorRPCService):
...
             assert result == {"command": command}
 
+            assert sensor.timeout ==expected_grpc_timeout(2.34)
```
njooma, line 65: Instead of checking fields on the mock, assert that the mocked's method is called with the expected inputs (see [here](https://github.com/njooma/viam-python-sdk/blob/d00eaaeb8ae7059ca9a04db1af56110c923021b6/tests/test_camera.py#L91))

njooma (changes requested):
This is closer! Some notes that follow the new direction we want to go with testing

PR #1223, opened 2026-05-28
RSDK-14017: Python tracing

lia-viam:
Here's the first pass at a python tracing PR. I've tried to review/revise but definitely a little out of my depth. In particular, the way we control for opentel being installed or not seems a bit weird, but as compared to the optionality of mlmodel it seems like it might not be as clean to work around since we are potentially injecting traces anywhere

### DIFF
--- a/src/viam/rpc/tracing.py
+++ b/src/viam/rpc/tracing.py
@@ -1,0 +1,357 @@
+# pyright: reportPossiblyUnboundVariable=false, reportRedeclaration=false, reportOptionalMemberAccess=false, reportAttributeAccessIssue=false, reportGeneralTypeIssues=false
+"""Automated OpenTelemetry tracing for gRPC RPCs in the Viam Python SDK.
+
+Tracing is an *optional* feature. Install the ``tracing`` extra to enable
+it::
+
+    pip install "viam-sdk[tracing]"
+
+Without the extra, every function in this module is a safe no-op so the
+rest of the SDK works unchanged. With the extra, spans are emitted only
+when the ``VIAM_MODULE_TRACING`` environment variable is set to a truthy
+value, mirroring viam-cpp-sdk.
+
+When enabled, the W3C Trace Context propagator is installed so incoming
+trace context is honored, and a custom exporter ships spans to the parent
+``viam-server`` via the ``RobotService.SendTraces`` RPC.
+
+OTLP encoding is implemented locally against the protobuf types bundled
+under ``viam.gen.opentelemetry`` so that the SDK does not depend on the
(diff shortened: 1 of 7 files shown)


src/viam/rpc/tracing.py
```diff
@@ -0,0 +1,357 @@
+# pyright: reportPossiblyUnboundVariable=false, reportRedeclaration=false, reportOptionalMemberAccess=false, reportAttributeAccessIssue=false, reportGeneralTypeIssues=false
+"""Automated OpenTelemetry tracing for gRPC RPCs in the Viam Python SDK.
+
+Tracing is an *optional* feature. Install the ``tracing`` extra to enable
+it::
+
+    pip install "viam-sdk[tracing]"
```
stuqdog, line 7: I wonder if this doesn't make tracing unnecessarily annoying to use in python. I suspect it does.

src/viam/rpc/tracing.py
```diff
@@ -0,0 +1,357 @@
...
+            end_time_unix_nano=span.end_time or 0,
+            attributes=_kvs(dict(span.attributes or {})),
+            events=events,
+            links=links,
+            status=_status_to_pb(span.status),
+        )
+
+    def _spans_to_resource_spans(spans: Iterable["ReadableSpan"]) -> List[ViamResourceSpans]:
+        """Group ``ReadableSpan``s into ``ResourceSpans`` by resource and scope."""
+        # resource_key -> (PbResource, {scope_key: (InstrumentationScope, [Span])})
```
stuqdog, line 234: I don't understand this comment, is it still necessary?

stuqdog (commented):
The core logic generally looks reasonable to me, though I admit I'm not an expert in this space and didn't test it. Provided you tested it, that's good enough for me!

That said, I do think the requirement to both set an env var _and_ separately install a new extra optional module from pip seems pretty overcomplicated. Truth be told, I'm not really a big fan of either approach. The pip thing has low discoverability and env vars are pretty flaky. Can this not just be something that a user passes when invoking the SDK? If not, I guess I slightly lean env var but definitely happy to chat!

lia-viam:
@stuqdog in practice the env var is set by the rdk if you have tracing enabled in the config! regarding the optional dependency, i agree and it complicates the implementation a fair bit as you've seen. this was on recommendation from @njooma to align with how we handle the mlmodel stuff

src/viam/rpc/tracing.py
lia-viam: yeah, just to show the shape of by_resource on the next line

stuqdog (approved):

PR #1233, opened 2026-06-22
Add viam.spatialmath: FFI bindings for rust-utils orientation math

HipsterBrown:
## Summary

Adds a new `viam.spatialmath` package: Pythonic, object-oriented wrappers over the shared `rust-utils` (`libviam_rust_utils`) spatial-math FFI, exposing orientation conversions whose results match Go `rdk/spatialmath`.

- **Six wrapper classes** — `Quaternion`, `Vector3`, `OrientationVector`, `EulerAngles`, `AxisAngle`, `RotationMatrix` — each owning an opaque FFI handle freed deterministically via `weakref.finalize`, with conversions between every representation.
- **Proto bridges** to `viam.proto.common` — `OrientationVector`/`Vector3` `from_proto`/`to_proto` and `Quaternion.from_pose`/`to_pose`. Classes are **radians-native**; the proto bridges convert to/from the Viam protos' **degrees** convention at the boundary.
- **Shared native loader** — extracted the `libviam_rust_utils` ctypes loading from `rpc/dial.py` into `viam/_native.py` so both the dialer and spatialmath load one cached `CDLL`.
- **No packaging change** — the spatialmath symbols already ship in the bundled `libviam_rust_utils`, and hatchling already bundles the lib and picks up the new sub-package.

### Conventions (matched to Go `rdk/spatialmath`)
- All angles are radians (Euler order Z-Y'-X" / yaw-pitch-roll); proto bridges convert deg↔rad.
- `RotationMatrix.elements` is row-major, matching rdk's convention.
- Zero-rotation `AxisAngle` returns the canonical axis `(0, 0, 1)`, matching rdk.

## Test Plan

- [x] **Golden parity** — `tests/spatialmath/fixtures/golden.json` is generated from canonical Go `rdk/spatialmath` (generator in `etc/spatialmath_golden/`); a parametrized test asserts all conversions (orientation-vector, Euler, axis-angle, rotation-matrix) match within `1e-9` across identity / 90°-X / 180°-Y / tumble cases.
- [x] **Unit tests** for each wrapper, proto bridges, and degrees/radians boundary behavior.
- [x] **Lifetime/leak** — 10k-iteration churn + repeated component-read tests confirm no double-free/segfault; null-pointer input raises `ValueError`.
- [x] `ruff check` and `pyright` clean on the new modules; `dial.py` refactor verified against existing dial tests.


etc/spatialmath_golden/README.md
stuqdog, line 1: I don't love this. Seems like we're introducing a go dependency to the SDK (even if only for tests), which isn't great. And I'd argue that this kind of parity testing more properly belongs in the rust-utils library.

etc/spatialmath_golden/README.md
```diff
@@ -0,0 +1,81 @@
+# spatialmath_golden
```
HipsterBrown: I'm ok with removing this directory if we feel like just keeping the generated fixture, as long as there are some docs to describe how to regenerate it if needed.

etc/spatialmath_golden/README.md
HipsterBrown: Yea, that makes sense. I'll remove this and keep the generated golden.json fixture for the verification tests. Or we can move those verifications to the rust-utils package, like you mentioned. This more serves as proof of implementation but doesn't need to be run all the time.

HipsterBrown:
@stuqdog I've removed that Go module responsible for generating the `golden.json` test fixture and added a note to the JSON to mention how it was created. I can work on a follow up in the rust-utils library to upstream the parity test, so other consumers of the spatialmath utils benefit from those changes.

stuqdog (approved):
Didn't personally test everything but given that tests are passing, assuming golden.json came from RDK this all seems good to me!

ghost:
Hey @HipsterBrown — this PR has been approved and CI has been green for 3+ business days. Ready to merge?


_Auto-comment from [overwatch](https://github.com/viamrobotics/overwatch). Will not re-nudge for 7 days._

PR #1236, opened 2026-06-30
Fix release workflow: download native lib before tests

stuqdog:
The release workflow ran `uv sync` + tests but never downloaded libviam_rust_utils, which uv sync does not fetch (only make install / etc/postinstall.sh does). The new spatialmath FFI tests load the native lib at runtime, so the release job failed with OSError (exit code 2). Add a postinstall step to download the lib, matching test.yml.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -42,6 +42,9 @@ jobs:
       - name: Install Package
         run: uv sync --all-extras
 
+      - name: Download Native Library
+        run: bash etc/postinstall.sh
+
       - name: Clean Format Test
         run: uv run make clean format typecheck test
 


HipsterBrown:
Thanks for fixing that! I'm surprised it didn't get caught in the PR that introduced the spatialmath utils.

stuqdog:
> Thanks for fixing that! I'm surprised it didn't get caught in the PR that introduced the spatialmath utils.

Yeah, turns out the test workflow for a PR was different from the test workflow for a release in that the release workflow previously wasn't downloading rust utils. Never was an issue until we needed rust utils for unit tests!

martha-johnston (approved):

PR #1245, opened 2026-07-21
switch RotationMatrix.elements to row-major FFI getter

JosephBorodach:
Switches `RotationMatrix.elements` to the new `viam_rotation_matrix_get_elements` FFI getter that rust-utils ships in the latest release, so it now returns honest row major R instead of the transpose


ale7714 (approved):
thank you!

PR #1250, opened 2026-07-29
Add move_through_joint_positions and get_3d_models to Arm (BREAKING)

HipsterBrown:
## ⚠️ Breaking change — requires a release note

`Arm` now declares two new `@abc.abstractmethod`s. **Every existing third-party arm module will fail at startup** until it implements both:

```
TypeError: Can't instantiate abstract class MyArm with abstract methods
get_3d_models, move_through_joint_positions
```

This is not a deprecation or a runtime warning — the module does not start. The release carrying this needs an explicit upgrade note, since the `TypeError` gives module authors no hint about what to implement.

Abstract (rather than concrete-raising-`MethodNotImplementedError`) was a deliberate choice for consistency with the other seven `Arm` methods.

## Summary

Brings the Python SDK's `Arm` to RPC parity with the Go SDK (`rdk`) by implementing the two arm RPCs the SDK did not expose:

- **`move_through_joint_positions(positions, options=None, ...)`** — executes a multi-waypoint joint trajectory under optional kinematic ceilings (`MoveOptions`). Previously the only way to do this from Python was a `move_to_joint_positions` loop with no velocity or acceleration control.
- **`get_3d_models() -> Mapping[str, Mesh]`** — retrieves the arm's 3D meshes.

Both are implemented across all three layers (ABC, `ArmRPCService`, `ArmClient`), plus all seven in-repo `Arm` subclasses. `MoveOptions` and `Mesh` are now re-exported from `viam.components.arm`.

No proto regeneration — the generated stubs and types already existed. `src/viam/gen/**` and `src/viam/proto/**` are untouched.

### Coverage before/after

| RPC | proto | Go RDK | Python (before) | Python (after) |
|---|---|---|---|---|
| GetEndPosition, MoveToPosition, GetJointPositions, MoveToJointPositions, Stop, IsMoving, DoCommand, GetStatus, GetKinematics, GetGeometries | ✅ | ✅ | ✅ | ✅ |
| **MoveThroughJointPositions** | ✅ | ✅ | ❌ | ✅ |
| **Get3DModels** | ✅ | ✅ | ❌ | ✅ |
| MoveThroughJointPositionsStreamed | ✅ | ❌ | ❌ | ❌ (intentional) |

## Notes for reviewers

**`MoveOptions` presence is two levels deep.** The `options` field has explicit presence, so the service branches on `request.HasField("options")` — without it, a caller who omitted options would hand the driver a zeroed `MoveOptions`, and `max_vel_degs_per_sec == 0.0` could be misread as "do not move" rather than "no limit requested". The scalars *inside* `MoveOptions` have presence too, so the ABC docstring directs implementers to check `options.HasField(...)` per field. It also documents the proto's precedence rule (`max_vel_degs_per_sec` is ignored when `max_vel_degs_per_sec_joints` is set), which nothing else in the SDK surfaced.

**`get_3d_models` returns `dict(response.models)`, not the raw container.** Protobuf map fields mimic C++ semantics — `__getitem__` on a missing key default-constructs *and inserts* rather than raising `KeyError`. Returning the live container would mean `models["gripper_link"]` on an absent link yields an empty `Mesh` and silently grows the caller's mapping, breaking the declared `Mapping[str, Mesh]` contract. A test pins the `KeyError`.

**No client-side joint-limit validation, unlike Go.** `rdk`'s client calls `Kinematics()` and runs `checkDesiredJointPositions` before every `MoveThroughJointPositions`. Python has no equivalent of `referenceframe.Model` — `get_kinematics()` returns raw URDF/SVA bytes with no parser — so there is nothing to validate against. The asymmetry is documented in the ABC docstring's `Note:`: the Go client refuses out-of-limit waypoints, the Python client forwards them to the driver. Adding a parser is its own project.

**`MoveThroughJointPositionsStreamed` deliberately left unimplemented** — it exists in the generated stubs but is implemented in neither `rdk` nor here.

## Known follow-up (not in this PR)

`get_kinematics` returns `response.meshes_by_urdf_filepath` — the same live protobuf container, with the same missing-key auto-insert bug described above. Pre-existing, so it was left out of scope here, but it warrants a separate fix.

## Test Plan

- [x] `make test` — 876 passed, 0 failed (includes `tests/test_examples.py`, which launches the `server` and `complex_module` example modules)
- [x] `make typecheck` — 0 errors (1 pre-existing warning in `src/viam/module/types.py`)
- [x] `make lint` / `make format` — clean, 301 files
- [x] `git diff --stat -- src/viam/gen src/viam/proto` — empty
- [x] `Arm.__abstractmethods__` contains both new methods; `ArmClient.__abstractmethods__` is empty
- [x] End-to-end over `ChannelFor`: waypoints and `MoveOptions` round-trip intact; omitted options arrive as `None` rather than zeroed; `get_3d_models` returns a real `dict` that raises `KeyError` without inserting
- [x] All seven `Arm` subclasses instantiate — verified in subprocesses for the example/docs arms, which no lint, typecheck, or test covers
- [x] `docs/examples/example.ipynb` — valid JSON, cell sources still list-of-lines, embedded `MyModularArm` parses

10 new tests across `TestArm`, `TestService`, and `TestClient`.

https://claude.ai/code/session_015RbNNgUjePEdPuUGAdU3qb


10zingpd:
@HipsterBrown is this ready for review? anything to call out or coordinate with other repos since this is tagged as breaking change?

seanavery (commented):
First pass here. Makes sense to upgrade the interface to match the Go SDK. 

These are the python arm modules i could find in our org (none in viam-modules):

- viam-labs/spray-coverage
- viam-labs/bullet-racecar
- viam-labs/so101-deepmimic


~~Wondering though if a default implementation raising `MethodNotImplementedError` would be better than abstract methods, which make this a breaking change for existing modules.~~

~~@njooma any opinion? Both strategies exist in the SDK today (get_kinematics is abstract, get_geometries has a raising default).~~

HipsterBrown:
@seanavery Do you have any particular feedback to address in the PR? We can take a look at updating those modules with these changes.

src/viam/components/arm/client.py
```diff
@@ -139,6 +158,17 @@ async def get_kinematics(
         response: GetKinematicsResponse = await self.client.GetKinematics(request, timeout=timeout, metadata=md)
         return (response.format, response.kinematics_data, response.meshes_by_urdf_filepath)
```
seanavery, line 160: Should we put a TODO comment and ticket to handle the case where an empty mesh is inserted into the response?
(from the Known Follow-up section in the PR description)

seanavery (approved):

src/viam/components/arm/client.py
```diff
@@ -137,8 +156,20 @@ async def get_kinematics(
...
+    async def get_3d_models(
+        self, *, extra: Optional[Dict[str, Any]] = None, timeout: Optional[float] = None, **kwargs
+    ) -> Mapping[str, Mesh]:
+        md = kwargs.get("metadata", self.Metadata()).proto
+        request = Get3DModelsRequest(name=self.name, extra=dict_to_struct(extra))
+        response: Get3DModelsResponse = await self.client.Get3DModels(request, timeout=timeout, metadata=md)
+        # Copy out of the protobuf map container: `__getitem__` on an absent key would
+        # otherwise create and insert a default-constructed value instead of raising
+        # KeyError, silently violating the Mapping contract.
+        return dict(response.models)
```
njooma, line 171: I don't think I agree with this -- I think we can just return `response.models` here, and returning a default empty value is better than raising an error I think.

acmorrow:
@HipsterBrown - Note that #1242 also implements `move_through_joint_positions`. @njooma?

src/viam/components/arm/service.py
```diff
@@ -78,6 +82,25 @@ async def MoveToJointPositions(self, stream: Stream[MoveToJointPositionsRequest,
...
 
+    async def MoveThroughJointPositions(self, stream: Stream[MoveThroughJointPositionsRequest, MoveThroughJointPositionsResponse]) -> None:
+        request = await stream.recv_message()
+        assert request is not None
+        name = request.name
+        arm = self.get_resource(name)
+        timeout = stream.deadline.time_remaining() if stream.deadline else None
+        # `options` has explicit presence; passing a zeroed message when the caller sent
+        # nothing would be indistinguishable from a real zero limit.
+        options = request.options if request.HasField("options") else None
```
njooma, line 93: [nit] I'm not sure about this -- if `request.options` can be `None`, then we don't have to do this additional check. Super minor and also I'm not sure of the structure of the object

njooma (changes requested):

njooma:
> Nick Hehr (@HipsterBrown) - Note that #1242 also implements `move_through_joint_positions`. Naveed Jooma (@njooma)?

True. If we can hold off on this (if no one is using it), we can push it all together with streaming move as well. Although, the implementations are different -- this PR has `options` for `move_through...`. I do think this version is more complete

src/viam/components/arm/service.py
HipsterBrown: `request.options` is never `None`; proto3 singular message fields return a default-constructed instance (is None: False, HasField: False). Without `HasField`, a caller who sent nothing is indistinguishable from one who sent a zeroed `MoveOptions`. We use this pattern for src/viam/components/audio_out/service.py:40 as well.

src/viam/components/arm/client.py
HipsterBrown: `response.models` is a `MessageMapContainer`, so a missing-key read doesn't just return an empty `Mesh`; it inserts it into the response and it becomes visible to len(), iteration, and serialization. So `models["nope"]` silently grows the map every lookup. `dict()` is the copy that makes the return type `Mapping[str, Mesh]` accurate.

njooma: Ahhh ok gotcha

njooma (approved):

PR #1256, opened 2026-09-02
RSDK-14459: Add manual mode arm API

npmenard:
Adds `get_properties`, `set_manual_mode(manual_mode, enabled_for)`, and `get_manual_mode` to the `Arm` abstract class. I think this is a breaking change I will update the drivers downstream.


src/viam/components/arm/arm.py
```diff
@@ -31,6 +32,11 @@ class Arm(ComponentBase):
     For more information, see `Arm component <https://docs.viam.com/dev/reference/apis/components/arm/>`_.
     """
 
+    @dataclass
+    class Properties:
+        support_manual_mode: bool = False
+        support_cartesian_commands: bool = False
+
```
njooma, line 39: Similar to the other PR -- typealias this to `arm_proto.GetPropertiesResponse` similar to https://github.com/viamrobotics/viam-python-sdk/pull/1261/changes#diff-c30fd93b57998c904ba0e0e10a7d3d7bf9a94e3931473e4f83ab74ff8f232596R35

njooma (changes requested):

src/viam/components/arm/arm.py
npmenard: took a stab at it lmk

njooma (approved):

